PR #77, opened 2023-11-15
CSUB-833: Pallet bridge (Draft)

zacharyfrederick:
Need to add weights and work on more unit tests. Also, have some questions about the flowcharts from the document.


pallets/bridge/src/types.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[derive(Clone, Encode, Decode, Eq, PartialEq, TypeInfo, MaxEncodedLen, RuntimeDebug)]
+pub enum FailureReason {
+    UnknownBurn,
+    Unauthorized,
+    BridgeError,
+}
+
+#[derive(Clone, Encode, Decode, Eq, PartialEq, TypeInfo, MaxEncodedLen, RuntimeDebug)]
+pub struct BurnId(pub u64);
```
nathanwhit, line 22: I know this isn't in the doc, but I think we should change this to

```rust
pub enum BurnId {
    Creditcoin2(u64)
}
```

So that we could add more in the future without needing a storage migration (for instance if we want to use a TX hash to refer to a burn on ethereum)

pallets/bridge/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+    pub trait Config: frame_system::Config + pallet_balances::Config {
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+        type Currency: ReservableCurrency<Self::AccountId>;
+    }
+
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(super) fn deposit_event)]
+    pub enum Event<T: Config> {
+        TestEvent,
```
nathanwhit, line 29: Remove

pallets/bridge/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+
+            ensure!(
+                in_progress.is_some(),
+                Error::<T>::InProgressCollectionNotFound
+            );
+
+            let in_progress = in_progress.expect("This should never fail");
+
+            ensure!(
+                in_progress.status != CollectionStatus::Completed,
```
nathanwhit, line 124: Feels weird that you can have things in the `InProgress` storage that aren't in progress

nathanwhit (changes requested):
A few quick comments

pallets/bridge/src/types.rs
zacharyfrederick: This is a good idea. I used a simple enum and a switch statement for the current implementation. It might be smart to move all cc2 functionality into its own pallet. Have each bridge implementation fulfill a trait with `collect, approve, and reject` calls and then have a bridge pallet that ties them all together and exposes the public interface. That would make testing and expanding all the bridge sources easier than keeping a single pallet.

PR #104, opened 2023-11-30
Bridge cleanup

nathanwhit:

When working on the worker for the bridge, I noticed some improvements to be made in the bridge pallet: 

(commit-by-commit)
- Remove the `Creditcoin3` variant from `BurnId`. It's unused and we'll have to do a runtime upgrade to handle that case in the future (if we ever do) anyway
- Add more information to the bridge events (this allows the worker to operate solely off events)
- Combine `InProgress` and `Collections` into just `Collections`. I felt like the separate `InProgress` storage item didn't carry its weight, plus it led to some weird possibilities (like an item in `InProgress` with status `Completed`). Also don't persist failure, as currently we would, for instance, refuse to process a burn ID that had failed in the past. So if someone were to try to collect for a burn with id `100` we would fail, persist that failure, and then refuse to process burn `100` in the future (even though burn id `100` will _eventually_ exist).
- Decouple the bridge pallet from the balances pallet. This one is a nitpicky thing, but basically as-is the bridge pallet was tightly coupled to the balances pallet despite the fact that it has a `Currency` type in its Config (and the only real use for doing that is to decouple from the balances pallet). So I just doubled down on decoupling and got rid of the direct dependencies on the balances pallet.

This _technically_ should require a storage migration, but we only have devnet currently and there's nothing on it so writing a whole migration (that's realistically not going to actually have any data to work on) doesn't seem worth it. 
---


atodorov (approved):
LGTM, @zacharyfrederick WDYT?

atodorov (approved):
Conflicted with https://github.com/gluwa/creditcoin3/pull/105, sorry about that. Resolved conflicts and force-pushed!

zacharyfrederick (approved):
LGTM.

PR #106, opened 2023-11-30
CSUB-883: Add sanity tests for bridge pallet

atodorov:
# Description of proposed changes

Now that we're having pallets again we need to bring some of the missing checks that we've got in CC2. Starting with a sanity check about weights and benchmarks.

Note: this builds on top of #105.

---

Practical tips for PR review & merge:

- [ ] All GitHub Actions report PASS
- [ ] Newly added code/functions have unit tests
  - [ ] Coverage tools report all newly added lines as covered
  - [ ] The positive scenario is exercised
  - [ ] Negative scenarios are exercised, e.g. assert on all possible errors
  - [ ] Assert on events triggered if applicable
  - [ ] Assert on changes made to storage if applicable
- [ ] Modified behavior/functions - try to make sure above test items are covered
- [ ] Integration tests are added if applicable/needed

PR #119, opened 2023-12-06
CSUB-883: Rename pallets/bridge to pallet-bridge

atodorov:

for consistency with other pallets and future test scripts which will be
added as part of https://github.com/gluwa/creditcoin3/pull/106.

---


pallets/bridge/src/mock.rs
```diff
@@ -0,0 +1,99 @@
...
+    type RuntimeEvent = RuntimeEvent;
+    type Currency = Balances;
+}
+
+#[derive(Default)]
+pub struct ExtBuilder;
+
+impl ExtBuilder {
+    pub fn build(self) -> sp_io::TestExternalities {
+        let mut t = system::GenesisConfig::default()
```
pLabarta, line 89:
```suggestion
        let mut t = system::GenesisConfig::<Test>::default()
```

pallets/bridge/src/mock.rs
```diff
@@ -0,0 +1,99 @@
...
+            .build_storage::<Test>()
```
pLabarta, line 90:
```suggestion
            .build_storage()
```

pallets/bridge/src/migrations/v1.rs
```diff
@@ -0,0 +1,92 @@
...
+            log::info!(
+                    target: "runtime::PalletBridge",
+                    "migrate V1, already migrated.",
+            );
+            return Weight::zero();
+        }
+
+        move_pallet(OLD_PALLET_NAME.as_bytes(), new_pallet_name.as_bytes());
+
+        Weight::MAX
```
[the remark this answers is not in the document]
atodorov: > Q: Why do we use `Weight::MAX` and what does it mean? Is it the max weight for a singular block? 

On the meaning [docs](https://paritytech.github.io/polkadot-sdk/master/frame_support/weights/struct.Weight.html#associatedconstant.MAX) say `The maximal weight in all dimensions.`

> Is there a reason for not calculating the actual weight of the migration?

The answer here is 2 fold:

1) Initially I was calculating some weight, assuming moving a storage item is 1 read + 1 write, however in practice it is probably not like that (I'm assuming prefix operations are faster). See [Actually migrate pallet-bridge storage items](https://github.com/gluwa/creditcoin3/pull/119/commits/faf8959efb50f79173a057804fce21358ed8ae7a).

2) Because of how we've designed the existing migrations and assertions in unit tests we expect the weight of the migration to be non-zero. This assert is at the bottom of `migrations/mod.rs`. Naturally I started looking to populate the storage items from the pallet with the old name and then execute migrate() as part of this unit test to satisfy the assertion. 

Then I found the example from Substrate at 
https://github.com/paritytech/substrate/pull/9115/files#diff-1dee027b2fa4c245bc171233ac07ec714f3fdc9a83a0f428adf67a2966049717R66 and introduced the commit containing `Weight::MAX` because it was the fastest option so I can push the code and get some review on it.


I wasn't able to find any context what the actual weight of a migration calling `move_pallet()` should be so I left it like that.

pLabarta (approved):

nathanwhit (approved):

PR #125, opened 2023-12-11
[CSUB-913] Selectively apply options flags

pLabarta:

---


cli/src/commands/options.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const ecdsaOption = new Option('--ecdsa', 'Use ECDSA signature instead of mnemonic');
+
+export const addressOption = new Option('--address [address]', 'Specify Substrate address');
+
+export const amountOption = new Option('--amount [amount]', 'CTC amount');
+
+export const recipientOption = new Option('--recipient [recipient]', 'Specify recipient address');
+
+export const evmAddressOption = new Option('--address [address]', 'Specify EVM address');
```
atodorov, line 20: The cli switch, `--address`, is kind of dulicate with `adressOption`. IDK what will happen if by mistake both of these are added to the same command.  My guess is that the last one to be added will override the previous one. 

Suggestion: rename to `--evm-address` to be more explicit ?  If we go this route, then maybe rename the original `--address` to `--substrate-address` to be very clear what's expected where. WDYT ?

cli/src/commands/evm/send.ts
```diff
@@ -46,7 +48,9 @@ async function evmSendAction(options: OptionValues) {
 
 function parseOptions(options: OptionValues) {
     const amount = parseAmountOrExit(requiredInput(options.amount, 'Failed to send CTC: Must specify an amount'));
-    const recipient = parseEVMAddressOrExit(requiredInput(options.to, 'Failed to send CTC: Must specify a recipient'));
+    const recipient = parseEVMAddressOrExit(
+        requiredInput(options.recipient, 'Failed to send CTC: Must specify a recipient'),
+    );
```
atodorov, line 53: 2 notes here: 

1) It will be better if validator functions can be specified directly in `options.ts`. I think this is possible.
2) For `--recipient`  sometimes a Substrate address is expected, other times an EVM address is expected. This contradicts 1) and is prone to copy-paste errors in the future

3) About 2) - I like how `--recipient` is very verbose and a no-brainer, on the other hand I also like the explicitness of `--evm-address` and `--substrate-address`.  It would be great if we can somehow combine best of both worlds! 

Maybe the `--xxxx-address` options will be understandable enough in the context of their specific commands for both the developer and the user! Food for some thought!

atodorov (changes requested):

cli/src/commands/options.ts
pLabarta: I had a thought about this, yeah would make them more descriptive.

cli/src/commands/evm/send.ts
pLabarta: >     1. It will be better if validator functions can be specified directly in `options.ts`. I think this is possible.

WDYM by "validator functions", you mean extracting the options for `bond`, `chill`, etc. to `options.ts`? They have some very specific options, but I think some of them could be extracted. You find it easier having all options bundled in the `options.ts` file?

I'm also thinking about doing some bits of refactoring and turning most options into arguments to improve the error handling and reduce boilerplate code (since Commander would handle that for us).

atodorov: > WDYM by "validator functions",

I mean the functions validating user input for the option. For `--recipient` they are `parseEVMAddressOrExit` and `parseAddressOrExit` - both used against the same option.

And FTR I think they can be specified as part of 
```
new Option('--recipient [recipient]', 'Specify recipient address');
```

cli/src/commands/evm/index.ts
```diff
@@ -13,8 +14,8 @@ export function makeEvmCommand() {
     cmd.addCommand(makeEvmWithdrawCommand());
 
     cmd.commands.forEach((command) => {
-        command.option('-u, --url [url]', 'URL for the Substrate node', 'ws://127.0.0.1:9944');
-        command.option('--no-input', 'Disable interactive prompts');
+        command.addOption(noInputOption);
+        command.addOption(urlOption);
```
atodorov, line 18: this should be `removeOption`, no ? 

IOW:

`evm --help` - shows no `--url`

`evm balance --help` shows the `--url` option.

atodorov (changes requested):

cli/src/commands/evm/balance.ts
```diff
@@ -2,12 +2,14 @@ import { Command, OptionValues } from 'commander';
...
 import { getEVMBalanceOf, logEVMBalance } from '../../lib/evm/balance';
+import { jsonOption, urlOption } from '../options';
 
 export function makeEvmBalanceCommand() {
     const cmd = new Command('balance');
     cmd.description('Show balance of an EVM account');
     cmd.argument('<address>', 'Address to check balance of');
-    cmd.option('--json', 'Output as JSON');
+    cmd.addOption(jsonOption);
+    cmd.addOption(urlOption);
```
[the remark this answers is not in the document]
atodorov: https://github.com/gluwa/creditcoin3/pull/125#discussion_r1424377161

cli/src/commands/options.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// Most used options are URL, NO INPUT, JSON, eCDSA, aDDRESS, AMOUNT, TO, EMV-ADDRESS
+// Create consts for each one using the Option class and export them to be used by other commands
+
+// Connection
+export const urlOption = new Option('-u, --url [url]', 'URL of the node to connect to').default('ws://127.0.0.1:9944');
+
+// Addresses
+export const evmAddressOption = new Option('--evm-address [address]', 'Specify EVM address');
+export const substrateAddressOption = new Option('--substrate-address [address]', 'Specify Substrate address');
```
atodorov, line 11: in commands/status.ts you are calling `options.address`, same in balance. That will fail.

cli/src/commands/options.ts
```diff
@@ -0,0 +1,22 @@
+import { Option } from 'commander';
+
+// Most used options are URL, NO INPUT, JSON, eCDSA, aDDRESS, AMOUNT, TO, EMV-ADDRESS
+// Create consts for each one using the Option class and export them to be used by other commands
+
+// Connection
+export const urlOption = new Option('-u, --url [url]', 'URL of the node to connect to').default('ws://127.0.0.1:9944');
+
+// Addresses
+export const evmAddressOption = new Option('--evm-address [address]', 'Specify EVM address');
```
atodorov, line 10: this variable is unused!

atodorov (changes requested):

cli/src/commands/evm/balance.ts
```diff
@@ -2,12 +2,13 @@ import { Command, OptionValues } from 'commander';
...
 import { getEvmUrl } from '../../lib/evm/rpc';
 import { getEVMBalanceOf, logEVMBalance } from '../../lib/evm/balance';
+import { jsonOption } from '../options';
 
 export function makeEvmBalanceCommand() {
     const cmd = new Command('balance');
     cmd.description('Show balance of an EVM account');
-    cmd.argument('<address>', 'Address to check balance of');
-    cmd.option('--json', 'Output as JSON');
+    cmd.argument('<address>', 'EVM address to check balance of');
```
atodorov, line 10: Preferably that would also be `--evm-address` as an option to keep it consistent. 

Bonus: if `parseEVMAddressOrExit` is specified for the Option, then you will not have to repeat it on line 17 below.

atodorov (approved):

zacharyfrederick (approved):
LGTM

PR #169, opened 2024-01-12
[CSUB-1011] EVM Tracer to CC3

frankli-dev:
# Description of proposed changes

<describe what this PR is about and why we want it>

---

Practical tips for PR review & merge:

- [ ] All GitHub Actions report PASS
- [ ] Newly added code/functions have unit tests
  - [ ] Coverage tools report all newly added lines as covered
  - [ ] The positive scenario is exercised
  - [ ] Negative scenarios are exercised, e.g. assert on all possible errors
  - [ ] Assert on events triggered if applicable
  - [ ] Assert on changes made to storage if applicable
- [ ] Modified behavior/functions - try to make sure above test items are covered
- [ ] Integration tests are added if applicable/needed

PR #182, opened 2024-01-23
CSUB-1011: Tracing rpc tests

nathanwhit:

---


atodorov (changes requested):
@frankli-dev, @nathanwhit code review from https://github.com/gluwa/creditcoin3/pull/169#pullrequestreview-1839032848 still holds true. 

Additionally these 2 commits can be squashed together.

atodorov (changes requested):
@frankli-dev additionally this comment on the original PR is also unresolved https://github.com/gluwa/creditcoin3/pull/169#discussion_r1452283533. Please resolve it.

cli/src/test/blockchain-tests/evm-tracing.test.ts
```diff
@@ -0,0 +1,67 @@
...
+import { deployContract } from './helpers';
+import { WebSocketProvider, ethers, parseEther } from 'ethers';
+
+describe.only('EVM Tracing', (): void => {
+    let provider: WebSocketProvider;
+    let deployedContractAddress: string;
+    let txHash: string;
+
+    beforeAll(async () => {
+        provider = new WebSocketProvider(ALICE_NODE_URL);
```
atodorov, line 12: Use `(global as any).CREDITCOIN_API_URL` b/c these tests get executed against multiple different targets which is controlled via configuration files. The variable `ALICE_NODE_URL` is still a hard-coded value. 

See https://github.com/gluwa/creditcoin3/blob/dev/cli/src/test/blockchain-tests/danger-will-brick-the-blockchain.test.ts#L8C29-L8C63

cli/src/test/blockchain-tests/evm-tracing.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        const alith = new ethers.Wallet('0x5fb92d6e98884f76de468fa3f6278f8807c48bebc13595d45af5bdc4da702133').connect(
```
atodorov, line 14: Hard-coded value should not be used. 

Should use `createSigner('lender')` to create a known Substrate address:
https://github.com/gluwa/creditcoin3/blob/dev/cli/src/test/blockchainSetup.ts#L4

and then `substrateAddressToEvmAddress()` to get the associated EVM address:
https://github.com/gluwa/creditcoin3/blob/dev/cli/src/lib/evm/address.ts#L12

atodorov: Note to self: `lender` and `borrower` don't make sense anymore, I will push a PR to update these to something like `alice` & `bob`.

cli/src/test/blockchain-tests/evm-tracing.test.ts
```diff
@@ -0,0 +1,67 @@
...
+        const response = await alith.sendTransaction({
+            to: deployedContractAddress,
+            value: parseEther('50'),
+        });
+        await response.wait();
+
+        // call contract method sendForMe to random address
+        const call = await contract
+            .getFunction('sendForMe')
+            .call(contract, randomEvmAccount().address, parseEther('10'));
```
atodorov, line 32: Question:  Does this call reduce the balance of `alith` ? 

Assuming that's the case we would like to use well known address for the receiver, instead of a random account. 

The reason being is that we use the same test suite against multiple targets (local, Devnet, Testner & Mainnet) and for Testnet & Mainnet we may want to recirculate the funds used during testing. 

See above for the usage of `createSigner` and `substrateAddressToEvmAddress`. 

See https://github.com/gluwa/creditcoin/blob/dev/integration-tests/src/mainnetSetup.ts#L4 for an example how the `createSigner` function looks for CC2 Mainnet.

.github/workflows/ci.yml
```diff
@@ -295,7 +295,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --release --features fast-runtime
+          args: --release --features fast-runtime --features=evm-tracing
```
atodorov, line 298: Can you please also rename the CI job name to `build-creditcoin-node-for-ci` to reflect the new reality.

.github/workflows/ci.yml
```diff
@@ -558,6 +558,7 @@ jobs:
             --chain dev \
             --validator --alice  --pruning archive \
             --node-key d182d503b7dd97e7c055f33438c7717145840fd66b2a055284ee8d768241a463 \
+            --ethapi=debug,trace,txpool \
```
atodorov, line 561: Adding this comment here for visibility. 

From the [original code review](https://github.com/gluwa/creditcoin3/pull/169#discussion_r1452283533):

We should add all of the newly committed Cargo.toml files into dependabot.yml in case there's ever a dependency which doesn't come from the workspace. I will send a PR to check for anything that is missing!

atodorov (changes requested):
Some changes requested

.github/workflows/ci.yml
atodorov: Additional CI checks for these missing Cargo.toml files are in https://github.com/gluwa/creditcoin3/pull/187, the list of missing files can be found at https://github.com/gluwa/creditcoin3/actions/runs/7653952405/job/20856903770?pr=187.

cli/src/test/blockchain-tests/evm-tracing.test.ts
frankli-dev: The value is private key of alice account. Not an address. Any good idea to get that?

.github/workflows/ci.yml
frankli-dev: Cool. Thanks @atodorov

.github/workflows/ci.yml
frankli-dev: done

cli/src/test/blockchain-tests/evm-tracing.test.ts
frankli-dev: done

cli/src/test/blockchain-tests/evm-tracing.test.ts
atodorov: > The value is private key of alice account. Not an address.

arght. Thanks for explanation. 

> Any good idea to get that?

You can define another function inside `blockchainSetup.ts` similar to `createSigner` to return private keys. We can then override this function in other config files as necessary. In general for a non-local environment we pass these private keys via ENV variables before starting the test suite. The details around this are in https://github.com/gluwa/creditcoin/blob/dev/integration-tests/README.md

atodorov: @frankli-dev see `(global as any).CREDITCOIN_EVM_PRIVATE_KEY ` from https://github.com/gluwa/creditcoin3/pull/191/commits/613ac2c1844f4a02980a62a9a9124e30dd9c39e7, #191

atodorov (changes requested):
- needs addressing https://github.com/gluwa/creditcoin3/pull/182#discussion_r1466189918
- ideally also a rebase

frankli-dev:
> * needs addressing [CSUB-1011: Tracing rpc tests #182 (comment)](https://github.com/gluwa/creditcoin3/pull/182#discussion_r1466189918)
> * ideally also a rebase

Done

atodorov (approved):

PR #241, opened 2024-02-16
Devnet Upgrade

zacharyfrederick:


atodorov (changes requested):
The failure from `Check runtime changes / danger-will-brick-the-blockchain` can be ignored b/c this is an intentional change and we're prepared to redeploy Devnet b/c of this breaking change.

The rest of the failures are valid and need to be fixed.

runtime/src/lib.rs
```diff
@@ -149,7 +149,7 @@ pub const MILLISECS_PER_BLOCK: u64 = prod_or_fast!(15_000, 5_000);
 
 pub const SLOT_DURATION: u64 = MILLISECS_PER_BLOCK;
 
-const BLOCKS_FOR_FASTER_EPOCH: u32 = if cfg!(feature = "zombienet") { 45 } else { 15 };
+const BLOCKS_FOR_FASTER_EPOCH: u32 = if cfg!(feature = "zombienet") { 45 } else { 2 * HOURS };
```
atodorov, line 152: @pLabarta correct me if I'm wrong but I think that this is causing the timeouts we see in `validatorCycle.test.ts` because on a few occasions we're waiting for a number of eras to complete (either for rewards or for unbonding). 

This change is making the era time too long for integration tests. 


@zacharyfrederick I think yet another compiler flag will be necessary here, similar to the existing `zombienet` feature.

atodorov (changes requested):

atodorov:
Merging after comment from Dave Cote:

> Hey Alex, Zach is telling for the devnet specfile change PR, we could merge it and address later the integration test failing,Can you please approve and merge it?
For context, it's blocking blockscout to re-index devnet

PR #243, opened 2024-02-21
CSUB-1054 All integration tests failing & blocked by recent change to Devnet

frankli-dev:
# Description of proposed changes

<describe what this PR is about and why we want it>

---

Practical tips for PR review & merge:

- [ ] All GitHub Actions report PASS
- [ ] Newly added code/functions have unit tests
  - [ ] Coverage tools report all newly added lines as covered
  - [ ] The positive scenario is exercised
  - [ ] Negative scenarios are exercised, e.g. assert on all possible errors
  - [ ] Assert on events triggered if applicable
  - [ ] Assert on changes made to storage if applicable
- [ ] Modified behavior/functions - try to make sure above test items are covered
- [ ] Integration tests are added if applicable/needed

PR #248, opened 2024-02-23
CSUB-1054: Updates for release pipeline - build with appropriate feature flags for Devnet

zacharyfrederick:


.github/workflows/release.yml
```diff
@@ -91,11 +91,19 @@ jobs:
...
 
       - name: Build
+        if: ${{ env.TAG_NAME }} != 'devnet'
         uses: actions-rs/cargo@v1
         with:
           command: build
           args: --release
 
+      - name: Build devnet
+        if: ${{ env.TAG_NAME }} == 'devnet'
```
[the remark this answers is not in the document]
zacharyfrederick: I think you can use `contains`. [Docs here ](https://docs.github.com/en/actions/learn-github-actions/expressions#contains)

So 
```
if: contains(${{ env.TAG_NAME }}, 'devnet') == true
```

zacharyfrederick:
@atodorov This should be a lot closer. Lmk what ya think

.github/workflows/release.yml
```diff
@@ -206,6 +214,22 @@ jobs:
 
           docker logout
 
+      - name: Build docker image Devnet
+        if: contains(${{ env.TAG_NAME }}, 'devnet') == true
+        run: |
+          docker build -t gluwa/creditcoin3:${{ env.TAG_NAME }} -f Dockerfile_Devnet .
```
atodorov, line 220: `Dockerfile_Devnet` doesn't exist and IMO it isn't needed. You can pass the build options to docker and use them as if using an environment variable. See for example:
https://github.com/gluwa/creditcoin3-staking-dashboard/blob/dev/.github/workflows/docker.yml#L43

.github/workflows/release.yml
```diff
@@ -152,13 +156,16 @@ jobs:
...
+        env:
+          BUILD_OPTIONS: --release ${{ needs.setup.outputs.build_options }}
         with:
           chain: "creditcoin"
           runtime_dir: runtime
           package: creditcoin3-runtime
           tag: ${{ env.RUSTC_VERSION }}
           workdir: ${{ github.workspace }}
 
+
```
atodorov, line 168: extra new line

.github/workflows/release.yml
```diff
@@ -152,13 +156,16 @@ jobs:
       - name: Build WASM runtime
         id: srtool_build
         uses: chevdor/srtool-actions@v0.8.0
+        env:
+          BUILD_OPTIONS: --release ${{ needs.setup.outputs.build_options }}
```
atodorov, line 160: 1). This job does not specify `setup` in its dependency specification. 

2) The name of the environment variable is incorrect -> https://github.com/chevdor/srtool-actions?tab=readme-ov-file#environmental-variables-and-build_opts

.github/workflows/release.yml
```diff
@@ -94,7 +98,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --release
+          args: --release ${{ needs.setup.outputs.build_options}}
```
atodorov, line 101: This job does not specify setup in its dependency specification.

.github/workflows/release.yml
```diff
@@ -9,6 +9,10 @@ on:
 permissions: read-all
 
 jobs:
+  setup:
+    runs-on: ubuntu-22.04
+    outputs:
+      build_options: ${{ contains( env.TAG_NAME, 'devnet') && '--features devnet' || '' }}
```
atodorov, line 15: I am not certain that this is going to work b/c environment variables are not propagated automatically between jobs as far as I know.  `TAG_NAME` is coming from `sanity-check` and we also define it independently inside every job that uses it. 

A better approach will be:

1) Make `TAG_NAME` an output of sanity-check, update all other jobs respectively b/c they depend on sanity-check anyway & remove the redundant steps redefining the ENV variable from the rest of the jobs.

2) Place `setup` after `sanity-check` and make it depend on it. That way you can consume 
`sanity-check.outputs.tag_name` inside `setup`.

atodorov (changes requested):

atodorov (changes requested):
See MegaLinter failures too https://github.com/gluwa/creditcoin3/actions/runs/8022463304/job/21916863418?pr=248

.github/workflows/release.yml
```diff
@@ -29,6 +32,12 @@ jobs:
       - name: Check tag name vs. Cargo.toml version
         run: |
           ./.github/check-version-vs-git-tag.sh
+  setup:
```
atodorov, line 35: The current definition is invalid, yields:
```
No steps defined in `steps` and no workflow called in `uses` for the following jobs: setup
```

See https://github.com/gluwa/creditcoin3/actions/runs/8052408066

It needs a dummy step which doesn't do anything useful.

.github/workflows/release.yml
```diff
@@ -11,6 +11,8 @@ permissions: read-all
 jobs:
   sanity-check:
     runs-on: ubuntu-22.04
+    outputs:
+      TAG_NAME_OUTPUT: ${{ steps.step2.output.TAG_NAME }}
```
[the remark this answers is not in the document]
atodorov: Typo in `outputs`

```suggestion
      TAG_NAME: ${{ steps.step2.outputs.TAG_NAME }}
```

atodorov (changes requested):

.github/workflows/release.yml
```diff
@@ -29,6 +32,12 @@ jobs:
       - name: Check tag name vs. Cargo.toml version
         run: |
           ./.github/check-version-vs-git-tag.sh
+  setup:
+    runs-on: ubuntu-22.04
+    needs: sanity-check
+    outputs:
+      build_options: ${{ contains( needs.sanity-check.outputs.TAG_NAME_OUTPUT, 'devnet') && '--features devnet' || '' }}
+      docker_build_options: ${{ contains(needs.sanity-check.outputs.TAG_NAME_OUTPUT, 'devnet') && 'BUILD_ARGS="--features devnet"'  || '' }}
```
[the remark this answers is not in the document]
zacharyfrederick: The dockerfile uses env variable `BUILD_ARGS` as the feature flag.  So I need to have that value dropped down to the the docker level and preserve the string. What is the smarter way to do it?

.github/workflows/release.yml
```diff
@@ -29,9 +33,23 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    needs: sanity-check
+    outputs:
+      build_options: ${{ contains( needs.sanity-check.outputs.TAG_NAME, 'devnet') && '--features devnet, fast-runtime' || '' }}
+      docker_build_options: ${{ contains(needs.sanity-check.outputs.TAG_NAME, 'devnet') && 'BUILD_ARGS="--features devnet, fast-runtime"'  || '' }}
+    steps:
+      - name: DEBUG
+        shell: bash
+        run: |
+          echo "Build options: ${{ contains( needs.sanity-check.outputs.TAG_NAME, 'devnet') && '--features devnet, fast-runtime' || '' }}"
```
atodorov, line 46: This `echo` isn't stable b/c it duplicates the expression above. The best way for this would be

1) add a step to set the otput vars
2) inside the DEBUG step echo the output vars from the previous step
3) inside the outputs section use the steps from the step defining the vars

Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | /bin/sh -s -- -y
 COPY --chown=creditcoin:creditcoin . /creditcoin-node/
 # shellcheck source=/dev/null
 RUN source ~/.cargo/env && \
-    cargo build --release
+    cargo build --release $BUILD_ARGS
```
atodorov, line 33: MegaLinter wants quotes here.

atodorov, line 33: Ideally also you add at the top of this file:

`ARG BUILD_ARGS=""` in case this isn't defined

atodorov (changes requested):

.github/workflows/release.yml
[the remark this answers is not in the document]
atodorov:
```
docker build -t gluwa/creditcoin3:${{needs.sanity-check.outputs.TAG_NAME  }} --build-arg BUILD_ARGS=${{ needs.setup.outputs.build_options }}".
```

.github/workflows/release.yml
```diff
@@ -29,9 +33,27 @@ jobs:
...
+      - name: Set build options
+        id: set-build-options
+        shell: bash
+        run: |
+          echo "build_options=${{ contains( needs.sanity-check.outputs.TAG_NAME, 'devnet') && '--features devnet, fast-runtime' || '' }}" >> "$GITHUB_OUTPUT"
+
+      - name: View the build options
+        shell: bash
+        run: |
+          echo "build_options = ${{ steps.set-build-options.outputs.build_options}}""
```
atodorov, line 51: Redundant double-quote at the end will cause bash & MegaLinter failures.

atodorov (changes requested):

Dockerfile
```diff
@@ -30,7 +31,7 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | /bin/sh -s -- -y
 COPY --chown=creditcoin:creditcoin . /creditcoin-node/
 # shellcheck source=/dev/null
 RUN source ~/.cargo/env && \
-    cargo build --release
+    cargo build --release "$BUILD_ARGS"
```
atodorov, line 34: hmmm, this causes build failures because of the quotes I think. 

You can try `${BUILD_ARGS}` syntax instead. If MegaLinter is still not happy we should disable the warnings it gives.

atodorov (changes requested):

Dockerfile
atodorov: FTR the `${BUILD_ARGS}` syntax works however MegaLinter still isn't happy so just update the list of ignored issues in Dockerfile.

.github/workflows/release.yml
```diff
@@ -29,9 +33,27 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    needs: sanity-check
+    outputs:
+      build_options: ${{ steps.set-build-options.outputs.build_options}}
+    steps:
+      - name: Set build options
+        id: set-build-options
+        shell: bash
+        run: |
+          echo "build_options=${{ contains( needs.sanity-check.outputs.TAG_NAME, 'devnet') && '--features devnet, fast-runtime' || '' }}" >> "$GITHUB_OUTPUT"
```
atodorov, line 46: The way `--features` is specified is not valid cargo syntax:

```
cargo build --release --features devnet, fast-runtime
error: unexpected argument 'fast-runtime' found

Usage: cargo build [OPTIONS]

For more information, try '--help'.
```

See https://github.com/gluwa/creditcoin3/actions/runs/8063126555/job/22024333191 from https://github.com/gluwa/creditcoin3/releases/tag/xx.yy.zz-devnet

It is either `--features "devnet fast-runtime"` with the quotes or specified twice like so:
`--features devnet --features fast-runtime`

.github/workflows/release.yml
```diff
@@ -192,15 +209,15 @@ jobs:
 
       - name: Build docker image
         run: |
-          docker build -t gluwa/creditcoin3:${{ env.TAG_NAME }} .
+          docker build -t gluwa/creditcoin3:${{needs.sanity-check.outputs.TAG_NAME  }} --build-arg="BUILD_ARGS=${{ needs.setup.outputs.build_options }}".
```
atodorov, line 212: Results in:

```
Run docker build -t gluwa/creditcoin3:xx.yy.zz-devnet --build-arg="BUILD_ARGS=--features devnet, fast-runtime".
ERROR: "docker buildx build" requires exactly 1 argument.
See 'docker buildx build --help'.
```

see https://github.com/gluwa/creditcoin3/actions/runs/8063126555/job/22024334310


Try:

```suggestion
docker build -t gluwa/creditcoin3:${{needs.sanity-check.outputs.TAG_NAME  }} --build-arg BUILD_ARGS="${{ needs.setup.outputs.build_options }}" .
```

atodorov (changes requested):
Changes requested b/c builds are failing:

PR used for debugging -> https://github.com/gluwa/creditcoin3/pull/251

Tag used for debugging -> https://github.com/gluwa/creditcoin3/releases/tag/xx.yy.zz-devnet

release.yml execution from the above tag -> https://github.com/gluwa/creditcoin3/actions/runs/8063126555

.github/workflows/release.yml
zacharyfrederick: that is the most embarrassing one of all

zacharyfrederick:
@atodorov Added the same build option setup logic to the wasm.yml CI file

.github/workflows/wasm.yml
```diff
@@ -11,8 +11,49 @@ on:
 permissions: read-all
 
 jobs:
+  sanity-check:
```
atodorov, line 14: All of this is actually not necessary and will probably even going to cause problems.

This CI job is used as part of regular PRs, not on tag pushes and will fail because you don't have tags on pull requests.

.github/workflows/wasm.yml
```diff
@@ -11,8 +11,49 @@ on:
...
+        id: identify-tag
+        shell: bash
+        run: |
+          TAG_NAME=$(git describe --tag)
+          echo "TAG_NAME=$TAG_NAME" >> "$GITHUB_ENV"
+          echo "TAG_NAME=$TAG_NAME" >> "$GITHUB_OUTPUT"
+
+      - name: Sanity check tag name suffix
+        run: |
+          echo "${{ env.TAG_NAME }}" | grep -E "mainnet|testnet|devnet"
```
atodorov, line 33: This will fail on merges to any of the target branches b/c the tag name will be incorrect anyway.

.github/workflows/wasm.yml
```diff
@@ -11,8 +11,49 @@ on:
...
+      - name: Set build options
+        id: set-build-options
+        shell: bash
+        run: |
+          echo "build_options=${{ contains( needs.sanity-check.outputs.TAG_NAME, 'devnet') && '--features devnet --features fast-runtime' || '' }}" >> "$GITHUB_OUTPUT"
+
+      - name: View the build options
+        shell: bash
+        run: |
+          echo "build_options = ${{ steps.set-build-options.outputs.build_options}}"
```
atodorov, line 50: this isn't really needed b/c this CI job isn't part of the release process.

.github/workflows/wasm.yml
```diff
@@ -11,8 +11,49 @@ on:
...
+    runs-on: ubuntu-22.04
+    needs: sanity-check
+    outputs:
+      build_options: ${{ steps.set-build-options.outputs.build_options}}
+    steps:
+      - name: Set build options
+        id: set-build-options
+        shell: bash
+        run: |
+          echo "build_options=${{ contains( needs.sanity-check.outputs.TAG_NAME, 'devnet') && '--features devnet --features fast-runtime' || '' }}" >> "$GITHUB_OUTPUT"
```
atodorov, line 45: If you want to be thorught and provide build_options only when running on `dev` or a PR against `dev` this aproach won't work b/c there are no tags here. 

Instead the approach should be similar to https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/runtime-upgrade.yml#L38 where values are configured depending on the target branch of a PR.  (github.base_ref)

WARNING: when running from a branch after a merge, not a PR then you don't have github.base_ref but instead      github.ref and the value is slightly different (see line 92 in this file).

.github/workflows/wasm.yml
```diff
@@ -30,6 +71,8 @@ jobs:
           package: creditcoin3-runtime
           workdir: ${{ github.workspace }}
           tag: ${{ env.RUSTC_VERSION }}
+        env:
+          BUILD_OPTS: "--release ${{ needs.setup.outputs.BUILD_ARGS }}"
```
atodorov: This is going to be fine for now b/c the only situation when this wasm output is consumed is in `upgrade-devnet` which is executed only when running on the `dev` branch (line 92).

atodorov (changes requested):
Let's drop the last changes to `wasm.yml` because they are opening a bigger can of worms!

.github/workflows/release.yml
```diff
@@ -152,13 +151,16 @@ jobs:
       - name: Build WASM runtime
         id: srtool_build
         uses: chevdor/srtool-actions@v0.8.0
+        env:
+          BUILD_OPTS: --release ${{ needs.setup.outputs.build_options }}
```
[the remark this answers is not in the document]
atodorov: Currently failing b/c of https://github.com/gluwa/creditcoin3/pull/243#discussion_r1502624717. We need #243 resolved/merged first before making progress here!

.github/workflows/wasm.yml
```diff
@@ -30,6 +30,8 @@ jobs:
           package: creditcoin3-runtime
           workdir: ${{ github.workspace }}
           tag: ${{ env.RUSTC_VERSION }}
+        env:
+          BUILD_OPTS: "--features devnet"
```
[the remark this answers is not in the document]
atodorov: **NOTE:** the comment above applies to the WASM build step in `release.yml`, not to `wasm.yml`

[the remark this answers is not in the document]
atodorov: https://github.com/gluwa/creditcoin3/commit/705910e62c9d39a07e0bd2245931dc3fe37d3a52

.github/workflows/release.yml
```diff
@@ -152,13 +164,16 @@ jobs:
       - name: Build WASM runtime
         id: srtool_build
         uses: chevdor/srtool-actions@v0.8.0
+        env:
+          BUILD_OPTS: --release "${{ needs.setup.outputs.build_options }}"
```
atodorov, line 168: Needs this patch: 

```suggestion
          BUILD_OPTS: "--release ${{ needs.setup.outputs.build_options }}"
```

https://github.com/gluwa/creditcoin3/commit/705910e62c9d39a07e0bd2245931dc3fe37d3a52

atodorov (changes requested):

Dockerfile
```diff
@@ -30,7 +31,7 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | /bin/sh -s -- -y
 COPY --chown=creditcoin:creditcoin . /creditcoin-node/
 # shellcheck source=/dev/null
 RUN source ~/.cargo/env && \
-    cargo build --release
+    cargo build --release ${BUILD_ARGS}
```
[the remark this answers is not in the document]
atodorov: @zacharyfrederick  the declaration of `BUILD_ARGS` needs to move into the appropriate build stage.

See:
- https://github.com/gluwa/creditcoin3/commit/640fdfd4ed735892b9db5426867cf4a68269524f
- https://github.com/gluwa/creditcoin3/actions/runs/8068323904/job/22040815572
- and logs
```
> [rust-builder 4/4] RUN source ~/.cargo/env &&     cargo build --release --features devnet --features fast-runtime:
```

atodorov:
This is ready for review (the Ready for review button currently results in an error)

- Build jobs from the `xx.yy.10-devnet` tag -> https://github.com/gluwa/creditcoin3/actions/runs/8085322225 all show the build options `--release --feature fast-runtime --feature devnet`
- [CI / integration-test-chainspec-devnet](https://github.com/gluwa/creditcoin3/actions/runs/8085237160/job/22093252857?pr=248) is a known failure, fixed in #249 
-

atodorov (approved):

PR #246, opened 2024-02-22
[CSUB-1062] Remove EVM wallet utils & evm-send command

pLabarta:
# Description of proposed changes

<describe what this PR is about and why we want it>

---

Practical tips for PR review & merge:

- [ ] All GitHub Actions report PASS
- [ ] Newly added code/functions have unit tests
  - [ ] Coverage tools report all newly added lines as covered
  - [ ] The positive scenario is exercised
  - [ ] Negative scenarios are exercised, e.g. assert on all possible errors
  - [ ] Assert on events triggered if applicable
  - [ ] Assert on changes made to storage if applicable
- [ ] Modified behavior/functions - try to make sure above test items are covered
- [ ] Integration tests are added if applicable/needed

PR #262, opened 2024-03-04
CSUB-1024:  Test cli commands with non-default --url

atodorov:

depends on #246.

---

### DIFF
--- a/cli/src/test/integration-tests/evm.test.ts
+++ b/cli/src/test/integration-tests/evm.test.ts
@@ -1,5 +1,5 @@
 import { signSendAndWatch } from '../../lib/tx';
-import { initAliceKeyring, ALICE_NODE_URL, randomFundedAccount, CLIBuilder } from './helpers';
+import { initAliceKeyring, ALICE_NODE_URL, BOB_NODE_URL, randomFundedAccount, CLIBuilder } from './helpers';
 import { ApiPromise, KeyringPair, newApi } from '../../lib';
 import { randomEvmAccount } from './evmHelpers';
 import { getEVMBalanceOf } from '../../lib/evm/balance';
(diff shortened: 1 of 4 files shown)


cli/src/test/integration-tests/evm.test.ts
```diff
@@ -55,29 +55,33 @@ describeIf(process.env.PROXY_ENABLED === undefined || process.env.PROXY_ENABLED
...
-            // requires the CC_SECRET set above
-            CLI(`evm withdraw`);
-
-            // Check that the caller's Substrate account balance is greater than initial
-            const afterBalance = await getBalance(caller.address, api);
-            expect(BigInt(afterBalance.transferable.toString())).toBeGreaterThan(
-                BigInt(parseAmount('1000000').toString()),
-            ); // Greater than initial balance
-        }, 60000);
+        it.each([`--url ${ALICE_NODE_URL}`, `--url ${BOB_NODE_URL}`])(
```
pLabarta, line 58: Q: Do we need to test `evm withdraw` with both nodes? In other tests, we are using the CLI through BOB and asserting chain status through ALICE.

cli/src/test/integration-tests/showAddress.test.ts
```diff
@@ -9,31 +9,35 @@ describeIf(
     process.env.PROXY_ENABLED === undefined || process.env.PROXY_ENABLED === 'no',
     'Show address command',
     () => {
-        it('should return the correct address', async () => {
-            await cryptoWaitReady();
+        it.each([`--url ${ALICE_NODE_URL}`, `--url ${BOB_NODE_URL}`])(
```
pLabarta, line 12: Same as with the previous comment on testing with both nodes.

cli/src/test/integration-tests/evm.test.ts
atodorov: > Do we need to test evm withdraw with both nodes? 

Kind of. The point of 1024 is to exercise all commands at least once via a non-default URL. 

because `evm withdraw` is called only once across the entire test suite there's no convenient place where I could just stick `--url ${BOB_NODE_URL}` that's why I have to repeat the scenario explicitly.

atodorov: >  other tests, we are using the CLI through BOB and asserting chain status through ALICE.

This is still the same, it doesn't change.

cli/src/test/integration-tests/showAddress.test.ts
atodorov: Same answer as above. `show-address` isn't used anywhere else so the scenario needs to be repeated

pLabarta (approved):

beqaabu (approved):

PR #276, opened 2024-03-19
chore: remove bootnodes from chainspecs

DylanVerstraete:

Removed hardcoded local bootnodes from chainspecs

---

- [x] All GitHub Actions report PASS
- [x] Modified behavior/functions - try to make sure above test items are covered

### DIFF
--- a/chainspecs/devnetSpecRaw.json
+++ b/chainspecs/devnetSpecRaw.json
@@ -2,9 +2,7 @@
   "name": "Creditcoin3 Dev",
   "id": "creditcoin3_dev",
   "chainType": "Live",
-  "bootNodes": [
-    "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWFEX4ncCfVv2VqCZLTaPoGTuXQR4pY3M2aGV1eGSqpmUQ"
-  ],
+  "bootNodes": [],
   "telemetryEndpoints": null,
   "protocolId": null,
   "properties": {
--- a/.github/check-bootnodes.sh
+++ b/.github/check-bootnodes.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# Directory containing JSON files
+chainspecs_dir="chainspecs"
+
+# Check if chainspecs directory exists
+if [ ! -d "$chainspecs_dir" ]; then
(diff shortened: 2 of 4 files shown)


chainspecs/devnetSpecRaw.json
```diff
@@ -2,9 +2,7 @@
   "name": "Creditcoin3 Dev",
   "id": "creditcoin3_dev",
   "chainType": "Live",
-  "bootNodes": [
-    "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWFEX4ncCfVv2VqCZLTaPoGTuXQR4pY3M2aGV1eGSqpmUQ"
-  ],
+  "bootNodes": [],
```
atodorov, line 5: Do you mind adding a check which makes sure that there are no boot nodes defined in any of the chain-specs ?

Option 1 - add it somewhere here: 
https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L91

can be a simple grep/bash script or maybe slightly better parsing the json with `jq`.

Option 2  - we have a dedicated section for chainspec checks which executes one of the existing integration tests with a different config file: 
https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L687


Either one will work but 1) should fail faster so is probably better.

atodorov (changes requested):
Can you squash both commits together ? No need to have bootnodes added and then removed in the history IMO.

chainspecs/devnetSpecRaw.json
DylanVerstraete: Implemented 1

beqaabu (dismissed):
Nice

atodorov (changes requested):
LGTM but the last commit isn't expected. I think this PR needs a rebase and the commit [Auto-update TypeScript type definitions](https://github.com/gluwa/creditcoin3/pull/276/commits/2b36411b249c3fddddba9b345da9ad09a4d70bb3) should be removed.

atodorov (approved):
LGTM, waiting for CI jobs to report green before merging.

beqaabu (approved):

PR #282, opened 2024-03-21
CSUB-1100: Halve the number of blocks/epoch for `fast-runtime` feature

atodorov:
this should speed up integration tests in CI b/c we'd have to wait less time for unbonded funds to be unlocked

# Description of proposed changes

<describe what this PR is about and why we want it>

---

Practical tips for PR review & merge:

- [ ] All GitHub Actions report PASS
- [ ] Newly added code/functions have unit tests
  - [ ] Coverage tools report all newly added lines as covered
  - [ ] The positive scenario is exercised
  - [ ] Negative scenarios are exercised, e.g. assert on all possible errors
  - [ ] Assert on events triggered if applicable
  - [ ] Assert on changes made to storage if applicable
- [ ] Modified behavior/functions - try to make sure above test items are covered
- [ ] Integration tests are added if applicable/needed

PR #284, opened 2024-03-21
CSUB-1006: Don't commit creditcoin.json if it is the only file that changed

atodorov:

extracted from #282 with some improvements.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -584,13 +584,18 @@ jobs:
           yarn format
 
           echo "========== DEBUG DEBUG DEBUG =========="
-          git status --short
+          git status --short --untracked-files=no
           git diff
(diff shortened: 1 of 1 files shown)


atodorov:
Sample execution from an older commit where version number was deliberately updated:
https://github.com/gluwa/creditcoin3/actions/runs/8388283753/job/22972601643?pr=284

- in step `Auto-update typedefs` we can see a git status/git diff showing that only `yaml.lock` and `creditcoin.json` have been updated, but no type definitions have been changed
- then step `Commit changes for typedefs` is not executed b/c there is nothing substantial to commit.

DylanVerstraete (approved):

beqaabu (approved):

zacharyfrederick:
chefs kiss

PR #287, opened 2024-03-25
Remove duplicate & unnecessary variables from cli code

atodorov:

---


cli/src/lib/balance/index.ts
```diff
@@ -1,10 +1,10 @@
-import { ApiPromise, parseUnits } from '..';
+import { ApiPromise, parseUnits, MICROUNITS_PER_CTC } from '..';
 import { BN } from '@polkadot/util';
 import Table from 'cli-table3';
 
 import type { DeriveStakingAccount } from '@polkadot/api-derive/types';
 
-export const MICROUNITS_PER_CTC = new BN('1000000000000000000');
+const MICROUNITS_AS_BN = new BN(MICROUNITS_PER_CTC.toString());
```
DylanVerstraete, line 7: Why not declare this in the `constants` file? Also, I think `1e18` is more readeable than `1_000_000_000_000_000_000`

atodorov: Originally I didn't want to have the same 2 names defined and exported in 2 different files and then while working on the cleanup I realized that we need 2 different data types. 

I didn't want to make too many syntax changes for what seems like a minor diff hence this new variable but I guess it ended up in lots of changes anyways. 

I don't have a strong opinion whether or not we add another name in consts. OTOH if we're going to do that it probably makes sense to convert MICROUNITS_PER_CTC into a BN and have only one variable used everywhere.  WDYT ?

DylanVerstraete (approved):

beqaabu (approved):

PR #327, opened 2024-04-22
Substrate transfer precompile

beqaabu:

---


atodorov (changes requested):
cherry-pick https://github.com/gluwa/creditcoin3/pull/326/commits/c1080bbeec07f6e9626d090a5d639ee70463be0b to resolve the failures reported by gitleaks

runtime/src/precompiles.rs
```diff
@@ -26,11 +27,13 @@ where
             hash(5),
             hash(1024),
             hash(1025),
+            hash(4049),
```
atodorov, line 30: Q: Why not sequential? 

Is there anything between 1026 and 4049?

cli/src/test/blockchain-tests/precompile.test.ts
```diff
@@ -0,0 +1,41 @@
+import { WebSocketProvider, ethers, parseEther } from 'ethers';
+import contractABI = require('./artifacts/SubstrateTransfer.json');
+import { Keyring } from '@polkadot/keyring';
+import { mnemonicGenerate } from '@polkadot/util-crypto';
+
+describe('Substrate seamless transfer precompile', (): void => {
+    let provider: WebSocketProvider;
+    let precompileContractAddress: string;
```
atodorov, line 8: these can be `const` inside the `beforeAll` block.

cli/src/test/blockchain-tests/precompile.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        const contract = new ethers.Contract(precompileContractAddress, contractABI, alith);
+
+        const target = new Keyring();
+        const pair = target.addFromMnemonic(mnemonicGenerate());
+
+        const amount = parseEther('10.0');
+        const gasPrice = (await provider.getFeeData()).gasPrice;
+
+        const result = await contract.transfer_substrate(pair.addressRaw, amount, {
```
atodorov, line 29: From here onwards is the "action + assert" part of the testing scenario. It should live inside the `test()` function.

The flow is "setup-act-assert-teardown".

Note: I see that you've probably copied this from `evm-tracing.test.ts` but it is wrong in there too. I'll follow up with a patch on that.

cli/src/test/blockchain-tests/precompile.test.ts
```diff
@@ -0,0 +1,41 @@
...
+            gasPrice,
+        });
+
+        receipt = await result.wait();
+        txHash = result?.hash;
+    }, 25000);
+
+    test('substrate_transfer', () => {
```
atodorov, line 37: preferably name matches the contract function name

precompiles/metadata/sol/substrate_transfer.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity >=0.8.3;
+
+/// @dev The Deposit precompile address
+address constant SUBSTRATE_TRANSFER_ADDRESS = 0x0000000000000000000000000000000000002049;
```
atodorov, line 5: Q: Integration test and `precompiles.rs` say "4049" while this one says `2049`! Is there some sort of a typo here ?

atodorov (changes requested):
Overall looks good, added a few questions and tidbits. 

@beqaabu can you also tag with the ticket number in subject.

cli/src/test/blockchain-tests/precompile.test.ts
atodorov: Update: I was wrong. `evm-tracing.test.ts` actually does look good, and follows the division of setup:act+assert:teardown correctly. Only needed a small update
 https://github.com/gluwa/creditcoin3/pull/329

precompiles/metadata/sol/substrate_transfer.sol
DylanVerstraete: this is indeed a typo, the address should be: `0x0000000000000000000000000000000000000fd1`

atodorov: Hmmm, curious how did this work then https://github.com/gluwa/creditcoin3/actions/runs/8789056637/job/24118598047?pr=327 ?

runtime/src/precompiles.rs
beqaabu: The inspiration for categorizing the precompiles was taken from Moonbeam. They are divided into 3 categories and they can be seen [here](https://docs.moonbeam.network/builders/pallets-precompiles/precompiles/overview/#precompiled-contract-addresses). 4049 I believe is arbitrary but it should be in the 3rd category (2048 - 4095) - Creditcoin specific precompiles.

precompiles/metadata/sol/substrate_transfer.sol
DylanVerstraete: Because this is probably a placeholder value to inform developers on what address it is deployed. It's not the real address, the real address is configured in the runtime

runtime/src/precompiles.rs
atodorov: IMO it makes sense to add comments behind each address to know what it is, but you can do that in a second step if you want. IDK what the other addresses here are used for.

cli/src/test/blockchain-tests/precompile.test.ts
beqaabu: fixed on the latest commit

cli/src/test/blockchain-tests/precompile.test.ts
beqaabu: fixed on the latest commit

runtime/src/precompiles.rs
beqaabu: Yes, makes sense to have it documented somewhere

atodorov: FTR: https://gluwa.atlassian.net/browse/CSUB-1142

cli/src/test/blockchain-tests/precompile.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+        const privateKey = (global as any).CREDITCOIN_EVM_PRIVATE_KEY('alice');
+        const alith = new ethers.Wallet(privateKey, provider);
+
+        contract = new ethers.Contract(precompileContractAddress, contractABI, alith);
+
+        const target = new Keyring();
+        pair = target.addFromMnemonic(mnemonicGenerate());
+
+        amount = parseEther('10.0');
```
atodorov, line 26: this is also part of the "act" step, move it inside the `test()` function.

cli/src/test/blockchain-tests/precompile.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        // precompile contract deployed at 4049 to hex, see runtime/src/precompiles.rs for more
+        const precompileContractAddress = '0x0000000000000000000000000000000000000fd1';
+
+        const privateKey = (global as any).CREDITCOIN_EVM_PRIVATE_KEY('alice');
+        const alith = new ethers.Wallet(privateKey, provider);
+
+        contract = new ethers.Contract(precompileContractAddress, contractABI, alith);
+
+        const target = new Keyring();
+        pair = target.addFromMnemonic(mnemonicGenerate());
```
atodorov, line 24: nitpick: maybe we can rename this to `destination` to be more clear?

cli/src/test/blockchain-tests/precompile.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+        amount = parseEther('10.0');
+        gasPrice = (await provider.getFeeData()).gasPrice;
+    }, 25000);
+
+    test('transfer_substrate', async () => {
+        const result = await contract.transfer_substrate(pair.addressRaw, amount, {
+            gasPrice,
+        });
```
atodorov, line 33: Q1: is the transfer call coming from `Alice / Alith` ?

cli/src/test/blockchain-tests/precompile.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        const receipt = await result.wait();
+        const txHash = result?.hash;
+        expect(txHash).toBeDefined();
+        expect(receipt).toBeDefined();
```
atodorov, line 37: this isn't saying much as to whether or not we were successfull. Best case we can assert on the recipient balance and make sure they got a +10 increase. See for example https://github.com/gluwa/creditcoin3/blob/dev/cli/src/test/integration-tests/evm.test.ts#L70-L71 and https://github.com/gluwa/creditcoin3/blob/dev/cli/src/test/integration-tests/withdraw-unbonded.test.ts#L218-L235 for inspiration.

cli/src/test/blockchain-tests/precompile.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    }, 25000);
+});
```
atodorov, line 39: - Needs a negative scenarion where the sender doesn't have enough free funds - should error, [evm] balances shouldn't change

- Needs a negative scenario where the amount to send it higher than the available funds - should error,  [evm] balances should not change

(not sure if the sender will pay gas in case of failure)

precompiles/substrate-transfer/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+        handle.record_log_costs_manual(3, 32)?;
+
+        // Build call with origin.
+        {
+            log::debug!("bytes: {:?}", destination);
+            let origin = Runtime::AddressMapping::into_account_id(handle.context().caller);
+
+            let to = Runtime::AccountId::from(destination.0);
+
+            let amount = Self::u256_to_amount(amount).in_field("value")?;
```
atodorov, line 60: Q: is this going to panic causing the node to crash when the input value is too high or is it going to result in a transaction error ?

precompiles/substrate-transfer/src/tests.rs
```diff
@@ -0,0 +1,130 @@
...
+                    PCall::transfer_substrate {
+                        destination: H256::zero(),
+                        amount: 200.into(),
+                    },
+                )
+                .execute_returns(true);
+
+            let bob: Account = bob_account.0.into();
+            let bob_balance = Balances::free_balance(bob);
+            assert_eq!(bob_balance, 0);
```
atodorov, line 100: These asserts are redundant, b/c bob isn't affected in any way here. 

If you want to assert I guess it should be that the balance for account with address zero has changed.

But then how is `H256::zero()` different from any other address if it isn't treated any differently. Looks like a bit of a dulicate scenario.

precompiles/substrate-transfer/src/tests.rs
```diff
@@ -0,0 +1,130 @@
...
+                    PCall::transfer_substrate {
+                        destination: bob_account,
+                        amount: 0.into(),
+                    },
+                )
+                .execute_returns(true);
+
+            let bob: Account = bob_account.0.into();
+            let bob_balance = Balances::free_balance(bob);
+            assert_eq!(bob_balance, 0);
```
atodorov, line 128: Bob starts with zero, then he's given zero and the expected is compared back to zero - not really a good way IMO to tell whether bob's balance wasn't changed or whether this didn't result in a change because of another reason. 

If Bob's starting balance is > 0 then you can eliminate other possible side effects gaining more influence that his balance didn't change.

precompiles/substrate-transfer/src/mock.rs
```diff
@@ -0,0 +1,259 @@
...
+        ext
+    }
+}
+
+// pub(crate) fn events() -> Vec<RuntimeEvent> {
+//     System::events()
+//         .into_iter()
+//         .map(|r| r.event)
+//         .collect::<Vec<_>>()
+// }
```
atodorov, line 259: Leftover code maybe ? Remove ?

precompiles/substrate-transfer/src/tests.rs
```diff
@@ -0,0 +1,130 @@
...
+                .execute_reverts(|output| {
+                    from_utf8(output)
+                        .unwrap()
+                        .contains("Dispatched call failed with error: ")
+                        && from_utf8(output).unwrap().contains("Arithmetic(Underflow)")
+                });
+
+            let bob: Account = bob_account.0.into();
+            let bob_balance = Balances::free_balance(bob);
+            assert_eq!(bob_balance, 0);
```
atodorov, line 72: would be better if Bob starts with non-zero balance

precompiles/substrate-transfer/src/tests.rs
atodorov: Q: should address zero be treated any differently ? E.g. not allowed ? 

Q2: Do Alice's funds diminish after sending to this address? I guess yes so we can assert on Alice's balance too.

precompiles/substrate-transfer/src/tests.rs
```diff
@@ -0,0 +1,130 @@
...
+            precompiles()
+                .prepare_test(
+                    alice,
+                    Precompile,
+                    PCall::transfer_substrate {
+                        destination: bob_account,
+                        amount: 0.into(),
+                    },
+                )
+                .execute_returns(true);
```
atodorov, line 124: Q: What is the point from a user POV to send zero amount ? It looks to me Alice will pay gas fees but not achieve anything else.  If that is the case shouldn't we allow transfers only when amount  > 0 ?

atodorov (changes requested):

cli/src/test/blockchain-tests/precompile.test.ts
beqaabu: yes

cli/src/test/blockchain-tests/precompile.test.ts
beqaabu: updated on the latest commit

precompiles/substrate-transfer/src/lib.rs
beqaabu: when the balance is insufficient it results in something like this `execution reverted: "Dispatched call failed with error: Arithmetic(Underflow)"`. I'm unsure what happens if amount doesn't fit in the type

precompiles/substrate-transfer/src/mock.rs
beqaabu: removed

precompiles/substrate-transfer/src/tests.rs
beqaabu: agreed, removed useless tests

atodorov: However what about these questions: 

1)  Should caller be allowed to call this when amount is zero ? Looks like they will just be burning gas fees instead of getting an error. Is that a valid use case ? 

2) Should caller be allowed to call this with destination address `0x000...000` effectively burning their funds ?

atodorov (commented):
Added an FYI comment, still reviewing.

precompiles/substrate-transfer/src/tests.rs
beqaabu: I'm unsure, not my call, but I think the user should technically be allowed to send the funds wherever they want

atodorov (dismissed):
LGTM.

Failing MegaLinter job is because of https://github.com/gluwa/creditcoin3/pull/330 - I've rebased which updated the commit hashes which invalidated `.gitleaksignore`. 
**Note:** depending on how we merge this PR (rebase, squash, etc) `.gitleaksignore` will most likely need to be updated one more time.

runtime/src/precompiles.rs
```diff
@@ -44,11 +47,18 @@ where
...
         }
     }
 
     fn is_precompile(&self, address: H160, _gas: u64) -> IsPrecompileResult {
+        let answer = Self::used_addresses().contains(&address);
+        log::debug!(
+            "Checking is address is precompile: {:?}, answer: {:?}",
+            address,
+            answer
+        );
```
DylanVerstraete, line 61: remove this, was added for debugging but it will spam the logs

DylanVerstraete (dismissed):
One comment

atodorov (approved):

DylanVerstraete (approved):

PR #425, opened 2024-07-30
Plugging in Chainspecs for third CC3 mainnet dry run

BradleyOlson64:


atodorov (changes requested):
- address the failures in MegaLinter job
- you may also run prettier against these JSON files to remove the warning
-

beqaabu (commented):
The chainspec itself looks good. I think we can overwrite the old chainspec with the new one and not have too many chainspec files present in the repo. Other than that pls fix the ci for megalinter and should be good to go

BradleyOlson64:
Having trouble pleasing Mega Linter. Prettier didn't work. 

It seems that mega linter is erroneously deciding that having a sudo public key in the chainspec isn't allowed. 

The line in question appears to be the exact same as in our other chainspecs.

dlebee (approved):

PR #435, opened 2024-08-05
CSUB-1231: Change Benchmarking script and Release builds to use --profile=production

BradleyOlson64:
Enacting the same changes in CC3 as were made in CCNext here https://github.com/gluwa/creditcoin3-next/pull/137 

As described [here](https://gluwa.atlassian.net/browse/CSUB-1231) using --release to build for benchmarking and for live network images/artifacts results in network performance loss due to missing optimizations.

Anecdotally I've observed that execution time weights are about 10% worse using --release vs --profile=production

This pr doesn't try to change all CI jobs to use --profile=production though, because compiling that way takes about 2x longer. We don't want to waste loads of money on unnecessary compilation work.


rustlang-dev (approved):

DylanVerstraete (approved):

.github/workflows/release.yml
```diff
@@ -111,22 +111,22 @@ jobs:
         uses: gluwa/cargo@dev
         with:
           command: build
-          args: --release ${{ needs.setup.outputs.build_options}}
+          args: --profile=production ${{ needs.setup.outputs.build_options}}
```
atodorov, line 114: I think the same should be applied in `wasm.yml` as well, otherwise looks good.

atodorov (changes requested):
Minor change requested.

.github/workflows/release.yml
BradleyOlson64: wasm.yml changed 👍

BradleyOlson64: Applying this change in `wasm.yml` created a rather confusing wasm build error in CI.

I'm not so sure this change is helpful, since `release.yml` appears to run its own independent wasm build. We can do without this right?

atodorov: > We can do without this right?

IDK tbh. For Testnet & Mainnet releases we typically use the wasm blob from the release, e.g. the once created in release.yml. For automated Devnet upgrades we use the one from wasm.yaml. Both are compiled in the same way minus a few feature flags.

atodorov: If I am reading the error logs at https://github.com/gluwa/creditcoin3/actions/runs/10288255898/job/28473252780 correctly it is a problem with how the profile information is specified, which I would think should be the same between the 2 yaml files but for some reason it isn't. 

Let's look into it tomorrow.

atodorov: FTR: experimenting in https://github.com/gluwa/creditcoin3/pull/437. Let's not make any more changes here until we know what's going on.

BradleyOlson64:
Closing this PR, as the team decided to include these changes upstream in CCNext and let them trickle down.

PR #437, opened 2024-08-08
CSUB-1231: Use --profile=production for all builds & CI jobs

atodorov:

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -167,8 +167,9 @@ jobs:
         id: srtool_build
         uses: chevdor/srtool-actions@v0.8.0
         env:
-          BUILD_OPTS: "--profile=production ${{ needs.setup.outputs.build_options }}"
+          BUILD_OPTS: "${{ needs.setup.outputs.build_options }}"
         with:
+          profile: production
           chain: "creditcoin"
           runtime_dir: runtime
           package: creditcoin3-runtime
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,7 +111,7 @@ jobs:
         uses: gluwa/cargo@dev
         with:
           command: build
-          args: --release ${{ needs.setup.outputs.build_options}}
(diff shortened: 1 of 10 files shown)


.github/workflows/release.yml
```diff
@@ -167,7 +167,7 @@ jobs:
         id: srtool_build
         uses: chevdor/srtool-actions@v0.8.0
         env:
-          BUILD_OPTS: "--release ${{ needs.setup.outputs.build_options }}"
+          BUILD_OPTS: "--profile=production ${{ needs.setup.outputs.build_options }}"
```
atodorov, line 170: @BradleyOlson64 as-is this change is not being exercised when a PR is created because it gets triggered by a git tag. IMO if https://github.com/gluwa/creditcoin3/actions/runs/10303559931/job/28519918476?pr=437 is failing this one will fail as well.

BradleyOlson64 (approved):
These changes look right. 

Just to make sure, we're fine with CI runs taking 1.5x longer than before right? 

Production builds take quite a bit longer than release builds so I was envisioning only using production for final image builds.

atodorov:
> These changes look right.
> 
> Just to make sure, we're fine with CI runs taking 1.5x longer than before right?
> 

Can you be more precise? Which jobs are taking 1.5x because I'm not able to spot anything obvious ? 


> Production builds take quite a bit longer than release builds so I was envisioning only using production for final image builds.

Multiple integration tests with the `--profile=production` build are currently failing, https://github.com/gluwa/creditcoin3/actions/runs/10305191630/job/28527968917?pr=437 and IMO the only way to have some confidence in these builds is to actually exercise them before we ship. Otherwise we end up in the situation where we test one target, however we ship a different target which is completely untested - not good IMO.

atodorov:
Runtime upgrade testing -> https://github.com/gluwa/creditcoin3/actions/runs/10317027713/job/28566440763 seems to me to be working fine.

The rest of the testing is in this PR which also appears to be looking good. Will port to creditcoin3-next and close this one shortly after!

PR #440, opened 2024-08-13
Final dry run specs

BradleyOlson64:
The specs had to be uploaded using Git lfs because they now contain balances for all .5 million accounts on the CC2 mainnet. They look like they were just removed here because GitHub doesn't provide large file visibility. 

You can still download the files from this branch to check them locally.


BradleyOlson64:
Closed in favor of fresh branch from dev here https://github.com/gluwa/creditcoin3/pull/441

PR #521, opened 2024-11-07
fix: update runtime's polynomial fn to adjust formula of WeightToFeeCoef

rustlang-dev:

update runtime's polynomial fn to adjust formula of  WeightToFeeCoef to make tx fees on cc3 the same as it is on cc2.

with a new updated formula the fees on cc3 and cc2 are:
```
transfers tx: 9.8mili vs 11.35mili
staking payout_stakers: 10.27mili vs 11.38mili
```

---

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -335,7 +335,6 @@ parameter_types! {
         Weight::from_parts(WEIGHT_REF_TIME_PER_NANOS.saturating_mul(125_000), 0);
 }
 
-pub const TARGET_FEE_CREDO: Balance = 10_000_000_000_000_000;
 pub struct WeightToCtcFee<T>(PhantomData<T>);
 
 impl<T: frame_system::Config> WeightToFeePolynomial for WeightToCtcFee<T> {
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -43,7 +43,7 @@ use frame_support::{
         OnTimestampSet,
     },
     weights::{
-        constants::WEIGHT_REF_TIME_PER_MILLIS, ConstantMultiplier, Weight, WeightToFeeCoefficient,
+        constants::WEIGHT_REF_TIME_PER_MILLIS, constants::WEIGHT_REF_TIME_PER_NANOS, ConstantMultiplier, Weight, WeightToFeeCoefficient,
     },
     PalletId,
 };
@@ -148,6 +148,8 @@ macro_rules! prod_or_fast {
 pub const MILLISECS_PER_BLOCK: u64 = prod_or_fast!(15_000, 5_000);
 
 pub const SLOT_DURATION: u64 = MILLISECS_PER_BLOCK;
+
+pub const MILLI_CTC: u128 = 1_000_000_000_000_000;
 
 // WARNING: the next should be defined on a single line b/c of
 // the assertions made in .github/check-for-changes-in-epoch-duration.sh
@@ -326,6 +328,12 @@ impl pallet_balances::Config for Runtime {
     type MaxFreezes = ();
 }
 
+parameter_types! {
+    /// Executing a NO-OP `System::remarks` Extrinsic.
+    pub const ExtrinsicBaseWeight: Weight =
+        Weight::from_parts(WEIGHT_REF_TIME_PER_NANOS.saturating_mul(125_000), 0);
+}
+
 pub const TARGET_FEE_CREDO: Balance = 10_000_000_000_000_000;
 pub struct WeightToCtcFee<T>(PhantomData<T>);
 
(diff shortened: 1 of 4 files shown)


beqaabu:
The check version job is failing because if we modify the runtime we also need to bump the version, this release should be v3.36.0. See the job here https://github.com/gluwa/creditcoin3/actions/runs/11730824518/job/32679537016?pr=521

I'm not sure why the `danger-will-brick-the-blockchain` job is failing https://github.com/gluwa/creditcoin3/actions/runs/11730824518/job/32679537942?pr=521 because I don't see any changes in epoch duration in this pr, maybe @atodorov can shed some light on this?

Other 3 CI jobs are known to fail, so I think its okay. I'm testing the changes locally and will let you know how it all works out

DylanVerstraete (dismissed):
LGTM 👍 But indeed bump the version for runtime upgrade

atodorov:
> I'm not sure why the `danger-will-brick-the-blockchain` job is failing

b/c it detects `MILLISECS_PER_BLOCK` as part of the git diff context.  This is a false positive in this particular case and can be ignored.

runtime/src/lib.rs
```diff
@@ -326,28 +329,27 @@ impl pallet_balances::Config for Runtime {
     type MaxFreezes = ();
 }
 
+parameter_types! {
+    /// Executing a NO-OP `System::remarks` Extrinsic.
+    pub const ExtrinsicBaseWeight: Weight =
+        Weight::from_parts(WEIGHT_REF_TIME_PER_NANOS.saturating_mul(125_000), 0);
+}
+
 pub const TARGET_FEE_CREDO: Balance = 10_000_000_000_000_000;
```
atodorov, line 338: Now unused so can be removed.

runtime/src/lib.rs
```diff
@@ -149,6 +150,8 @@ pub const MILLISECS_PER_BLOCK: u64 = prod_or_fast!(15_000, 5_000);
 
 pub const SLOT_DURATION: u64 = MILLISECS_PER_BLOCK;
 
+pub const MILLI_CTC: u128 = 1_000_000_000_000_000;
```
atodorov, line 153: Does it need to be public ?

atodorov (commented):
Minor comments. Needs version bump.

runtime/src/lib.rs
```diff
@@ -326,28 +329,27 @@ impl pallet_balances::Config for Runtime {
...
-
-        let base = Balance::from(deal_order_weight.ref_time());
-        let ratio = TARGET_FEE_CREDO / base;
-        let rem = TARGET_FEE_CREDO % base;
-        smallvec::smallvec!(WeightToFeeCoefficient {
-            coeff_integer: ratio,
-            coeff_frac: Perbill::from_rational(rem, base),
-            negative: false,
+        let p = MILLI_CTC / 2;
+        let q = 100 * Balance::from(ExtrinsicBaseWeight::get().ref_time());
```
beqaabu, line 346: where do the values from `p` and `q` come from?

runtime/src/lib.rs
```diff
@@ -326,28 +329,27 @@ impl pallet_balances::Config for Runtime {
     type MaxFreezes = ();
 }
 
+parameter_types! {
+    /// Executing a NO-OP `System::remarks` Extrinsic.
+    pub const ExtrinsicBaseWeight: Weight =
+        Weight::from_parts(WEIGHT_REF_TIME_PER_NANOS.saturating_mul(125_000), 0);
+}
```
beqaabu, line 336: I checked a couple of different implementations from different projects, acala for example uses `ExtrinsicBaseWeight` from `frame_support::weights::constants` https://github.com/AcalaNetwork/Acala/blob/613470c47c65fb68e1de79e8991453d12c249740/runtime/acala/src/constants.rs#L44
do we need our own constant or can it be reused from somewhere else?

runtime/src/lib.rs
rustlang-dev: does `SLOT_DURATION` should not be public? 

Why you asked about MILLI_CTC and not about `SLOT_DURATION` const that was there way before?

runtime/src/lib.rs
rustlang-dev: from polkadot sdk parachain template.

runtime/src/lib.rs
rustlang-dev: done.

runtime/src/lib.rs
atodorov: > does `SLOT_DURATION` should not be public?
> 

Probably. IDK if there is a specific reasoning why some of these constants are public and some others aren't. The same rule should probably apply to all constants in this file. Opened CSUB-1360 to deal with that.

> Why you asked about MILLI_CTC and not about `SLOT_DURATION` const that was there way before?

I asked about `MILLI_CTC` b/c that's what is currently the new change which made me question whether these constants being public is actually needed or not. Thought you may know.

runtime/src/lib.rs
beqaabu: okay got it, I still think we should reuse the constant from `frame_support` instead of introducing our own because the value is almost identical (125_000 vs 124_414) and we wont have a random constant lying around.

runtime/src/lib.rs
beqaabu: I guess depending on how close we want to get to cc2 fees we can modify the multiplier 100 to different numbers but the fees are already really close, 11.2 milli ctc on cc2 vs 10.3 milli ctc on cc3 in my tests

atodorov (approved):

runtime/src/lib.rs
RustNinja: no need to be public the same as other consts.
but to have the same format as other const i declare with pub as other consts

atodorov: ^^^ this should be addressed in CSUB-1360. The fact that something is the same doesn't mean it is necessarily correct.

atodorov:
Merging, see https://gluwa.slack.com/archives/C03MQ532BGA/p1731078810606819

PR #545, opened 2025-01-21
fixed fees

beqaabu:

---


Cargo.lock
```diff
@@ -392,29 +392,30 @@ dependencies = [
 
 [[package]]
 name = "ark-scale"
-version = "0.0.10"
+version = "0.0.12"
```
atodorov, line 395: Similar question as in https://github.com/gluwa/creditcoin3/pull/546#discussion_r1923746447. 

Why do we need this change ? 

From what I can see these are only dependency upgrades, but this commit is not in `dev` & `testnet` branches and will mostl likely cause conflicts with commits from Dependabot in the future.

beqaabu: this change happened after I built and tested the binary locally, it also updated the cargo.lock file. will revert the last commit

atodorov (approved):

PR #546, opened 2025-01-21
Merge devnet into testnet

beqaabu:

---


Cargo.lock
```diff
@@ -687,16 +700,19 @@ dependencies = [
 
 [[package]]
 name = "bandersnatch_vrfs"
-version = "0.0.4"
-source = "git+https://github.com/w3f/ring-vrf?rev=0fef826#0fef8266d851932ad25d6b41bc4b34d834d1e11d"
+version = "0.0.1"
+source = "git+https://github.com/w3f/ring-vrf?rev=3119f51#3119f51b54b69308abfb0671f6176cb125ae1bf1"
```
atodorov, line 704: Hmmm, this is suspicious. 2 questions:

1) Why do we need to update Cargo.lock here ? It is already updated as part of the previous commit which bumps the version.  What did you do to introduce these changes ? 

2) This particular change is a downgrade of this library which is probably not what is intended.

atodorov (changes requested):

Cargo.lock
beqaabu: I just did a simple rebase. I dont understand how something like this happens because this dependency is v0.0.1 on both dev and testnet branches
dev - https://github.com/gluwa/creditcoin3/blob/d5ba556ba4b5e1465a7ad7886f07ede8027f299a/Cargo.lock#L688-L689

testnet - https://github.com/gluwa/creditcoin3/blob/887c639fa2776f3316ce77c82b44e29833c65cb1/Cargo.lock#L688-L689

atodorov: The change to v0.0.4 was introduced in c15aaf7543c1cab09314999f92db89b4b1daa302 (1st commit in this PR). 

Then it was overwritten in the last commit. 

Most likely both are results of running `cargo build` locally and committing the changes to Cargo.lock.   The fact that we ended up with the same version 0.0.1 is just a coincidence.

beqaabu: okay makes sense. reverted the last commit, should be better now

PR #576, opened 2025-03-11
Update default base fee

beqaabu:

Updates the default base fee per gas from 0.5 gwei to 700 gwei

---


DylanVerstraete (approved):

Cargo.toml
```diff
@@ -1,14 +1,14 @@
 [workspace]
 members = [
-	"node",
-	"runtime",
-	"runtime/generate-bags",
-	"precompiles/substrate-transfer",
+  "node",
+  "runtime",
+  "runtime/generate-bags",
+  "precompiles/substrate-transfer",
```
atodorov, line 6: Can you adjust your editor to use tabs instead of spaces? 

This has introduced lots of unnecessary formatting changes for pretty much every single toml file.

runtime/src/lib.rs
```diff
@@ -443,7 +443,7 @@ impl pallet_dynamic_fee::Config for Runtime {
 }
 
 parameter_types! {
-    pub DefaultBaseFeePerGas: U256 = U256::from(1_000_000_000);
+    pub DefaultBaseFeePerGas: U256 = U256::from(1_400_000_000_000_u128);
```
atodorov, line 446: Curious if this can't just be 1_400_000_000_000_u256

atodorov (changes requested):
If you can just deal with the tabs vs spaces issue that would be great.

Cargo.toml
beqaabu: its not ide, its taplo in the recommit hooks,and it wouldn't let me push without fixing the format

atodorov: I see now.

Here's what happened:

b23e2e9ad4954509abb71da7bf1f0c9c5342f49d makes taplo execute inside the pre-commit hook only if toml files have been updated, however I suspect that till now noone who had updated such files had the pre-commit hook installed on their repository.

on creditcoin3-next we also have 3afb55ce2 which runs taplo format in CI, but not here.

That's why it was missed and popped up here. 

Still on the creditcoin3-next repository taplo added 4 spaces instead of a tab. Here it added only 2. Weird but not a blocker.

atodorov (approved):

PR #579, opened 2025-03-12
Execute chain-snapshots testing only on Monday & Thursday

atodorov:
b/c devops will be making new snapshots every 3 days so there's no need to execute this consistently every day


beqaabu (approved):

atodorov:
Results for latest cc3-testnet snapshot -> https://github.com/gluwa/creditcoin3/actions/runs/13810232951/job/38630119408

Note: CC3 Mainnet & CC3 Devnet snapshots are 1 month old and still in the older format, thus failing.

PR #592, opened 2025-04-25
add frame-metadata-hash-extension and configure for runtime.

rustlang-dev:

add frame-metadata-hash-extension and configure for runtime.

---


runtime/build.rs
```diff
@@ -1,8 +1,21 @@
+#[cfg(all(feature = "std", not(feature = "metadata-hash")))]
```
beqaabu, line 1: If I recall correctly, we discussed yesterday that we don't want extra feature flags. Can you make `metadata-hash` enabled by default and remove the flag altogether?

beqaabu (changes requested):

beqaabu:
Also fix the ci please

atodorov (changes requested):
Please tend to the failing CI jobs:

- There are some compilation failures, see https://github.com/gluwa/creditcoin3/actions/runs/14664751971/job/41156820737?pr=592 and https://github.com/gluwa/creditcoin3/actions/runs/14664751971/job/41156825386?pr=592

- When changing anything under `runtime/` we require a version bump, see https://github.com/gluwa/creditcoin3/actions/runs/14664751976/job/41156816237?pr=592. Ideally that would be a separate commit at the end of this PR.

atodorov (changes requested):
- Failure for [docs-smart-contract-development-with-hardhat](https://github.com/gluwa/creditcoin3/actions/runs/14667710033/job/41167468392?pr=592#logs) is unrelated (account run out of funds) and can be ignored. 

- Failure for [integration-test-blockchain](https://github.com/gluwa/creditcoin3/actions/runs/14667710033/job/41167467823?pr=592#logs) however appears quite serious:

```
FAIL src/test/blockchain-tests/substrate-transfer.test.ts (8.118 s)
  Precompile: transfer_substrate()
    ✕ should work when caller has enough funds (1 ms)
    ✕ should fail when sending more than total issuance
    ✕ should fail when sending more than available funds (1 ms)

  ● Precompile: transfer_substrate() › should work when caller has enough funds

    RpcError: 1002: Verification Error: Runtime error: Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed
    WASM backtrace:
    error while executing at wasm backtrace:
        0: 0x305672 - creditcoin3_runtime.wasm!rust_begin_unwind
        1: 0xfc74 - creditcoin3_runtime.wasm!core::panicking::panic_fmt::hfeb332030c845b66
        2: 0x27081f - creditcoin3_runtime.wasm!TaggedTransactionQueue_validate_transaction: RuntimeApi, Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed\nWASM backtrace:\nerror while executing at wasm backtrace:\n    0: 0x305672 - creditcoin3_runtime.wasm!rust_begin_unwind\n    1: 0xfc74 - cre…
```

This looks like a panic when we try to execute a precompile and appears to be related to the wasm runtime. I don't know enough about why enabling this new pallet would cause this issue but this is something that we've not seen before and needs more investigation otherwise we risk breaking the chain.

atodorov:
FYI moved to Draft so we don't merge this accidentally. Talked to @DylanVerstraete and we'll attempt CSUB-1577 first, before coming back to this.

BradleyOlson64 (dismissed):
LGTM 👍

atodorov (changes requested):
You will have to update how we build release & testing artifacts.

1) In release.yaml add the new feature flag:
  - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/release.yml#L116
  - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/release.yml#L124
  - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/release.yml#L174
  - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/release.yml#L217
 2) in wasm.yaml - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/wasm.yml#L36
3) in ci.yaml
  - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L370
  - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L374
  - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L454
  - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L517
  -

beqaabu:
> You will have to update how we build release & testing artifacts.
> 
> 1. In release.yaml add the new feature flag:
> 
> * here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/release.yml#L116
> * here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/release.yml#L124
> * here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/release.yml#L174
> * here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/release.yml#L217
> 
> 2. in wasm.yaml - here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/wasm.yml#L36
> 3. in ci.yaml
> 
> * here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L370
> * here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L374
> * here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L454
> * here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L517
> * and here https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L33

I updated the dockerfile https://github.com/gluwa/creditcoin3/pull/592/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R23 to include the feature flag so I don't think changes to
https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/ci.yml#L33 and https://github.com/gluwa/creditcoin3/blob/dev/.github/workflows/release.yml#L217 are needed anymore

Dockerfile
```diff
@@ -20,7 +20,7 @@ COPY --chown=creditcoin:creditcoin . /creditcoin-node/
 
 
 FROM devel-base as rust-builder
-ARG BUILD_ARGS=""
+ARG BUILD_ARGS="--features metadata-hash"
```
atodorov, line 23: This isn't quite going to work in all situations. 

This is a default value which will work when doing `docker build` like in ci.yml line 33. 

However in release.yml 217 this value is explicitly set from the outside and is overriden effectively removing the feature. 

This compiler flag really needs to be specified everywhere. (please also add it in ci.yml 33 to make it explicit so we don't forget about how it works).

atodorov (changes requested):

Dockerfile
DylanVerstraete: We don't need this feature by default

.github/workflows/ci.yml
DylanVerstraete, line 1: not needed here to pass this flag

DylanVerstraete (commented):
See my reply for Dockerfile as well, otherwise good to go

.github/workflows/ci.yml
atodorov: We would like to make sure testing uses all features, that's why passing this flag everywhere test artifacts are built and used.  Otherwise issues may linger undetected until last minute.

Dockerfile
atodorov: I agree that we don't need this enabled by default. 

For building testing artifacts (as in ci.yml line 33) we can pass this explicitly to make sure this is actually exercised.

DylanVerstraete: We don't test ledger features, so this just adds build overhead

DylanVerstraete (approved):

atodorov (approved):
Looks good assuming nothing else fails.

beqaabu:
dead link in   
`Errors in testing/package-lock.json`
[500] https://github.com/request/request/issues/3142 | Network error: Internal Server Error
https://github.com/gluwa/creditcoin3/actions/runs/14886646675/job/41807917263?pr=592

BradleyOlson64 (approved):
LGTM 👍 

Just making sure, we've tested building with and without the metadata hash flag right?

beqaabu:
> LGTM 👍
> 
> Just making sure, we've tested building with and without the metadata hash flag right?

yup, they both work and we don't even break basic functionality (aside from ledger) even if we upgrade the runtime without the feature flag

DylanVerstraete (approved):

beqaabu (approved):

PR #643, opened 2025-08-11
Auto merge 2025-08-11/merge-dev-into-testnet to testnet

atodorov:


beqaabu (approved):

atodorov:
> Isn't it weird it would use your personal token?

If you use the default token GitHub recognizes that it is a bot account and will not schedule CI jobs to avoid going into an endless loop. We can substitute the current token with another one as long as it is a personal access token.

PR #686, opened 2025-11-14
CSUB-1852: migrations

beqaabu:

---

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -975,6 +975,7 @@ pub type Executive = frame_executive::Executive<
     frame_system::ChainContext<Runtime>,
     Runtime,
     AllPalletsWithSystem,
+    // Can be removed after the migration is applied
     (
         pallet_identity::migration::versioned::V0ToV1<Runtime, { u64::MAX }>,
         pallet_grandpa::migrations::MigrateV4ToV5<Runtime>,
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -80,6 +80,8 @@ use pallet_transaction_payment::Multiplier;
 mod precompiles;
 use precompiles::FrontierPrecompiles;
 
+mod migrations;
+
 /// Type of block number.
 pub type BlockNumber = u32;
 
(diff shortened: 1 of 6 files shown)


runtime/src/lib.rs
```diff
@@ -973,7 +975,12 @@ pub type Executive = frame_executive::Executive<
     frame_system::ChainContext<Runtime>,
     Runtime,
     AllPalletsWithSystem,
-    (pallet_identity::migration::versioned::V0ToV1<Runtime, { u64::MAX }>,),
+    // Can be removed after the migration is applied
+    (
+        pallet_identity::migration::versioned::V0ToV1<Runtime, { u64::MAX }>,
+        pallet_grandpa::migrations::MigrateV4ToV5<Runtime>,
+        migrations::staking_v13_to_v15::MigrateV13ToV15<Runtime>,
+    ),
```
atodorov, line 983: Correct me if I am wrong but I belive that if we enable the try-runtime feature and copy the structure from creditcoin2 these would be applied automatically and we don't have to specify them explicitly in the runtime. 

We would only have to specify migrations for our own pallets I believe (see https://github.com/gluwa/creditcoin/tree/dev/pallets/creditcoin/src/migrations).  As a side bonus that older structure allowed quite a lot of testing to be performed against these migrations. WDYT ?

DylanVerstraete: I'm not sure that is how it works @atodorov , try-runtime only tries to apply the specified runtime upgrades.

beqaabu: try-runtime is just a testing tool afaik, we would still need to manually specify which pallets to migrate, our own or third party. the same was true for cc2 aswell I think https://github.com/gluwa/creditcoin/blob/8ebb9672ef82f88f0c3e87192932974cfe868a32/runtime/src/lib.rs#L706-L713

didac-gluwa (approved):

runtime/src/lib.rs
dlebee: are those meant to be just the staking migrations? what about pallet_evm? we noticed that it lacked the version in the chain state? when you ran the migrations it had a trace and that one had a warning during the check.

beqaabu: These migrations are for staking and grandpa. Identity can be removed because its already migrated but we can remove all 3 of those together once we migrate. As for pallet_evm, it doesnt have any migrations, pallet version on our chain, in code and across other chains is 0

DylanVerstraete (approved):

PR #704, opened 2025-12-22
CSUB-1905: Reduce frequency of chain snapshots testing

atodorov:
Devnet - once a month
Testnet - every 2 weeks
Mainnet - every week


atodorov:
Merging b/c I need to be able to exercise this in order to check if it all works as expected.

PR #710, opened 2026-01-06
CSUB-1905: Adjust biweekly cron syntax for Testnet snapshots

atodorov:
the previous syntax was wrong and resulted in jobs being scheduled every day between 1-7 of the month! This new syntax should schedule on every 14th day of the month!

### DIFF
--- a/.github/workflows/check-testnet-snapshot.yml
+++ b/.github/workflows/check-testnet-snapshot.yml
@@ -3,8 +3,8 @@ name: Check Testnet Snapshot
 
 on:
   schedule:
-    # at 8:30, on Tusday, 1st and 3rd week of the month
-    - cron: "30 8 1-7,15-21 * 2"
+    # at 08:30 on every 14th day-of-month
+    - cron: "30 8 */14 * *"
   workflow_dispatch:
 
 permissions: read-all


DylanVerstraete (approved):

.github/workflows/check-testnet-snapshot.yml
```diff
@@ -3,8 +3,8 @@ name: Check Testnet Snapshot
 
 on:
   schedule:
-    # at 8:30, on Tusday, 1st and 3rd week of the month
-    - cron: "30 8 1-7,15-21 * 2"
+    # at 08:30 on every 14th day-of-month
+    - cron: "30 8 */14 * *"
```
Trantorian1, line 7: Just to make sure I understand correctly, the syntax means this action will be running every 14 days right?

cf: https://crontab.guru/#30_8_*/14_*_*

If the goal was to run on the 14th of each month (once a month)  then we should use `30 8  14 * *` instead.

cf: https://crontab.guru/#30_8_14_*_*

atodorov: > Just to make sure I understand correctly, the syntax means this action will be running every 14 days right?

Yes. Aka every 2 weeks.

Trantorian1 (approved):

PR #787, opened 2026-04-09
CSUB-1974: Offline checkpoint builder + checkpoint verifier

DylanVerstraete:
This PR contains two tools:

* `checkpoint-builder`: Reads a sled database and creates checkpoints from it based on config
* `checkpoint-validator` Creates a checkpoint from a block range fetching blocks and receipts directly from rpc


attestor/scripts/README.md
```diff
@@ -115,3 +115,64 @@ No special steps here. Simply register and run attestors as normal for the chain
 imported checkpoints to.
 
 The resulting attestor sync process should take a much shorter time than it otherwise would.
+
+## Importing Checkpoints from a CSV File
+
+`ImportCheckpointsFromCsv.js` is an alternative to the export-then-import workflow above.
+Use it when you already have checkpoints in a CSV file (e.g. produced by a checkpoint-builder
+tool) and want to import them directly into a chain without needing a live source chain to
+export from.
```
atodorov: Q: Why can't we have this tool create checkpoints in the same format as ExportCheckpoints.js ? 

Or maybe replace the export script with this tool ?

checkpoint-verifier/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "cli"
+    ],
+    "scripts": {
+        "build": "tsc",
+        "start": "node dist/cli.js",
+        "format": "prettier --write \"src/**/*.ts\"",
+        "check-format": "prettier --check \"src/**/*.ts\"",
+        "lint": "eslint -c .eslintrc.js --max-warnings 0 --ext .ts ./src",
+        "typecheck": "tsc --noEmit",
+        "prepare": "yarn build"
```
atodorov, line 27: Looks like this script dir should be added in typescript-checks.yml

checkpoint-verifier/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Overview
+
+The Checkpoint Verifier performs the following:
+
+1. Connects to an Ethereum-compatible JSON-RPC endpoint (HTTP or WebSocket)
+2. Fetches block data and transaction receipts for requested blocks
+3. Computes merkle roots using the same algorithm as the attestation system
+4. Chains checkpoint digests using `keccak256(blockNumber || merkleRoot || prevDigest)`
+5. Either outputs computed checkpoints or verifies them against a CSV file
```
atodorov, line 13: Both scenarios (output + verify) must be exercised as a CI job. 

See integration-test-attestator-network: - might be a good place to plug this in, or alternatively to copy the setup and ecercise it as a separate CI job.

checkpoint-verifier/README.md
```diff
@@ -0,0 +1,197 @@
...
+  --start-block 101 \
+  --end-block 200 \
+  --starting-digest 0x1234567890abcdef...
+
+# With progress output
+yarn start verify-range \
+  --rpc-url http://localhost:8545 \
+  --start-block 0 \
+  --end-block 1000 \
+  --verbose
```
atodorov, line 59: All of these 3 seem like good candidates to be executed as separate CI steps.

atodorov (changes requested):
Added comment wrt testing this new tool.

attestor/scripts/README.md
didac-gluwa: Replaced and unified scripts so now we only export and import the csv format!

checkpoint-verifier/README.md
didac-gluwa: Added them to `integration-test-attestator-network`

checkpoint-verifier/README.md
didac-gluwa: The work in `integration-test-attestator-network` should cover that

.github/workflows/ci.yml
```diff
@@ -401,23 +415,121 @@ jobs:
...
+          if [ ! -s "$OVERLAP" ]; then
+            echo "No overlapping block numbers between builder CSV and on-chain CSV" >&2
+            exit 1
+          fi
+
+          diff -u "$ONCHAIN" "$OVERLAP"
+
+      - name: Build checkpoint-verifier
+        if: always()
+        working-directory: ./checkpoint-verifier
```
atodorov, line 513: Since you will need a mandatory rebase, see https://github.com/gluwa/creditcoin3/pull/870/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR336 and add this location to cache-dependency-path for this CI job.

didac-gluwa: I've updated only the `integration-test-attestator-network` to use caching and adjusted install commands as recommended here: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#working-with-lockfiles

checkpoint-verifier/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
+    "semi": true,
+    "trailingComma": "all",
+    "singleQuote": true,
+    "printWidth": 120,
+    "tabWidth": 4,
+    "bracketSpacing": true
+}
```
atodorov: Duplicate: https://github.com/gluwa/creditcoin3/actions/runs/25001387194/job/73212180804?pr=787

if you insist on having the file, not a symlink then maybe add a dummy entry so that it's different.

atodorov (changes requested):
Minor issue

checkpoint-builder/src/sink/csv.rs
```diff
@@ -0,0 +1,152 @@
...
+use tracing::{debug, error, info};
+
+use attestor_primitives::AttestationCheckpoint;
+
+use crate::sink::CheckpointSink;
+
+/// CSV file sink for writing checkpoint data in CSV format
+/// Writes data as: block_number,digest (no headers)
+///
+/// The writer is lazily initialized on first write and reused for subsequent writes.
```
Trantorian1, line 17: Is there a benefit to lazy-initialization in this case? Why not just initialize the writer upfront? This removes any initialization check after `CsvFileSink::new` and simplifies the write flow.

checkpoint-builder/src/sink/csv.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        let buff = self.writer.as_mut().expect("writer should be initialized");
+
+        // Build CSV records and write them
+        for checkpoint in &checkpoints {
+            let formatted_digest = format!("0x{}", hex::encode(checkpoint.digest.as_bytes()));
+            let line = format!("{},{}\n", checkpoint.block_number, formatted_digest);
+            buff.write_all(line.as_bytes())
+                .with_context(|| "Failed to write checkpoint to CSV file")?;
+        }
```
Trantorian1, line 64: It probably won't make much difference for a tool which is used quite rarely, and is likely already quite fast as is, however re-using the `formattted_digest` and `line` strings instead of re-allocating on each iteration should help improve write speeds a lot.

Trantorian1: You can use [`String::clear`](https://doc.rust-lang.org/stable/std/string/struct.String.html#method.clear) for this purpose.

checkpoint-builder/src/sink/csv.rs
```diff
@@ -0,0 +1,152 @@
...
+    output_file: PathBuf,
+    commit_interval: usize,
+) -> tokio::task::JoinHandle<Result<()>> {
+    // Add timestamp to output filename
+    let timestamp = Utc::now().format("%Y%m%dT%H%M%S");
+    let timestamped_file = if let Some(stem) = output_file.file_stem() {
+        let extension = output_file
+            .extension()
+            .map(|e| format!(".{}", e.to_string_lossy()))
+            .unwrap_or_default();
```
Trantorian1, line 96: Can probably just assume `.csv`

checkpoint-builder/src/sink/csv.rs
```diff
@@ -0,0 +1,152 @@
...
+        let new_name = format!("{}_{}{}", stem.to_string_lossy(), timestamp, extension);
+        output_file.with_file_name(new_name)
+    } else {
+        output_file.with_file_name(format!("output_{timestamp}"))
```
Trantorian1:
```suggestion
        output_file.with_file_name(format!("output_{timestamp}.csv"))
```

checkpoint-builder/src/sink/csv.rs
```diff
@@ -0,0 +1,152 @@
...
+            batch.push(checkpoint);
+
+            // Write batch if we've reached the commit interval
+            if batch.len() >= commit_interval {
+                let batch_to_write = std::mem::take(&mut batch);
+
+                if let Err(e) = csv_sink.write_checkpoints(batch_to_write) {
+                    error!("Failed to write CSV batch: {}", e);
+                    return Err(e);
+                }
```
Trantorian1: This kind of defeats the whole purpose of re-using the same `batch`, as you are allocating a new `Vec` anyway on each call to `std::mem::take` then `batch.push`. Since `write_checkpoints` only needs to write each batch, and not mutate it, it should be enough to pass `batch` by reference, clearing the vector on a successful write.

checkpoint-builder/src/sink/mod.rs
```diff
@@ -0,0 +1,11 @@
+use anyhow::Result;
+
+use attestor_primitives::AttestationCheckpoint;
+
+pub mod csv;
+
+/// Generic trait for outputting generated checkpoints
+pub trait CheckpointSink {
+    /// Write a batch of checkpoints
+    fn write_checkpoints(&mut self, checkpoints: Vec<AttestationCheckpoint>) -> Result<()>;
```
Trantorian1: Consider refactoring this to 
```suggestion
    fn write_checkpoints(&mut self, checkpoints: &[AttestationCheckpoint]) -> Result<()>;
```
In order to avoid unnecessary allocations between batch writes.

checkpoint-builder/src/source/archive.rs
```diff
@@ -0,0 +1,157 @@
...
+            .json()
+            .with_context(|| format!("Failed to parse JSON from {url}"))?;
+
+        match (status.latest_archived_block, status.total_blocks) {
+            (Some(latest), total) if total > 0 => {
+                let first_height = (latest + 1).saturating_sub(total as u64);
+                Ok(self
+                    .fetch_range(first_height, first_height)?
+                    .into_iter()
+                    .next())
```
Trantorian1:
```suggestion
                self.get(first_height)
```

checkpoint-builder/src/source/archive.rs
```diff
@@ -0,0 +1,157 @@
...
+            .get(&url)
+            .send()
+            .with_context(|| format!("Failed to GET {url}"))?
+            .error_for_status()
+            .with_context(|| format!("HTTP error from {url}"))?
+            .json()
+            .with_context(|| format!("Failed to parse JSON from {url}"))?;
+
+        match latest.latest_block {
+            Some(h) => Ok(self.fetch_range(h, h)?.into_iter().next()),
```
Trantorian1:
```suggestion
            Some(h) => self.get(h),
```

checkpoint-builder/src/source/sled.rs
```diff
@@ -0,0 +1,167 @@
...
+/// Parse a digest from a sled value (32 bytes).
+fn parse_digest(value: &sled::IVec) -> Result<Digest> {
+    if value.len() != 32 {
+        anyhow::bail!(
+            "Invalid digest length: expected 32 bytes, got {}",
+            value.len()
+        );
+    }
+    Ok(Digest::from_slice(value.as_ref()))
+}
```
Trantorian1, line 121: It would be nice to have some basic unit tests for these parsers.

checkpoint-builder/src/config.rs
```diff
@@ -0,0 +1,770 @@
...
+/// A range from block 1000 to 4999 with interval 500:
+/// - Creates checkpoints at blocks: 500, 1000, 1500, 2000, 2500, 3000, 3500, 4000, 4500, 5000 (9 checkpoints)
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CheckpointRange {
+    /// First block height in this range (inclusive)
+    pub height_start: u64,
+    /// Last block height in this range (inclusive)
+    pub height_end: u64,
+    /// Number of blocks per checkpoint within this range
+    pub checkpoint_interval: usize,
```
Trantorian1, line 78: probably should be a  `std::num::NonZeroUsize`

checkpoint-builder/src/config.rs
```diff
@@ -0,0 +1,770 @@
...
+/// - Range contiguity ensures we can chain checkpoints across range boundaries
+/// - Divisibility ensures no partial chunks that would violate checkpoint_interval expectations
+/// - These constraints prevent silent data corruption or missing checkpoints
+pub fn validate_checkpoint_ranges(ranges: &[CheckpointRangeType]) -> Result<()> {
+    if ranges.is_empty() {
+        anyhow::bail!("At least one checkpoint range is required");
+    }
+
+    // Validate each range individually
+    for (i, range_type) in ranges.iter().enumerate() {
```
Trantorian1, line 210: Intuitively it seems like these checks can be merged into a single `for` loop

checkpoint-builder/src/config.rs
```diff
@@ -0,0 +1,770 @@
...
+
+    #[command(flatten)]
+    pub common: CommonArgs,
+}
+
+#[derive(Args, Debug)]
+pub struct ArchiverArgs {
+    /// Base URL of the archiver HTTP API (e.g. http://localhost:8080)
+    #[arg(long, env = "ARCHIVER_URL")]
+    pub archiver_url: String,
```
Trantorian1: You can use `url::Url` directly as it implements `FromStr`

atodorov (dismissed):
See error in https://github.com/gluwa/creditcoin3/actions/runs/25217579269/job/73941500339?pr=787

creditcoinprotoclaw (dismissed):
Approved on request from Didac (no review needed).

creditcoinprotoclaw (approved):
Re-approved after adjustment (requested by Didac).

didac-gluwa (approved):

PR #822, opened 2026-04-14
fix/USCP2-001

Trantorian1:

Fixes https://github.com/chainlight-io/2026-q1-ctc-usc-phase2-audit-issues/issues/1

## Fixes

- Identifies attestation by `digest` and `header_hash` instead of `digest` alone in order to avoid different bls signatures colliding into the same quorum.

### DIFF
--- a/attestor/attestor/src/worker/validation/pool/mod.rs
+++ b/attestor/attestor/src/worker/validation/pool/mod.rs
@@ -529,6 +529,15 @@ struct KeyDigestPending {
     prev_digest_tail: PrevDigestTail,
 }
 
+/// Attestation [digest computation] does not account for all fields in the [`AttestationData`].
+/// Namely, the attestation `header_hash` is absent from digest computation yet is still used for
+/// [attestation data serialization], **which is what attestors sign on**. This means the
+/// attestation digest alone is not a guarantee of uniqueness, and must be paired with the header
+/// hash to avoid collisions.
+///
+/// [digest computation]: attestor_primitives::Attestation::digest
+/// [`AttestationData`]:  attestor_primitives::AttestationData
+/// [attestation data serialization]: attestor_primitives::AttestationData::serialize
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
 struct CompoundDigest {
     digest: attestor_primitives::Digest,
@@ -557,6 +566,9 @@ impl From<CompoundDigest> for attestor_primitives::Digest {
     }
 }
 
+/// Identifying wrapper around [`AttestationInfo`], similar to [`CompoundDigest`].
+///
+/// [`AttestationInfo`]: stream::util::AttestationInfo
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
 struct CompoundInfo {
     height: common::types::Height,
--- a/attestor/attestor/src/worker/validation/pool/mod.rs
+++ b/attestor/attestor/src/worker/validation/pool/mod.rs
@@ -2248,6 +2248,7 @@ mod test {
 
         assert_eq!(inner.forks.votes.len(), 2);
         assert_eq!(inner.forks.forks_by_digest.len(), 2);
+        assert_eq!(inner.forks.forks_by_size.len(), 2);
 
         assert_eq!(
             inner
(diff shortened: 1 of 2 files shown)


atodorov (dismissed):

DylanVerstraete (dismissed):
LGTM 👍

didac-gluwa (dismissed):

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -472,55 +478,111 @@ impl AttestationPoolInner {
...
+/// hash to avoid collisions.
+///
+/// [digest computation]: attestor_primitives::Attestation::digest
+/// [`AttestationData`]:  attestor_primitives::AttestationData
+/// [attestation data serialization]: attestor_primitives::AttestationData::serialize
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+struct CompoundDigest {
     digest: attestor_primitives::Digest,
+    header_hash: attestor_primitives::Digest,
+}
```
beqaabu, line 545: would it be possible to have the same issues with `chain_key`? in other words, it is included in attestation data serialization, and it is checked during the bls signature check, so potentially two attestations having the same digest and hash, but a different `chain_key` would still poison the quorum. 

unless it is impossible to have 2 attestations with different `chain_key`s at this stage inside the pool. would be great if you could elaborate on that

DylanVerstraete (dismissed):

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -2132,10 +2218,62 @@ mod test {
...
+    #[rstest::rstest]
+    #[timeout(TIMEOUT)]
+    async fn attestation_pool_sanity_deduplicate_header_hash(
+        _logs: (),
+        #[from(attestation)]
+        #[with([ATTESTOR_VALID_0], 1, DIGEST_0, DIGEST_0)]
+        attestation_0: AttestationVote,
+        #[from(attestation)]
+        #[with([ATTESTOR_VALID_1], 1, DIGEST_0, DIGEST_1)]
+        attestation_1: AttestationVote,
```
beqaabu, line 2238: btw is this standard rust syntax that im not aware of or some kind of a crate magic? 😆 
using the `#[from(attestation)]` and `#[with(..)]`

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -2132,10 +2218,62 @@ mod test {
...
+            inner
+                .forks
+                .forks_by_digest
+                .get(&CompoundDigest {
+                    digest: attestation_1.attestation.digest(),
+                    header_hash: attestation_1.attestation.attestation_data.header_hash
+                })
+                .unwrap(),
+            &attestation_1
+        );
```
beqaabu, line 2275: I can see that the two attestations are stored separately, but can we also assert that they dont trigger a quorum even if they reach the threshold?

beqaabu (commented):
a couple of questions, but overall looks good

attestor/attestor/src/worker/validation/pool/mod.rs
Trantorian1: 100% crate magic haha. Using [`rstest`](https://docs.rs/rstest/latest/rstest/) for reusable fixtures here. `#[from]` specifies the fixture being used, `#[with]` the fixture parameters. Makes it easier to re-use test setup code as fixture dependencies are automatically resolved/injected :)

attestor/attestor/src/worker/validation/pool/mod.rs
Trantorian1: Good point. Added some extra assertions to be really clear about this :+1:

attestor/attestor/src/worker/validation/pool/mod.rs
Trantorian1: My mistake! I was assuming the p2p worker was already making that check prior to pool insertion, which it was not! I've added the necessary guardrails. Thanks for noticing :)

DylanVerstraete (approved):

beqaabu (approved):

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -472,55 +478,111 @@ impl AttestationPoolInner {
...
+/// attestation digest alone is not a guarantee of uniqueness, and must be paired with the header
+/// hash to avoid collisions.
+///
+/// [digest computation]: attestor_primitives::Attestation::digest
+/// [`AttestationData`]:  attestor_primitives::AttestationData
+/// [attestation data serialization]: attestor_primitives::AttestationData::serialize
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+struct CompoundDigest {
     digest: attestor_primitives::Digest,
+    header_hash: attestor_primitives::Digest,
```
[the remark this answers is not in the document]
Trantorian1: This is checked before insertion by the p2p worker.

atodorov (approved):

PR #826, opened 2026-04-15
fix/USCP2-005

Trantorian1:

## Fixes

- https://github.com/chainlight-io/2026-q1-ctc-usc-phase2-audit-issues/issues/5
- Updates validation logic so that attestor eligibility and bls signatures are checked on reception rather than on quorum.


DylanVerstraete (commented):
Shouldn't we keep the set of pubkeys at epoch rotation instead of requiring a query to the chain for each attestation reception?

Trantorian1:
> Shouldn't we keep the set of pubkeys at epoch rotation instead of requiring a query to the chain for each attestation reception?

That does seem like a better idea yea

attestor/attestor/src/worker/p2p/mod.rs
```diff
@@ -604,4 +644,113 @@ impl WorkerP2P {
...
+        } else {
+            Err(Interrupt::Cont(Error::InvalidAttestation(
+                InvalidCause::InvalidBls,
+            )))
+        }
+    }
+
+    async fn reconnect(
+        &mut self,
+        err: Error,
```
didac-gluwa: Seems a bit excessive to put this parameter just to log it. Wouldn't it be better to just log before calling this function?

didac-gluwa (changes requested):

attestor/attestor/src/worker/p2p/mod.rs
Trantorian1: Imo its easier this way as it centralizes logging into a single implementation. Otherwise we would be dealing with duplicate logs for this very situation, and that tends to lead to differing output over time which can be confusing.

attestor/attestor/src/worker/production/mod.rs
```diff
@@ -411,6 +417,14 @@ impl WorkerAttestationProduction {
                         );
                     }
 
+                    // 2. Update bls keys
+                    //
+                    // Update the set of BLS keys for use by the p2p worker
+                    self.bls
+                        .note_attestors_elected(&mut self.cc3, &attestors)
+                        .await
+                        .map_interrupt(Error::Bls)?;
```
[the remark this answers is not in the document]
Trantorian1: By the time this happens, the runtime would have had to elect an attestor with an invalid bls key, which would be a critical bug in the protocol. Crashing it in this case seems desirable.

DylanVerstraete (dismissed):

didac-gluwa (dismissed):

DylanVerstraete (dismissed):

didac-gluwa (dismissed):

DylanVerstraete (approved):

didac-gluwa (approved):

PR #827, opened 2026-04-15
fix: address informational audit findings

DylanVerstraete:
## Description

Addresses all 5 informational findings from the Q1 2026 CTC USC Phase 2 audit.

### Changes

**1. `AttestorsCount` O(1) counter (`pallet-attestation`)**

Adds a new `AttestorsCount` storage item (keyed by `ChainKey`) incremented in `register_attestor` and decremented in `unregister_attestor`. Replaces the O(N) `iter_prefix_values` call in `attestor_list_has_space` with an O(1) counter lookup.

**2. Typo fix: `fetch_babe_randomness_two_epoch_ego` → `two_epoch_ago` (`cc-client`)**

Fixes the typo in both the function definition and its call site.

**3. EVM block reorg mitigation via `eth_block_confirmations` (`proof-gen-api-server`)**

Adds configurable `eth_block_confirmations` (default: 12) to the proof-gen server. `validate_blocks` rejects requests for blocks that have not reached the required confirmation depth. Configurable via YAML, `--eth-block-confirmations` CLI flag, and `ETH_BLOCK_CONFIRMATIONS` env var.

**4. Upper boundary digest verification (`proof-gen-api-server`)**

`get_attestation_boundaries` and `get_checkpoint_boundaries` now return the upper digest. `build_proof_from_roots` verifies the computed digest at the upper checkpoint matches the cached on-chain value, confirming the proof connects to the real upper attestation/checkpoint entry.

---

- [ ] Modified behavior/functions are covered


attestor/attestor/src/worker/validation/mod.rs
```diff
@@ -674,6 +674,26 @@ impl WorkerAttestationValidation {
...
+                tracing::error!(
+                    ?digest,
+                    height,
+                    "⛔ Quorum votes contain inconsistent continuity proofs"
+                );
+                return Err(Interrupt::Cont(Error::InvalidAttestation(
+                    InvalidCause::InconsistentContinuityProofs,
+                )));
+            }
+        }
```
Trantorian1: A better approach to this would be to include a hash of the continuity proof in the attestation pool de-duplication logic so as to ensure that a quorum cannot be reached on competing proofs. See #822 for an example of how this might be addressed.

pallets/attestation/src/impls.rs
```diff
@@ -748,10 +750,7 @@ impl<T: Config> Pallet<T> {
     }
 
     pub fn attestor_list_has_space(chain_key: ChainKey) -> bool {
-        let count = Attestors::<T>::iter_prefix_values(chain_key)
-            .collect::<Vec<_>>()
-            .len() as u32;
-        count < MaxAttestors::<T>::get(chain_key)
+        AttestorsCount::<T>::get(chain_key) < MaxAttestors::<T>::get(chain_key)
```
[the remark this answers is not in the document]
Trantorian1: This seems like a valid concern.

proof-gen-api-server/src/services/continuity_service/helpers.rs
```diff
@@ -111,6 +119,31 @@ impl ContinuityService {
...
+                    "upper boundary digest mismatch — possible EVM reorg or stale cache"
+                );
+                return Err(ServiceError::Internal {
+                    message: format!(
+                        "upper boundary digest mismatch at block {upper_checkpoint}: \
+                         computed {computed_upper_digest:?}, expected {upper_checkpoint_digest:?}"
+                    ),
+                });
+            }
+        }
```
Trantorian1: Consider adding a test for this case.

proof-gen-api-server/src/services/continuity_service/mod.rs
```diff
@@ -354,6 +359,25 @@ impl ContinuityService {
             });
         }
 
+        // Reorg mitigation: require a minimum number of confirmations before including
+        // a block in a continuity proof. This mirrors the attestor's block-delayed approach.
+        let safe_block = current_block.saturating_sub(self.eth_block_confirmations);
+        if let Some(&header_number) = header_numbers.iter().find(|h| **h > safe_block) {
```
Trantorian1: Is `header_numbers` ordered? Consider that sorting it, then checking for the max is an `O(log(n))` operation, versus linearly checking the maximum which is an `O(n)` operation.

proof-gen-api-server/src/services/continuity_service/mod.rs
```diff
@@ -419,50 +443,50 @@ impl ContinuityService {
...
+    ) -> Option<(u64, H256, u64, H256)> {
         let cache = chain.attestation_cache.read().await;
 
         // Lower: greatest attestation strictly before min_query.
         let lower = cache.range(..min_query).next_back().map(|(&k, &v)| (k, v));
 
-        // Upper: smallest attestation ≥ max_query
-        let upper = cache.range(max_query..).next().map(|(&k, _)| k);
+        // Upper: smallest attestation ≥ max_query (returns both block and digest)
+        let upper = cache.range(max_query..).next().map(|(&k, &v)| (k, v));
```
Trantorian1: This map can be removed
```suggestion
        let upper = cache.range(max_query..).next();
```

proof-gen-api-server/src/services/continuity_service/mod.rs
```diff
@@ -419,50 +443,50 @@ impl ContinuityService {
...
 
         // Lower: greatest checkpoint strictly before min_query.
         // Must be strictly less than min_query so that build_proof_from_roots
         // includes the queried block (it builds from lower_checkpoint + 1).
         let lower = cache.range(..min_query).next_back().map(|(&k, &v)| (k, v));
 
-        // Upper: smallest checkpoint ≥ max_query
-        let upper = cache.range(max_query..).next().map(|(&k, _)| k);
+        // Upper: smallest checkpoint ≥ max_query (returns both block and digest)
+        let upper = cache.range(max_query..).next().map(|(&k, &v)| (k, v));
```
Trantorian1: Same here.
```suggestion
        let upper = cache.range(max_query..).next();
```

.gitignore
```diff
@@ -71,3 +71,4 @@ bridge-usage-example/lib/*
 ./config.json
 primitives/data/*
 prover/artifacts/*
+alice-data/
```
Trantorian1, line 74: Is used in any other context than for local testing? If not please remove this.

attestor/attestor/src/worker/validation/mod.rs
DylanVerstraete: Done. Introduced `CompoundDigest` in `pool/mod.rs` pairing the attestation digest with a continuity proof digest (via `ContinuityProof::compute_continuity_digest`). All active-fork keys (`forks_by_digest`, `KeyHeight`, `KeySize`, `KeyDigest`) now use `CompoundDigest`. `Permit` carries `proof_digest` so `mark_invalid` can reconstruct the exact compound key. Attestors voting with different continuity proofs for the same `(height, attestation_digest)` will now occupy separate fork entries and cannot merge toward quorum.

pallets/attestation/src/impls.rs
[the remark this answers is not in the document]
DylanVerstraete: Added `MigrateAttestorsCountV1ToV2` migration in `pallets/attestation/src/migrations.rs`. It iterates the `Attestors` double-map, counts per-chain-key, and writes into `AttestorsCount`. Storage version bumped from 1 → 2 and the migration is registered in `runtime/src/lib.rs` `SingleBlockMigrations`. Includes try-runtime pre/post hooks that verify counts match.

proof-gen-api-server/src/services/continuity_service/helpers.rs
DylanVerstraete: Added `upper_boundary_digest_mismatch_returns_internal_error` test in `helpers.rs`. It uses a `BadDigestEthProvider` that builds blocks with wrong roots (root = block_number + 9999 instead of the mock CC chain formula), so the computed upper digest diverges from the cached on-chain value and `build_proof_from_roots` returns `ServiceError::Internal`.

proof-gen-api-server/src/services/continuity_service/mod.rs
DylanVerstraete: Done. `get_proof_batch` now calls `sort_unstable()` on `header_numbers` before passing to `validate_blocks`. The two max-checks in `validate_blocks` now use `last().filter(...)` (O(1)) instead of `find(h > ...)` (O(n)). The subsequent `from_header`/`to_header` lookups use `first()`/`last()` as well.

proof-gen-api-server/src/services/continuity_service/mod.rs
DylanVerstraete: Fixed. Removed the `.map(|(&k, &v)| (k, v))` wrapper; the `match` arms now destructure the reference with `Some((&upper_block, &upper_digest))` instead.

proof-gen-api-server/src/services/continuity_service/mod.rs
DylanVerstraete: Fixed here too — same change applied to `get_checkpoint_boundaries`.

.gitignore
DylanVerstraete: Removed. `alice-data/` was only used as a `--base-path` in CI workflows for local node testing. Also removed the already-committed `alice-data/` directory from git tracking.

atodorov (dismissed):

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -535,37 +542,40 @@ struct KeyDigestPending {
...
+///
 /// [digest computation]: attestor_primitives::Attestation::digest
 /// [`AttestationData`]:  attestor_primitives::AttestationData
 /// [attestation data serialization]: attestor_primitives::AttestationData::serialize
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
 struct CompoundDigest {
     digest: attestor_primitives::Digest,
     header_hash: attestor_primitives::Digest,
+    /// Hash of the continuity proof chain (prevents quorum on competing proofs).
+    proof_digest: attestor_primitives::Digest,
```
Trantorian1: Imo and based on the digest computation method name it would be clearer for this to be called `continuity_digest`.
```suggestion
    continuity_digest: attestor_primitives::Digest,
```

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -665,17 +678,21 @@ impl AttestationPoolForks {
         let digest = attestation.digest();
         let attestor = attestation.attestor_id();
         let header_hash = attestation.attestation_data.header_hash;
+        let proof = &attestation.continuity_proof;
+        let start_block = height.saturating_sub(proof.len() as u64);
```
Trantorian1: An underflow here indicates an invalid attestation: we can't have a continuity proof reaching further back than genesis. More generally, `start_block` being less than the configured chain genesis constitutes an invalid attestation. This should be used as a rejection reason to avoid storing known invalid votes in the pool.

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -2138,6 +2161,19 @@ mod test {
...
+        let proof = &attestation.continuity_proof;
+        let start_block = attestation
+            .header_number()
+            .saturating_sub(proof.len() as u64);
+        CompoundDigest {
+            digest: attestation.digest(),
+            header_hash: attestation.attestation_data.header_hash,
+            proof_digest: proof.compute_continuity_digest(start_block),
+        }
+    }
```
Trantorian1: It seems cleaner to refactor this as a `From` implementation, something like
```suggestion
impl From<&common::types::Attestation> for CompoundDigest {
    fn from(attestation: &common::types::Attestation) -> Self {
        let proof = &attestation.continuity_proof;
        let start_block = attestation
            .header_number()
            .saturating_sub(proof.len() as u64);
        CompoundDigest {
            digest: attestation.digest(),
            header_hash: attestation.attestation_data.header_hash,
            proof_digest: proof.compute_continuity_digest(start_block),
        }
    }
}
```
This can be scoped to any of the `#[cfg(test)]` modules to make sure this trait isn't implemented outside of testing conditions.

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -2216,11 +2252,15 @@ mod test {
...
+            height: att0.header_number(),
             digest: CompoundDigest {
-                digest: attestation_0.attestation.digest(),
-                header_hash: attestation_0.attestation.attestation_data.header_hash
+                digest: att0.digest(),
+                header_hash: att0.attestation_data.header_hash,
+                proof_digest: proof0.compute_continuity_digest(start0),
             }
         }));
     }
```
Trantorian1:
```suggestion
        assert!(inner.forks.votes_invalid.contains(&KeyDigest {
            height: attestation_0.attestation.header_number(),
        digest: compound(&attestation_0.attestation),
    }));
}
```

Trantorian1: Also consider replacing this by a call to `into` if ever you go down the route of the `From` implementation described above.

.gitignore
DylanVerstraete: Keeping it — `alice-data/` is used as the node `--base-path` in multiple CI workflows (`ci.yml`, `proof-gen-api-server.yml`, `chainspec.yml`). The .gitignore entry prevents the generated chain data from being accidentally committed. Not local-testing-only.

attestor/attestor/src/worker/validation/pool/mod.rs
DylanVerstraete: Done — renamed `proof_digest` → `continuity_digest` throughout `CompoundDigest`, `CompoundInfo`, and all construction sites. Pushed in latest commit.

attestor/attestor/src/worker/validation/pool/mod.rs
DylanVerstraete: Done — added an explicit underflow check before computing `start_block`: if `proof.len() as u64 > height` we return `Err(Error::InvalidProof(attestor, height))` and log an error. Added `InvalidProof` to the `Error` enum, handled in `log_error` with a `tracing::error!` and in `p2p/mod.rs` as a gossipsub `Reject` (same treatment as `Unauthorized` — data that cannot be valid regardless of chain state).

attestor/attestor/src/worker/validation/pool/mod.rs
DylanVerstraete: Also simplified the `votes_invalid` assert to use `CompoundDigest::from(att0)` directly, eliminating the `att0`/`proof0`/`start0` temporaries.

attestor/attestor/src/worker/validation/pool/mod.rs
DylanVerstraete: Done — added an explicit underflow check before computing `start_block`: if `proof.len() as u64 > height` we return `Err(Error::InvalidProof(attestor, height))`. Added `InvalidProof` to the `Error` enum with `log_error` handler and treated as gossipsub `Reject` in `p2p/mod.rs`.

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -679,15 +679,26 @@ impl AttestationPoolForks {
...
+        if proof.len() as u64 > height {
+            tracing::error!(
+                %attestor,
+                height,
+                proof_len = proof.len(),
+                "⛔ Continuity proof length exceeds attestation height (reaches before genesis)"
+            );
+            return Err(Error::InvalidProof(attestor, height));
+        }
+        let start_block = height - proof.len() as u64;
```
Trantorian1: This can be simplified as:
```rust
let start_block = height.checked_sub(proof.len() as u64).ok_or(Error::InvalidProof(attestor, height))?;
```
The added log isn't necessary here as it is already provided by calling `Error::log_error`

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -2257,11 +2269,7 @@ mod test {
         let start0 = att0.header_number().saturating_sub(proof0.len() as u64);
         assert!(inner.forks.votes_invalid.contains(&KeyDigest {
             height: att0.header_number(),
-            digest: CompoundDigest {
-                digest: att0.digest(),
-                header_hash: att0.attestation_data.header_hash,
-                proof_digest: proof0.compute_continuity_digest(start0),
-            }
+            digest: CompoundDigest::from(att0),
```
Trantorian1: using the compiler-generated `into` implementation might be cleaner where applicable.

attestor/attestor/src/worker/validation/pool/mod.rs
DylanVerstraete: Done — simplified to `height.checked_sub(proof.len() as u64).ok_or(Error::InvalidProof(attestor.clone(), height))?`. Removed the explicit `tracing::error!` block.

attestor/attestor/src/worker/validation/pool/mod.rs
DylanVerstraete: Done — updated to `att0.into()`, also removed the now-unused `proof0`/`start0` temporaries.

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -433,11 +433,18 @@ impl AttestationPoolInner {
             let height = fork.attestation.header_number();
             let digest = fork.attestation.digest();
             let header_hash = fork.attestation.attestation_data.header_hash;
+            let proof = &fork.attestation.continuity_proof;
+            let start_block = fork
```
BradleyOlson64: As Cursor said, please use checked sub here as well and Error::InvalidProof upon failure

beqaabu (dismissed):

attestor/attestor/src/worker/validation/pool/mod.rs
DylanVerstraete: Already done in the latest commit — replaced the `if proof.len() > height { tracing::error!(...); ... }` block with:
```rust
let start_block = height
    .checked_sub(proof.len() as u64)
    .ok_or(Error::InvalidProof(attestor.clone(), height))?;
```
No extra log at the call site — logging is handled by `Error::log_error` in the p2p layer.

attestor/attestor/src/worker/validation/pool/mod.rs
DylanVerstraete: Done — the test now uses `att0.into()` (leveraging the `#[cfg(test)] impl From<&Attestation> for CompoundDigest`), replacing the manual `CompoundDigest { digest: att0.digest(), header_hash: ..., continuity_digest: ... }` struct literal.

attestor/attestor/src/worker/validation/pool/mod.rs
DylanVerstraete: Done. Changed `peek` return type to `Result<Option<(Quorum, Permit)>, Error>` and replaced `saturating_sub` with `checked_sub().ok_or(Error::InvalidProof(attestor, height))?`. Updated `poll_next` to match the new signature — `Err` path calls `log_error` and returns `Pending`.

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -433,11 +433,18 @@ impl AttestationPoolInner {
...
+                .attestation
+                .header_number()
+                .saturating_sub(proof.len() as u64);
+            let continuity_digest = proof.compute_continuity_digest(start_block);
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed. `peek` now returns `Result<Option<(Quorum, Permit)>, Error>` and uses `checked_sub().ok_or(Error::InvalidProof(...))` — consistent with `push`. `poll_next` updated to handle the `Err` path.

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -427,32 +427,42 @@ impl AttestationPoolInner {
...
+                let quorum = Quorum(fork.votes.clone());
+                let height = fork.attestation.header_number();
+                let digest = fork.attestation.digest();
+                let header_hash = fork.attestation.attestation_data.header_hash;
+                let proof = &fork.attestation.continuity_proof;
+                let attestor = fork.attestation.attestor_id();
+                let start_block = height
+                    .checked_sub(proof.len() as u64)
+                    .ok_or(Error::InvalidProof(attestor, height))?;
+                let continuity_digest = proof.compute_continuity_digest(start_block);
```
Trantorian1: Mb: you actually don't need to check for underflow here as this was already verified during initial insertion into the pool, therefore this cannot have become invalid since.

PR #832, opened 2026-04-16
fix/USCP2-010

Trantorian1:

---

## Fixes

- https://github.com/chainlight-io/2026-q1-ctc-usc-phase2-audit-issues/issues/10
- Makes the attestation pool unbounded by removing any max size consideration

In practice this is mitigated by the fact that we now only allow registered attestors to submit attestations to the pool, however this is only a temporary fix while we can work out a better pool eviction policy.

### DIFF
--- a/attestor/attestor/src/lib.rs
+++ b/attestor/attestor/src/lib.rs
@@ -28,7 +28,6 @@ pub struct Config {
 
     p2p: worker::p2p::ConfigIncomplete,
     api: worker::api::ConfigIncomplete,
-    pool: worker::validation::pool::ConfigIncomplete,
 }
 
 // ---------------------------------------- [ Main loop ] -------------------------------------- //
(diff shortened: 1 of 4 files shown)


Trantorian1 (commented):
Finished self-review.

atodorov (commented):
This PR seems to be up-to-date, yet the docker build failed with an NPM error. 

It looks like the package it tries to download is missing though:
message	"/base64-js/-/base64-js-1.5.1.tgz: does not exist"

DylanVerstraete (approved):

Trantorian1:
> This PR seems to be up-to-date, yet the docker build failed with an NPM error.

Passing now

BradleyOlson64 (approved):

atodorov (approved):

PR #839, opened 2026-04-20
fix(pool): continuity proof deduplication

Trantorian1:

## Fixes

- https://github.com/chainlight-io/2026-q1-ctc-usc-phase2-audit-issues/issues/15
- Attestation continuity proof deduplication

## Tests

- Attestations differing only by their continuity proof do not contribute to the same quorum

### DIFF
--- a/attestor/attestor/src/worker/validation/pool/mod.rs
+++ b/attestor/attestor/src/worker/validation/pool/mod.rs
@@ -514,14 +514,15 @@ struct KeyDigestPending {
     prev_digest_tail: PrevDigestTail,
 }
 
-/// Attestation [digest computation] does not account for all fields in the [`AttestationData`].
+/// Attestation [digest computation] does not account for all fields in an [`Attestation`].
 /// Namely, the attestation `header_hash` is absent from digest computation yet is still used for
-/// [attestation data serialization], **which is what attestors sign on**. This means the
-/// attestation digest alone is not a guarantee of uniqueness, and must be paired with the header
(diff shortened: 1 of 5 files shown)


Trantorian1 (commented):
Finished self-review

didac-gluwa (approved):

attestor/attestor/src/worker/validation/pool/mod.rs
```diff
@@ -1590,16 +1598,69 @@ impl AttestationVote {
...
+        CompoundInfo {
+            height: self.attestation.header_number(),
+            digest: self.compound_digest(),
+        }
+    }
 }
 
 impl std::cmp::PartialEq for AttestationVote {
     fn eq(&self, other: &Self) -> bool {
+        let matching_digest = || self.attestation.digest() == other.attestation.digest();
```
BradleyOlson64, line 1631: A bit odd here to use closures rather than just setting a boolean var for each condition.

Works fine though 🤷‍♂️

BradleyOlson64 (approved):
LGTM

attestor/attestor/src/worker/validation/pool/mod.rs
Trantorian1: Keeps the computation lazy so we don't evaluate later conditions if the prior turn out to be false. In particular this avoids computing the continuity digest on votes which do not match.

DylanVerstraete (approved):

PR #841, opened 2026-04-20
fix(proof-gen-api-server): guard against EVM block reorgs via block_confirmation_depth

DylanVerstraete:
Blocks near the EVM chain tip can be reorganised away. The server was previously accepting any block that existed at tip height, which means a proof could be generated for a block that later disappears.

## Changes
- Add `block_confirmation_depth: u64` to `ContinuityConfig` and its builder.
- Add `ContinuityBuilder::get_confirmed_last_block()` that returns `tip - block_confirmation_depth` (saturating), mirroring the block-delayed approach used by the attestor.
- `validate_blocks` in `ContinuityService` now calls `get_confirmed_last_block()` instead of `get_last_block()`.
- Wire the new field through `ChainConfig`, `ChainConfigFile` (YAML), and the server CLI (`--block-confirmation-depth` / `BLOCK_CONFIRMATION_DEPTH` env).
- Defaults to 0 (no lag) for backward compatibility.
- Document the new option in `config.example.yaml` and `.env.example`.

Typical safe value for Ethereum mainnet: **12**.

Addresses https://github.com/chainlight-io/2026-q1-ctc-usc-phase2-audit-issues/issues/15 point 4


atodorov (dismissed):

didac-gluwa (dismissed):

.gitignore
```diff
@@ -71,3 +71,4 @@ bridge-usage-example/lib/*
 ./config.json
 primitives/data/*
 prover/artifacts/*
+alice-data/
```
Trantorian1: Is there any reason to have this beyond local testing?

DylanVerstraete: The bot has this local dir, idk why it keeps adding this

didac-gluwa (dismissed):

BradleyOlson64 (dismissed):

common/continuity/src/builder/mod.rs
```diff
@@ -593,6 +593,17 @@ impl ContinuityBuilder {
         self.eth_provider.get_last_block().await
     }
 
+    /// Get both the raw chain tip and the confirmed block height for reorg protection.
+    ///
+    /// Returns `(tip, tip - block_confirmation_depth)` (confirmed saturates at 0).
+    /// Use the confirmed value to decide whether to accept a requested block;
+    /// use the tip value in user-facing error messages so clients see the real chain height.
+    pub async fn get_confirmed_last_block(&self) -> Result<(u64, u64)> {
+        let tip = self.eth_provider.get_last_block().await?;
```
beqaabu, line 608: can we add a note to `get_last_block` that it doesnt care about reorgs? same way we document `get_confirmed_last_block`

common/continuity/src/builder/mod.rs
```diff
@@ -593,6 +593,17 @@ impl ContinuityBuilder {
         self.eth_provider.get_last_block().await
     }
 
+    /// Get both the raw chain tip and the confirmed block height for reorg protection.
+    ///
+    /// Returns `(tip, tip - block_confirmation_depth)` (confirmed saturates at 0).
+    /// Use the confirmed value to decide whether to accept a requested block;
+    /// use the tip value in user-facing error messages so clients see the real chain height.
+    pub async fn get_confirmed_last_block(&self) -> Result<(u64, u64)> {
```
beqaabu, line 607: also can we add tests for this fn?

beqaabu (dismissed):

common/continuity/src/builder/mod.rs
DylanVerstraete: Done — added a `# Note` section pointing users to `get_confirmed_last_block` for reorg-safe usage.

common/continuity/src/builder/mod.rs
DylanVerstraete: Done — added 4 tests: depth 0, normal depth subtraction, saturation at 0, and that `get_last_block` ignores depth.

beqaabu (dismissed):

didac-gluwa (approved):

beqaabu (approved):

PR #849, opened 2026-04-21
fix: use deno's native websocket in ethers

beqaabu:

---

### DIFF
--- a/scripts/traffic-simulator/src/watchers/blockWatcher.ts
+++ b/scripts/traffic-simulator/src/watchers/blockWatcher.ts
@@ -34,7 +34,7 @@ export class BlockWatcher extends BaseWatcher {
   protected async connect(): Promise<void> {
     try {
       console.log(`🔗 Connecting to source chain: ${this.rpcUrl}`);
-      this.provider = new WebSocketProvider(this.rpcUrl);
+      this.provider = new WebSocketProvider(new WebSocket(this.rpcUrl));
       await this.provider.ready;
 
       const network = await this.provider.getNetwork();


DylanVerstraete (approved):

scripts/traffic-simulator/src/watchers/blockWatcher.ts
```diff
@@ -34,7 +34,7 @@ export class BlockWatcher extends BaseWatcher {
   protected async connect(): Promise<void> {
     try {
       console.log(`🔗 Connecting to source chain: ${this.rpcUrl}`);
-      this.provider = new WebSocketProvider(this.rpcUrl);
+      this.provider = new WebSocketProvider(new WebSocket(this.rpcUrl));
```
atodorov, line 37: Can you explain what's the difference between the two  and why the new change is better ? I don't get it.

beqaabu: essentially we were facing this issue: https://github.com/denoland/deno/issues/33358
it was a quick fix for the `TypeError: Cannot read properties of undefined (reading 'Symbol(Deno.internal.rid)')` which was observed on devnet traffic sim and was caused by a bug in deno's node:http package. 

when doing `new WebSocket(this.rpcUrl)` we use denos global native `Websocket` which avoids `node:http` completely, thus avoiding the buggy code. a cleaner fix right now would be updating deno version because a new release has fixed this, but at the time the new release didnt exist yet. I'll check if the issue is fixed upstream

atodorov: See https://github.com/gluwa/creditcoin3/pull/860

beqaabu: deno v2.7.13 seems to fix the initial ws connection issue but now there is a new panic on ws closure. so I think leaving this fix is the best way until they fix it upstream

atodorov (approved):

PR #851, opened 2026-04-21
feat(precompile): add attestation interface precompile

DylanVerstraete:
## feat(precompile): attestor stash interface

Adds the **AttestorStash precompile** (`0x0FD4` / 4052) — an EVM-callable interface for stash-facing `pallet_attestation` dispatchables.  Stash accounts that live in the EVM address space (i.e. whose `H160` maps to a Substrate `AccountId` via `HashedAddressMapping`) can now register, chill, unregister, and withdraw unbonded funds directly through the EVM without constructing raw Substrate extrinsics.

### Precompile interface

**Write operations** (dispatched as the caller's mapped stash)
| Function | Pallet call |
|----------|-------------|
| `registerAttestor(uint64 chainKey, bytes32 attestorId)` | `attestation::register_attestor` |
| `chill(uint64 chainKey, bytes32 attestorId)` | `attestation::chill` |
| `unregisterAttestor(uint64 chainKey, bytes32 attestorId)` | `attestation::unregister_attestor` |
| `withdrawUnbonded()` | `attestation::withdraw_unbonded` |

**View functions**
| Function | Returns |
|----------|---------|
| `getAttestor(uint64, bytes32)` | `AttestorInfo` (exists, status, stash, hasBlsKey) |
| `isActiveAttestor(uint64, bytes32)` | `bool` |
| `getAttestorsCount(uint64)` | `uint32` |
| `getLedger(bytes32)` | `LedgerInfo` (totalStaked, active, unlockingChunks, withdrawable) |
| `getMinBondRequirement(uint64)` | `uint128` |

Each write emits a matching EVM log: `AttestorRegistered`, `AttestorUnregistered`, `AttestorChilled`, `UnbondedWithdrawn`.

### CLI migration

All attestor CLI commands (`registerAttestor`, `unregisterAttestor`, `chill`, `withdrawUnbonded`) now route through the precompile instead of direct Substrate extrinsics. Read commands (`showStatus`, `showStashBalance`) use the new view functions. `showAttestorsFor` still uses the Substrate RPC path (full iteration required).

### Tests

- Rust unit tests: full coverage of all precompile functions including specific error variant assertions.
- TypeScript blockchain tests (`attestor-stash.test.ts`): end-to-end register → chill → unregister flow via EVM.

**Bug fix (CI):** Auto-estimated gas for `unregisterAttestor` was ~83,850 — insufficient to cover the PoV component (~71,700 gas at ratio 14.3) of the underlying pallet call. The block-builder silently dropped the tx (PoV exhausted), leaving Alith's nonce stuck and cascading into `substrate-transfer.test.ts`. Fixed by using explicit `gasLimit: 300_000` on all write calls in the test.


cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
```diff
@@ -0,0 +1,217 @@
...
+        expect(log).toBeDefined();
+        expect(log.topics[1]).toBe(chainKeyTopic(chainKey));
+        expect(log.topics[2]).toBe(attestorId.toLowerCase());
+        expect(log.topics[3]).toBe(addressTopic(alith.address));
+    }, 60_000);
+
+    // -----------------------------------------------------------------------------
+    // unregisterAttestor
+    // -----------------------------------------------------------------------------
+
```
Trantorian1, line 158: Would be good to add a test against an unknown attestor as well. Even if the behavior is in principle already covered by the second call to `unregisterAttestor` this makes it more explicit.

precompiles/attestor-stash/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
+                    Precompile,
+                    PCall::register_attestor {
+                        chain_key: TEST_CHAIN_KEY,
+                        attestor_id: attestor_id(),
+                    },
+                )
+                .execute_reverts(|output| {
+                    from_utf8(output)
+                        .unwrap()
+                        .contains("Dispatched call failed with error: ")
```
atodorov: Can you assert on the actual error please? We want to be sure that it failed b/c of no enough balance, not for some other reason.

precompiles/attestor-stash/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
+                    Precompile,
+                    PCall::register_attestor {
+                        chain_key: unsupported_chain,
+                        attestor_id: attestor_id(),
+                    },
+                )
+                .execute_reverts(|output| {
+                    from_utf8(output)
+                        .unwrap()
+                        .contains("Dispatched call failed with error: ")
```
atodorov: Same comment: please assert on the actual error.

precompiles/attestor-stash/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
+                    Precompile,
+                    PCall::unregister_attestor {
+                        chain_key: TEST_CHAIN_KEY,
+                        attestor_id: attestor_id_b(),
+                    },
+                )
+                .execute_reverts(|output| {
+                    from_utf8(output)
+                        .unwrap()
+                        .contains("Dispatched call failed with error: ")
```
atodorov: please assert on the actual error

precompiles/attestor-stash/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
+                    Precompile,
+                    PCall::chill {
+                        chain_key: TEST_CHAIN_KEY,
+                        attestor_id: attestor_id(),
+                    },
+                )
+                .execute_reverts(|output| {
+                    from_utf8(output)
+                        .unwrap()
+                        .contains("Dispatched call failed with error: ")
```
atodorov: Please assert on the actual error.

precompiles/attestor-stash/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
```
atodorov: Please assert on the actual error.

precompiles/attestor-stash/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
+    ExtBuilder::default()
+        .with_balances(vec![(Alice, 10 * MIN_BOND)])
+        .build()
+        .execute_with(|| {
+            precompiles()
+                .prepare_test(alice, Precompile, PCall::withdraw_unbonded {})
+                .execute_reverts(|output| {
+                    from_utf8(output)
+                        .unwrap()
+                        .contains("Dispatched call failed with error: ")
```
atodorov: Please assert on the actual error.

precompiles/attestor-stash/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
+                )
+                .execute_returns(true);
+
+            // Withdraw is expected to revert because the unbonding duration has
+            // not elapsed yet in the test harness. We still assert the call
+            // path is reachable by the precompile (i.e. the call is dispatched
+            // and the pallet is the one that rejects).
+            precompiles()
+                .prepare_test(alice, Precompile, PCall::withdraw_unbonded {})
+                .execute_some();
```
atodorov: I'm not sure I follow the logic here. Can't we wait/progress forward a number of blocks so that this call actually succeeds and assert on that fact ?

atodorov: OTOH if "Withdraw is expected to revert" how come we get an event emitted afterwards ?

precompiles/attestor-stash/src/tests.rs
```diff
@@ -0,0 +1,353 @@
...
+
+            // Sanity check: event selector is wired correctly by emitting
+            // through the helper directly (pre-existing event wiring check).
+            let _ = log2(
+                Precompile,
+                SELECTOR_LOG_UNBONDED_WITHDRAWN,
+                H256::from(alice),
+                Vec::<u8>::new(),
+            );
```
atodorov: This snippet doesn't actually appear to check anything or am I missing something ?

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
```diff
@@ -0,0 +1,217 @@
...
+        );
+        expect(log).toBeDefined();
+        expect(log.topics[1]).toBe(chainKeyTopic(chainKey));
+        expect(log.topics[2]).toBe(attestorId.toLowerCase());
+        expect(log.topics[3]).toBe(addressTopic(alith.address));
+    }, 90_000);
+
+    test('registerAttestor twice for the same stash on the same chain reverts', async () => {
+        await expect(contract.registerAttestor(chainKey, attestorId, { gasPrice })).rejects.toThrow(
+            /Dispatched call failed with error:/,
```
atodorov: Please assert on the actual error.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
```diff
@@ -0,0 +1,217 @@
...
+        );
+    });
+
+    test('registerAttestor on an unsupported chain reverts', async () => {
+        const unsupportedChainKey = 9_999_999;
+        // Also use a fresh attestor id: this mustn't conflict with the one we just registered.
+        const freshAttestorId = randomAttestorId();
+
+        await expect(contract.registerAttestor(unsupportedChainKey, freshAttestorId, { gasPrice })).rejects.toThrow(
+            /Dispatched call failed with error:/,
```
atodorov: Please assert on the actual error.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
```diff
@@ -0,0 +1,217 @@
...
+    });
+
+    // -----------------------------------------------------------------------------
+    // chill
+    // -----------------------------------------------------------------------------
+
+    test('chill on an unknown attestor id reverts (AddressNotAttestor)', async () => {
+        const unknownAttestorId = randomAttestorId();
+        await expect(contract.chill(chainKey, unknownAttestorId, { gasPrice })).rejects.toThrow(
+            /Dispatched call failed with error:/,
```
atodorov: Please assert on the actual error.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
```diff
@@ -0,0 +1,217 @@
...
+        );
+        expect(log).toBeDefined();
+        expect(log.topics[1]).toBe(chainKeyTopic(chainKey));
+        expect(log.topics[2]).toBe(attestorId.toLowerCase());
+        expect(log.topics[3]).toBe(addressTopic(alith.address));
+    }, 60_000);
+
+    test('unregisterAttestor a second time reverts (AddressNotAttestor)', async () => {
+        await expect(contract.unregisterAttestor(chainKey, attestorId, { gasPrice })).rejects.toThrow(
+            /Dispatched call failed with error:/,
```
atodorov: Please assert on the actual error.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
```diff
@@ -0,0 +1,217 @@
...
+        const after = await api.query.attestation.attestors(chainKey, attestorId);
+        expect(after.isSome).toBe(true);
+        const storedAttestor = after.unwrap();
+        expect(storedAttestor.stash.toHex().toLowerCase()).toBe(stashAccountIdHex.toLowerCase());
+
+        // Log topic/arg assertions.
+        const log = receipt.logs.find(
+            (l: any) =>
+                l.address.toLowerCase() === attestorStashAddress.toLowerCase() &&
+                l.topics[0] === ATTESTOR_REGISTERED_TOPIC,
```
atodorov, line 104: Q: Is this event going to be picked up by cc3-indexer ? 

IMO repeating the same test scenarios here as are in the unit test doesn't bring lots of additional value (that is to say register success/error IMO is enough) but registering attestors via precompile and asserting cc3-indexer picks up on that sounds to me more valuable.

atodorov (changes requested):
Overall looks good. A number of changes related to tests & assertions could improve this PR but that may also be a follow-up.

precompiles/attestor-stash/src/tests.rs
DylanVerstraete: Fixed — now asserts `InsufficientBalance` specifically.

precompiles/attestor-stash/src/tests.rs
DylanVerstraete: Fixed — now asserts ChainNotSupported specifically.

precompiles/attestor-stash/src/tests.rs
DylanVerstraete: Fixed — now asserts AddressNotAttestor specifically.

precompiles/attestor-stash/src/tests.rs
DylanVerstraete: Fixed — now asserts AddressNotAttestor specifically.

precompiles/attestor-stash/src/tests.rs
DylanVerstraete: Fixed — now asserts NotYourAttestor specifically.

precompiles/attestor-stash/src/tests.rs
DylanVerstraete: Fixed — now asserts NotStash specifically.

precompiles/attestor-stash/src/tests.rs
DylanVerstraete: Done — rewrote the test to write pallet_staking::CurrentEra directly past the bonding duration, then assert the precompile call succeeds and emits UnbondedWithdrawn. No need for full session/babe machinery.

DylanVerstraete: Good catch — the comment was wrong and misleading. The withdraw now actually succeeds (we advance the era), and the event is properly asserted via expect_log.

precompiles/attestor-stash/src/tests.rs
DylanVerstraete: Removed — that dead log2() snippet was only there as a compile check for the selector. Now that the test actually calls withdraw_unbonded successfully and asserts the event, the snippet is gone.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
DylanVerstraete: Fixed — now matches /Dispatched call failed with error:.*AlreadyAttestor/.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
DylanVerstraete: Fixed — now matches /Dispatched call failed with error:.*ChainNotSupported/.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
DylanVerstraete: Fixed — now matches /Dispatched call failed with error:.*AddressNotAttestor/.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
DylanVerstraete: Fixed — now matches /Dispatched call failed with error:.*AddressNotAttestor/.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
DylanVerstraete: The events emitted by this precompile are standard Substrate events deposited via `deposit_event` in `pallet-attestation` — they go through the same event pipeline that cc3-indexer listens to. The EVM log emitted by the precompile is a separate artifact on the EVM side.

Regarding testing the cc3-indexer integration: that would be a cross-repo concern (and would need the indexer running against a live dev chain). Agreed that it would be valuable, but I think it belongs in a separate issue/PR tracking indexer coverage rather than blocking this precompile PR. Happy to file that as a follow-up.

Trantorian1 (commented):
Looks good. See @atodorov's requests for changes -also note https://github.com/gluwa/creditcoin3/actions/runs/24732972097/job/72352182724?pr=851, which is failing in CI.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
DylanVerstraete: Added — `unregisterAttestor an unknown attestor id reverts (AddressNotAttestor)` now sits before the second-call test, using a fresh `randomAttestorId()` to make the error path explicit.

cli/src/test/blockchain-tests/precompiles/attestor-stash.test.ts
atodorov: > Regarding testing the cc3-indexer integration: that would be a cross-repo concern (and would need the indexer running against a live dev chain).

cc3-indexer test suite lives in the same repository so I'm not sure what you are talking about. It's not a cross-repo concern from what I can see and doesn't need a live chain in order to be exercised.  If you mean something else please clarify? 


> Agreed that it would be valuable, but I think it belongs in a separate issue/PR tracking indexer coverage rather than blocking this precompile PR. Happy to file that as a follow-up.

Agreed, can be a follow up. Please open a ticket so we don't forget about it. Feel free to assign to me.

DylanVerstraete:
@atodorov I think there's a decision to be made first, either we only allow attestor operations (register, chill, ..) via the precompile and change the indexer schema to attach an EVM account instead of substrate ID. This is because the EVM account will not map to a substrate account ID. 

The pallet's events will still trigger right now, so the stash account value will be wrong in any case. If we change the flow to only use the precompile, more changes are needed in the zombienet for example. I'll double check with David first.

didac-gluwa (commented):
Looks good! One question I have: do we also plan to provide like a set of view methods to inspect the pallet storage (for example, checking if attestor is active or not)?

DylanVerstraete:
> Looks good! One question I have: do we also plan to provide like a set of view methods to inspect the pallet storage (for example, checking if attestor is active or not)?

We don't have that for now, but yes that might be good to add indeed.

beqaabu (dismissed):

cli/src/commands/attestor/registerAttestor.ts
```diff
@@ -1,31 +1,41 @@
...
-    console.log(result.info);
-    process.exit(result.status);
+    try {
+        const tx = await contract.registerAttestor(BigInt(chainKey), attestorId32);
+        const receipt = await tx.wait();
+        if (receipt.status === 1) {
+            console.log(`Transaction included at block (hash: ${receipt.blockHash})`);
+            process.exit(0);
+        } else {
+            console.log('Transaction failed');
```
atodorov: use console.error for error messages

cli/src/test/integration-tests/attestor/chill.test.ts
```diff
@@ -8,34 +8,24 @@ import fs = require('fs');
...
     chain_Anvil1_Url,
     chain_Anvil2_Key,
 } from '../../blockchain-tests/pallets/supported-chains/consts';
+import { parseAmount } from '../../../commands/options';
 
 describe('chill', () => {
     let api: ApiPromise;
     let caller: any;
-    let proxy: any;
-    let wrongProxy: any;
```
atodorov, line 34: Q: do we lose the ability to use proxy accounts here, it looks like it?

If that's the case the whole test matrix needs to be simplified as well.

atodorov (changes requested):

cli/src/test/integration-tests/attestor/chill.test.ts
DylanVerstraete: Yes, correct — proxy-signed stash operations aren't supported on the EVM path. The attestor-stash precompile dispatches with the AccountId derived from the EVM msg.sender via HashedAddressMapping; there's no pallet_proxy::proxy wrapping on that call path, and adding one without also mapping Substrate proxy delegations into EVM identity wouldn't be meaningful. So for registerAttestor / unregisterAttestor / chill / withdrawUnbonded the stash must sign directly from its EVM key.

.github/workflows/ci.yml
```diff
@@ -1216,26 +1216,14 @@ jobs:
...
-
-          - proxy: "yes"
-            secret: not-a-proxy
-
-          - proxy: "yes"
-            secret: valid-proxy
-            proxy_type: All
-
-    name: attestor-cli-testing / proxy=${{ matrix.proxy }} / ${{ matrix.secret }} / type=${{ matrix.proxy_type }}
+    name: attestor-cli-testing
```
atodorov: You can remove this line b/c it's not relevant anymore.

.github/workflows/ci.yml
```diff
@@ -1216,26 +1216,14 @@ jobs:
       - uses: actions/checkout@v6
 
   attestor-cli-testing:
+    # Attestor stash operations are dispatched through the EVM precompile, which
+    # uses the EVM `msg.sender` as origin via HashedAddressMapping. Substrate
+    # `pallet_proxy` delegation is not plumbed through the EVM, so there's no
+    # proxy matrix here — the stash signs directly.
     needs:
       - build-creditcoin-node-for-testing-ci
       - trigger-attestor-cli
```
atodorov: trigger-attestor-cli CI job is not relevant anymore and you can remove it

cli/src/test/integration-tests/helpers.ts
```diff
@@ -142,3 +150,82 @@ export async function setMinBondConfig(api: ApiPromise, value: BN | number | str
     const sudoKeyring: CallerKeyring = { type: 'caller', pair: initAliceKeyring() };
     await setStakingConfig(sudoKeyring, api, null, value, null, null, null, null, null);
 }
+
+// Transition a registered attestor into the Active set by submitting `attest()`
```
atodorov: Looks like this comment should be on top of the activateAttestor function.

cli/src/commands/attestor/registerAttestor.ts
```diff
@@ -1,31 +1,41 @@
...
+    substrateAddressToBytes32,
+    extractEvmError,
+} from '../../lib/attestor/precompile';
+import { getStringFromEnvVar } from '../../lib/account/keyring';
 
 export function makeRegisterAttestorCommand() {
     const cmd = new Command('register');
     cmd.description('Register an attestor and bond funds from a stash account');
-    cmd.addOption(proxyForOption);
     cmd.addOption(attestorAddressOption.makeOptionMandatory());
```
atodorov, line 13: This still validates as a Substrate address. I think you should replace it with `evmAddressOption`.

cli/src/test/integration-tests/attestor/show-attestors-for.test.ts
```diff
@@ -144,8 +103,12 @@ describe('show-attestors-for', () => {
         }, 30_000);
 
         it('should display attestor address when passing stash address as argument', () => {
+            // The attestor-stash precompile records the stash as the EVM-derived
+            // AccountId (HashedAddressMapping from the caller's EVM address), not
+            // the sr25519 address. `show-attestors-for` matches on the recorded
+            // `stash` field, so query by `stash.evmStashAddress`.
             const result = CLI(
-                `attestor show-attestors-for --substrate-address ${stash.address} --chain ${chain_Anvil1_Key}`,
+                `attestor show-attestors-for --substrate-address ${stash.evmStashAddress} --chain ${chain_Anvil1_Key}`,
```
atodorov: Hmmm,
it's probably a better user experience to pass `--evm-address` directly here (and for all other attestor commands) instead of going through the trouble to convert to/from Substrate addresses.

atodorov (changes requested):

cli/src/test/integration-tests/attestor/show-attestors-for.test.ts
DylanVerstraete: Or we can just rename to `--address` instead

cli/src/commands/attestor/registerAttestor.ts
atodorov: @DylanVerstraete can you explain this item? I see it's been marked as resolved but neither code has changed, nor a comment was provided. Don't we want to pass EVM addresses to these commands ?

cli/src/test/integration-tests/attestor/show-attestors-for.test.ts
atodorov: FYI long ago we decided to move away from `--address` and use `--substrate-address` and `--evm-address` to make it more explicit what the options expect and to avoid confusion during development. Some commands may have diverged over time but the majority still follow this pattern.

cli/src/commands/attestor/registerAttestor.ts
DylanVerstraete: attestorAddress is a substrate account ID still, not evm address

.github/workflows/ci.yml
```diff
@@ -1204,9 +1204,11 @@ jobs:
...
           killall -9 creditcoin3-node
 
-  # preceeds all of the attestor-cli jobs so that we can
-  # easily trigger all of them when something fails
-  trigger-attestor-cli:
+  attestor-cli-testing:
+    # Attestor stash operations are dispatched through the EVM precompile, which
+    # uses the EVM `msg.sender` as origin via HashedAddressMapping. Substrate
+    # `pallet_proxy` delegation is not plumbed through the EVM, so there's no
+    # proxy matrix here — the stash signs directly.
```
Trantorian1, line 1211: Are we sure we want to run the attestor cli tests unconditionally on each PR? This will also trigger on unrelated changes, such as to the indexer, archiver and proofgen API server

Trantorian1 (dismissed):

cli/src/test/integration-tests/attestor/show-attestors-for.test.ts
DylanVerstraete: @atodorov I reworked the tests to show attestors for a given evm stash account, it shows all of the registered attestors under that key instead of one.

.github/workflows/ci.yml
DylanVerstraete: It's already gated. if:` ${{ needs.should-run.outputs.attestor_cli_testing == 'true' }}` (line 1215), backed by the attestor_cli_testing filter. Indexer/archiver/proofgen aren't in that filter, so PRs that only touch those don't trigger this job. The filter is intentionally a bit broad on runtime/** / common/** / precompiles/** because the precompiled artifacts those produce do affect the test, but the unrelated services you named don't.

didac-gluwa (approved):

Trantorian1 (approved):

beqaabu (approved):

cli/src/test/integration-tests/attestor/show-attestors-for.test.ts
atodorov: > @atodorov I reworked the tests to show attestors for a given evm stash account, it shows all of the registered attestors under that key instead of one.

@DylanVerstraete can you elaborate here? From what I can see the `show-attestors-for` command has been removed and its relevant tests.  How do we show all registered attestors for a given evm address ?

DylanVerstraete: @atodorov see 

https://github.com/gluwa/creditcoin3/pull/851/changes/8b8d9c7fd1cb9e7b86bafa2ae228e8530372d747

show-attestor-for is removed because it's a duplicate of show status now (shows status for all your attestors given a stash account (evm))

atodorov:
blockchain-tests timing out 2/2 with 

```
  ● Precompile: block-prover › Frontier calldata threshold (verifyAndEmit with calldata > 2893 bytes) › should succeed when calldata exceeds 2893-byte Frontier threshold

    thrown: "Exceeded timeout of 360000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      21 |
      22 | export const testIf = (condition: boolean, name: string, fn: any, timeout = 30000) =>
    > 23 |     condition ? test(name, fn, timeout) : test.skip(name, fn, timeout);
         |                 ^
      24 |
      25 | export const try_catch_else_finally = function (
      26 |     protected_code: any,

      at testIf (utils.ts:23:17)
      at blockchain-tests/precompiles/block-prover.test.ts:140:15
      at blockchain-tests/precompiles/block-prover.test.ts:131:5
      at Object.<anonymous> (blockchain-tests/precompiles/block-prover.test.ts:15:1)
```

doesn't seem to be related. Restarted again.

atodorov (approved):

PR #881, opened 2026-04-27
fix(pool): quorum panic

Trantorian1:

---

## Features

- [Recomputes quorums on target sample size change](https://github.com/gluwa/creditcoin3/pull/881/changes#diff-bb440cdeae707e0007839d405c32851ebafa706a67c9931a98d9c2d65753d8c0R1127-R1162).

## Fixes

- Rare bug in the attestation pool, where [shrinking the target sample size](https://github.com/gluwa/creditcoin3/blob/099a331c9dde023edf634053eab6faea9fd681d4/attestor/pool/src/lib.rs#L1288) then [invalidating a past quorum](https://github.com/gluwa/creditcoin3/blob/099a331c9dde023edf634053eab6faea9fd681d4/attestor/pool/src/lib.rs#L451) would cause a panic on a [missing `quorum_by_height`](https://github.com/gluwa/creditcoin3/blob/099a331c9dde023edf634053eab6faea9fd681d4/attestor/pool/src/lib.rs#L877-L880).

## Tests

- [Added regression tests](https://github.com/gluwa/creditcoin3/pull/881/changes#diff-bb440cdeae707e0007839d405c32851ebafa706a67c9931a98d9c2d65753d8c0R2188-R2224) to guard against this issue happening again in the future.
- [Added coverage for all cc3 attestation pool event handlers](https://github.com/gluwa/creditcoin3/pull/881/changes#diff-bb440cdeae707e0007839d405c32851ebafa706a67c9931a98d9c2d65753d8c0R2665-R2907).
- Added [`pretty_assertions`](https://docs.rs/pretty_assertions/latest/pretty_assertions/index.html) as a test dependency to help spot error diffs in large struct comparisons.

### DIFF
--- a/attestor/pool/src/lib.rs
+++ b/attestor/pool/src/lib.rs
@@ -1158,6 +1158,7 @@ impl AttestationPoolForks {
         }
 
         self.validate_quorum.target_quorum = quorum_new;
+        self.forks_best = self.find_best();
     }
 
     fn note_attestation_chain_reversion(&mut self, info: stream::util::AttestationInfo) {
@@ -2825,6 +2826,8 @@ mod test {
                 size: 1,
                 digest: attestation_2.compound_digest()
             }));
+
+            assert_eq!(forks.forks_best, Some(attestation_0.clone()));
         }
 
         sx.note_target_sample_size_change(1);
@@ -2843,6 +2846,8 @@ mod test {
                 size: 1,
                 digest: attestation_2.compound_digest()
             }));
+
+            assert_eq!(forks.forks_best, Some(attestation_0.clone()));
         }
     }
 
@@ -2885,6 +2890,8 @@ mod test {
                 size: 1,
                 digest: attestation_2.compound_digest()
             }));
+
+            assert_eq!(forks.forks_best, Some(attestation_2.clone()));
         }
 
         sx.note_target_sample_size_change(2);
@@ -2903,6 +2910,8 @@ mod test {
                 size: 1,
                 digest: attestation_2.compound_digest()
             }));
+
+            assert_eq!(forks.forks_best, Some(attestation_0));
         }
     }
 
--- a/attestor/pool/src/lib.rs
+++ b/attestor/pool/src/lib.rs
@@ -1124,6 +1124,42 @@ impl AttestationPoolForks {
             .note_attestation_interval_change(interval_new);
     }
 
+    /// Updating the target sample size is a sparse operation which may invalidate past quorums. As
+    /// a result, we discard all past quorums and recompute the set of quorums under the new target
+    /// sample size.
+    ///
+    /// While this has `O(n)` complexity, in practice this is a rare enough event that the cost of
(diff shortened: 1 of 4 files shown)


didac-gluwa (dismissed):

attestor/pool/src/lib.rs
```diff
@@ -1124,6 +1124,42 @@ impl AttestationPoolForks {
...
+    /// sample size.
+    ///
+    /// While this has `O(n)` complexity, in practice this is a rare enough event that the cost of
+    /// recomputing quorums is negligible once amortized.
+    fn note_target_sample_size_change(&mut self, quorum_new: std::num::NonZeroUsize) {
+        self.quorums_by_height.clear();
+
+        let key_start = KeySize {
+            size: quorum_new.into(),
+            height: 0,
```
DylanVerstraete, line 1140: height 0?

Trantorian1: Yep. This is the start key for the iterator over `forks_by_size`. Since `forks_by_size` is primarily indexed by size, then
```rust
        let index = (
            std::ops::Bound::Included(key_start),
            std::ops::Bound::Unbounded,
        );
```
will iterate over all forks with a size greater than or equal to the new quorum, irrespective of height or digest (since height is 0 and digest is min any value above that will be yielded by the iterator).

didac-gluwa (dismissed):

attestor/pool/src/lib.rs
```diff
@@ -2582,11 +2663,263 @@ mod test {
...
+                size: 2,
+                digest: attestation_0.compound_digest()
+            }));
+            assert!(forks.quorums_by_height.contains(&KeyHeight {
+                height: 2,
+                size: 1,
+                digest: attestation_2.compound_digest()
+            }));
+
+            assert_eq!(forks.forks_best, Some(attestation_0.clone()));
```
BradleyOlson64: Wouldn't we want this to be `attestation_2` after the target sample size is reduced to 1?

Because based on what I see in `attestation_pool_note_target_sample_size_grows` it appears that this is what we call "best":
1. Meets quorum threshold
2. Has highest block height among attestations which meet threshold

BradleyOlson64 (dismissed):
One question. Otherwise looks good!

DylanVerstraete (dismissed):

beqaabu (dismissed):

DylanVerstraete (dismissed):

attestor/pool/src/lib.rs
Trantorian1: Good catch! There was a bug in the implementation of `find_next` which would return the _lowest_ known quorum height instead of the _highest_. Went ahead and fixed that :+1:

beqaabu (dismissed):

DylanVerstraete (approved):

didac-gluwa (approved):

DylanVerstraete (approved):

PR #879, opened 2026-04-27
feat(attestor): enact chill on epoch rotation

DylanVerstraete:
# Description of proposed changes

PR #895, opened 2026-04-29
fix(chill): prevent quorum submission when inelligible

Trantorian1:

> [!NOTE]
> _Follow-up to_ #879 

##  Features

- Attestors share chill status between the production and validation worker.

##  Fixes

- Attestors no longer try to submit quorums if they are chilled.

## Known limitations

- Submission can still fail with "AttestorNotActive" if a runtime call to `commit_attestation` is made right before an attestor is chilled.

---

# Testing

To test this PR, start a local cc3 chain against anvil **with fast-runtime disabled**. Spawn zombienet and rotate the attestor set. wait for initial attestations to come in. Chill some attestors, update the target sample size accordingly, apply pending updates and force-rotate the attestor set. You will notice quorum is still observed and chilled attestors no longer try to commit attestations.

### DIFF
--- a/attestor/attestor/src/lib.rs
+++ b/attestor/attestor/src/lib.rs
@@ -372,12 +372,14 @@ impl Attestor {
 
         tracing::info!("⏳ [2/4] Starting attestation validation worker");
 
+        let can_attest = std::sync::Arc::new(std::sync::atomic::AtomicBool::new(true));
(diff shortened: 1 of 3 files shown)


Trantorian1 (commented):
Finished self-review.

DylanVerstraete (approved):

beqaabu (approved):

PR #898, opened 2026-04-29
Improve proof-gen-api health endpoint to account for archiver issues

didac-gluwa:
Adds a new method `is_healthy` to ethereum rpc interface and adjusts proof-gen server health check to fail if any of the configured archivers is having trouble

### DIFF
--- a/common/continuity/src/archiver.rs
+++ b/common/continuity/src/archiver.rs
@@ -169,6 +169,20 @@ impl EthRpcProvider for ArchiverEthProvider {
     async fn get_chain_id(&self) -> Result<u64> {
         self.eth_fallback.get_chain_id().await
     }
+
+    async fn is_healthy(&self) -> Result<bool> {
+        // Check both the archiver and the fallback RPC for health.
+        let archiver_healthy = self
+            .archiver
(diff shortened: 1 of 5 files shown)


common/continuity/src/archiver.rs
```diff
@@ -169,6 +169,20 @@ impl EthRpcProvider for ArchiverEthProvider {
     async fn get_chain_id(&self) -> Result<u64> {
         self.eth_fallback.get_chain_id().await
     }
+
+    async fn is_healthy(&self) -> Result<bool> {
```
DylanVerstraete, line 173: Better check would be is see if the last block from archiver is in line with the eth_fallback provider's one

didac-gluwa: But what if the archiver is slightly behind? Or the RPC? They would not really be unhealth in that scenario

DylanVerstraete: in practice they'll be pointing to the same eth rpc

DylanVerstraete (approved):

atodorov (approved):

PR #897, opened 2026-04-29
Add socket timeout error handling to Traffic Simulator

didac-gluwa:
Based on this error trace:

```sh
Batch Proof API request {
  url: "https://proof-gen-api.usc-testnet2.creditcoin.network/api/v1/proof-batch/1",
  queries: 3
}
    📦 Batch txs: 0x808a76f7..., 0xa8e7a401..., 0xc0858102..., 0x8b7a8e6c..., 0x0de571db...
Submitting batch { chainKey: 1, batchSize: 5 }
📦 Block 10755813 queued (143 txs, queue: 11)
❌ Failed to submit proof for 0x062060aa...: Single submit simulation failed: Single submit simulation timed out after 30000ms
📤 Submitting single proof for tx 0x73287df0... (block 10755780, index 64)
Fetching proof { chainKey: 1, block: 10755780, txIndex: 64 }
Proof API request {
  url: "https://proof-gen-api.usc-testnet2.creditcoin.network/api/v1/proof/1/10755780/64"
}
Submitting single proof { chainKey: 1, blockHeight: 10755780, txBytesLen: 3072 }
📦 Block 10755814 queued (139 txs, queue: 12)
    ❌ Batch failed (blocks 10755790-10755792): Batch submit simulation failed: request timeout (code=TIMEOUT, version=6.16.0)
📤 Submitting single proof for tx 0x808a76f7... (block 10755790, index 20)
Fetching proof { chainKey: 1, block: 10755790, txIndex: 20 }
Proof API request {
  url: "https://proof-gen-api.usc-testnet2.creditcoin.network/api/v1/proof/1/10755790/20"
}
error: Uncaught (in promise) Error: request timeout (code=TIMEOUT, version=6.16.0)
            error = new Error(message);
                    ^
    at makeError (file:///deno-dir/npm/registry.npmjs.org/ethers/6.16.0/src.ts/utils/errors.ts:698:21)
    at ClientRequest.<anonymous> (file:///deno-dir/npm/registry.npmjs.org/ethers/6.16.0/src.ts/utils/geturl.ts:66:24)
    at ClientRequest.emit (ext:deno_node/_events.mjs:439:20)
    at TLSSocket.emitRequestTimeout (node:_http_client:743:9)
    at Object.onceWrapper (ext:deno_node/_events.mjs:562:14)
    at TLSSocket.emit (ext:deno_node/_events.mjs:451:22)
    at TLSSocket._onTimeout (node:net:1024:8)
[...]

PR #899, opened 2026-04-29
feat(proof-api): add retry for eth rpc calls

DylanVerstraete:

Stability fix

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2610,6 +2610,7 @@ dependencies = [
  "sp-core",
  "thiserror 1.0.69",
  "tokio",
+ "tokio-retry",
  "tracing",
  "usc-abi-encoding",
  "user",
--- a/common/continuity/Cargo.toml
+++ b/common/continuity/Cargo.toml
@@ -14,6 +14,7 @@ async-trait = { workspace = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
+tokio-retry = { workspace = true }
 tracing = { workspace = true }
 
 # Creditcoin / USC dependencies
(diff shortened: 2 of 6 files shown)


DylanVerstraete (commented):
Rebased onto `usc-dev` and resolved the `common/continuity/src/rpc.rs` conflict (kept the `ReconnectingEthRpcProvider::run` retry wrapper, adopted the upstream `"Rpc error resolving tx position"` context wording from #897 so we don't lose the operator-triage clarification). `cargo fmt --check` and `cargo clippy -p continuity --all-targets --no-deps -- -D warnings` are clean locally.

Review notes:

* Retry logic (`ETH_RPC_MAX_ATTEMPTS = 3`, `RECONNECT_MAX_ATTEMPTS = 5` with exp backoff + jitter) mirrors the attestor's CC3 reconnect — consistent and easy to reason about.
* Off-by-one fix in commit 2 is correct: `tokio_retry::Retry::spawn` runs the action once and then drains the strategy iterator on each retry, so `take(N)` = 1 + N total attempts. `RECONNECT_MAX_ATTEMPTS.saturating_sub(1)` makes the constant mean total attempts and stays consistent with `ETH_RPC_MAX_ATTEMPTS`. 👍
* `reconnect()` only swaps `url`/`rpc_provider`/`chain_id` on `eth::Client` and leaves `cache` intact, so Redis block caching survives a reconnect — good.
* Minor: if N concurrent RPC calls all fail, each enters its own `reconnect()` and serializes on `client.write()`. They'll each try a fresh reconnect even after the first one succeeds. Not a correctness bug (the next attempt just re-validates a healthy client) but a small inefficiency — could be tightened later with a `Notify`/single-flight if it shows up in production logs. Not blocking.
* `eth::Client` is `Clone` and its inner `AlloyProvider` / `block_cache::Cache` are Arc-wrapped, so the `self.client.read().await.clone()` per call is cheap.

LGTM after rebase. Ready for human review.

beqaabu (approved):

atodorov (approved):

PR #903, opened 2026-04-30
feat(traffic-simulator): fee headroom + configurable tip + log-on-submit at info level

DylanVerstraete:
## What

Three related changes that together make the simulator more resilient to mempool contention and far easier to diagnose when txs get stuck.

### 1. Fee headroom (`FEE_HEADROOM_PERCENT`, default `100` → 2×)

`getFeeData()` returns a `maxFeePerGas` that tracks the **current** base fee. When the next block's base fee jumps, a tx broadcast at that level sits in the mempool until base fee falls back — manifesting as `confirmation timed out` from `tx.wait()`.

Applying a headroom multiplier on top of the node's suggestion is the standard fix and matches what production wallets do under the hood. Applied on:

- the EIP-1559 path (bumps **both** `maxFeePerGas` and `maxPriorityFeePerGas`)
- the legacy `gasPrice` path

`maxFeePerGas` is kept `>= tip` so ethers won't reject the tx. Set `FEE_HEADROOM_PERCENT=0` to opt out entirely.

### 2. Configurable tip floor + ceiling

- `MIN_PRIORITY_FEE_GWEI` is now env-driven (default `1`, same as before).
- New `MAX_PRIORITY_FEE_GWEI` caps the tip after the headroom multiplier — purely a safety net against misconfiguration; default `0` disables the cap to preserve current behavior.

### 3. Log fees at info level on broadcast

The `${label} fees` and `${label} sent` lines were `console.debug` — invisible at production log levels, which made stuck-tx diagnosis hard because we couldn't tell what fees we'd actually paid for the timed-out tx.

Both lines are now `console.info`, with fees rendered as gwei strings (not raw wei) and the `sent` log carrying `txHash`, `nonce`, `from`, **and the fees** together. Receipt timeouts can now be cross-referenced against the actual broadcast parameters without enabling debug logs.

## Why

Follow-up to #902. Working theory (raised by @beqaabu in <https://gluwa.slack.com|#C0AGC5PC6VC>) is that the `Single submit confirmation timed out` errors are caused by underpriced txs sitting in the mempool, not actual chain stalls. This PR makes the fee strategy resilient to base-fee jumps and makes the diagnostic data visible at prod log levels so we can confirm or rule that out from logs alone.

## Sample log output

Before (debug-only, raw wei):
```
[debug] Single submit fees { maxFeePerGas: 1500000000n, maxPriorityFeePerGas: 1000000000n }
[debug] Single submit sent { txHash: '0xa62432f5...', nonce: 4271 }
```
After (info, gwei, full context):
```
[info] Single submit fees { maxFeePerGas: '3.0 gwei', maxPriorityFeePerGas: '2.0 gwei' }
[info] Single submit sent { txHash: '0xa62432f5...', nonce: 4271, from: '0xabc...', maxFeePerGas: '3.0 gwei', maxPriorityFeePerGas: '2.0 gwei' }
```

## Breaking-ish change

The default 2× headroom raises the gas cost of submissions on the happy path. Operators on cheap chains who want the old behavior can set `FEE_HEADROOM_PERCENT=0`. Worth flagging because some cost dashboards may notice.

## Checks

- `deno fmt --check` ✅
- `deno lint` ✅
- `deno check src/main.ts` ✅

## Out of scope (deliberately separate)

- Stuck-tx rebroadcast loop (re-send same nonce with bumped fees after a watchdog instead of waiting for `RECEIPT_TIMEOUT_MS`). Real new logic, deserves its own review.

### DIFF
--- a/scripts/traffic-simulator/src/constants.ts
+++ b/scripts/traffic-simulator/src/constants.ts
@@ -51,7 +51,39 @@ export const SINGLE_PROOF_GAS_LIMIT = 5_000_000n;
 export const BATCH_PROOF_GAS_LIMIT = 10_000_000n;
 
 // Fee settings
-export const MIN_PRIORITY_FEE_GWEI = 1n;
+//
+// Tip floor (`MIN_PRIORITY_FEE_GWEI`): the minimum priority fee we'll attach
+// to a tx. Override via env for chains where 1 gwei is not enough to outbid
+// the mempool floor under contention.
+const _minPriority = Number(Deno.env.get("MIN_PRIORITY_FEE_GWEI"));
+export const MIN_PRIORITY_FEE_GWEI =
+  Number.isFinite(_minPriority) && _minPriority > 0
+    ? BigInt(Math.floor(_minPriority))
+    : 1n;
+
+// Tip ceiling (`MAX_PRIORITY_FEE_GWEI`): hard cap on the tip after the
+// headroom multiplier. Keeps a misconfigured headroom from producing
+// absurdly high tips. `0` disables the cap.
+const _maxPriority = Number(Deno.env.get("MAX_PRIORITY_FEE_GWEI"));
+export const MAX_PRIORITY_FEE_GWEI =
+  Number.isFinite(_maxPriority) && _maxPriority > 0
+    ? BigInt(Math.floor(_maxPriority))
+    : 0n;
+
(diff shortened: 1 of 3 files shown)


BradleyOlson64:
Closing for now. Traffic sim is working again, and we may want a different fix

PR #904, opened 2026-04-30
feat(traffic-simulator): dynamic gasLimit via estimateGas + balance preflight

DylanVerstraete:
## What

Stop shipping a fixed `5_000_000` (single) / `10_000_000` (batch) `gasLimit` on every submission. Both values are now treated as **ceilings** and **fallbacks**; the actual `gasLimit` is derived from `eth_estimateGas` + a configurable buffer + per-continuity-block padding, then floored and capped.

## Why this matters

A fixed-and-large `gasLimit` has two failure modes that match the `confirmation timed out after Nms` symptom we've been chasing in #902 / #903:

1. **Block-builder deprioritization.** Validators ordering by effective gas price *for the slot a tx takes up* will deprioritize fat txs in favor of skinnier ones at the same fee, even when the fat tx pays more in absolute terms. Yours sits in the mempool and `tx.wait()` eventually times out.
2. **Upfront balance check.** Nodes reject txs where `gasLimit * maxFeePerGas > sender.balance` even though the actual spend would be a fraction of that. Easy to trip on a low-balance signer with `gasLimit = 10_000_000`, *especially* after #903 raised `maxFeePerGas` headroom. On some Frontier-derived nodes the rejection manifests as silent non-propagation rather than a clean error.

### `submission/gas.ts` (new)
- `computeGasLimit()` — wraps `estimateGas` in `withTimeout(RPC_TIMEOUT_MS, ...)`, adds `GAS_BUFFER_PERCENT` (default 20%), adds `GAS_PER_CONTINUITY_BLOCK` (default `50_000`) per continuity root, floors at `MIN_DYNAMIC_GAS_LIMIT` (`200_000`), caps at the per-call ceiling, and **falls back to the ceiling on `estimateGas` errors** (with a warning) instead of crashing the caller.
- `checkBalanceForGas()` — pre-flight balance check that logs prominently when `balance < gasLimit * feeCap`. Deliberately doesn't throw (letting the node reject is a cleaner error path) but surfaces what would otherwise be invisible silent rejection.

### `constants.ts`
- New env vars:
  - `GAS_BUFFER_PERCENT` (default `20`)
  - `GAS_PER_CONTINUITY_BLOCK` (default `50000`)
  - internal `MIN_DYNAMIC_GAS_LIMIT = 200_000n` (not env-exposed; if you want a different floor, change the constant)
- `SINGLE_PROOF_GAS_LIMIT` / `BATCH_PROOF_GAS_LIMIT` are now documented as **ceilings and fallback values** rather than unconditional values. Same numbers as before, different role.

### `precompile.ts`
- `PrecompileParams`: `gasLimit (bigint)` replaced with `gasCeiling (bigint) + continuityBlocks (number)`.
- `executePrecompileCall()` runs `estimateGas` after fee data, logs estimated/padded/final gas at **info** level (same diagnostic philosophy as #903), runs the balance preflight, then broadcasts.
- `submitSingleToPrecompile` passes `continuityBlocks: 1`.
- `submitBatchToPrecompile` passes `continuityProof.roots.length` — batch verification cost scales with the number of continuity roots, not the batch size in tx count.

## Behavior change

Most submissions will now broadcast with a *smaller* `gasLimit` than before, which is the whole point — getting out of the heaviest gas-tier in the mempool/block-builder. The existing constants still apply as **hard ceilings**, so nothing can ship with a higher `gasLimit` than today. On `estimateGas` failure we fall back to the old constant value with a warning, so the worst-case behavior is the current behavior.

## Sample log output

```
[info] Single submit fees { maxFeePerGas: '3.0 gwei', maxPriorityFeePerGas: '2.0 gwei' }
[info] Single submit gas { estimated: '342118', padded: '460541', gasLimit: '460541', cappedAtCeiling: false, usedFallback: false }
[info] Single submit sent { txHash: '0xa62432f5...', nonce: 4271, ... }
```

When estimation fails:
```
[warn] ⚠️  Single submit: estimateGas failed (execution reverted), falling back to ceiling=5000000
[info] Single submit gas { estimated: '<fallback>', padded: '<fallback>', gasLimit: '5000000', cappedAtCeiling: false, usedFallback: true }
```

When balance is tight:
```
[warn] ⚠️  Single submit: signer balance 1234567890 < required upfront 18421640000000000 (gasLimit=460541 * feeCap=40000000000)
```

## Stacking with #902 and #903

- #902 (receipt timeout bump + enriched logs) — independent, mergeable in any order.
- #903 (fee headroom + log-on-submit) — independent. If this PR succeeds in eliminating the timeouts, the 2× headroom in #903 may be belt-and-braces overkill; we can dial it back via `FEE_HEADROOM_PERCENT` without code changes.
- This PR (#904) — independent. Branched off `usc-dev`, not off the others.

## Checks

- `deno fmt --check` ✅
- `deno lint` ✅
- `deno check src/main.ts` ✅

## Out of scope

- Stuck-tx rebroadcast loop (re-send same nonce with bumped fees after a watchdog) — separate concern, keep tracking it.
- Persisting an EWMA of estimateGas results to skip the RPC call on the hot path — possible follow-up if estimateGas latency becomes a bottleneck.

### DIFF
--- a/scripts/traffic-simulator/src/constants.ts
+++ b/scripts/traffic-simulator/src/constants.ts
@@ -47,8 +47,37 @@ export const MAX_TRANSIENT_RETRIES = 3;
 export const TRANSIENT_RETRY_BASE_DELAY_MS = 2_000;
 
 // Gas limits
+//
+// These are now treated as **ceilings** rather than unconditional values.
+// Submission code calls `eth_estimateGas` and pads the result; the
+// constants below cap that estimate so a misbehaving node can't hand us
+// an absurd value, and act as a fallback if estimation itself fails.
 export const SINGLE_PROOF_GAS_LIMIT = 5_000_000n;
 export const BATCH_PROOF_GAS_LIMIT = 10_000_000n;
+
+// Floor on the dynamically-computed gasLimit. Anything below this is
+// almost certainly a bad estimate (e.g. Frontier returning 21000 for a
+// precompile call) and we want to reject it rather than ship a tx that
+// will out-of-gas mid-call.
+export const MIN_DYNAMIC_GAS_LIMIT = 200_000n;
+
+// Buffer (`GAS_BUFFER_PERCENT`) applied on top of `eth_estimateGas`.
+// Default 20%. Bigger proofs and Frontier RPC sometimes underestimate by
+// more than EVM does, so we pad before capping.
+const _gasBuffer = Number(Deno.env.get("GAS_BUFFER_PERCENT"));
+export const GAS_BUFFER_PERCENT = Number.isFinite(_gasBuffer) && _gasBuffer >= 0
+  ? BigInt(Math.floor(_gasBuffer))
+  : 20n;
+
+// Per-continuity-block extra gas (`GAS_PER_CONTINUITY_BLOCK`). Continuity
+// verification cost scales with the number of roots in the continuity
+// proof, and `estimateGas` doesn't always capture that growth precisely
+// for large batches. Default 50k per block.
+const _gasPerBlock = Number(Deno.env.get("GAS_PER_CONTINUITY_BLOCK"));
+export const GAS_PER_CONTINUITY_BLOCK =
+  Number.isFinite(_gasPerBlock) && _gasPerBlock >= 0
+    ? BigInt(Math.floor(_gasPerBlock))
+    : 50_000n;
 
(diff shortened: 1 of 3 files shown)


BradleyOlson64:
Putting this one on pause, as the traffic sims are working currently. May still move ahead with it eventually.

PR #908, opened 2026-05-01
fix(attestation): prevalidate commit_attestation in txpool

DylanVerstraete:
## Features

- Added new [`PrevalidateAttestationCommit`](https://github.com/gluwa/creditcoin3/pull/908/changes#diff-a6ec5058aab39c46685a26b6019341b9c00dfd3cd103ba5ecebc9d07876858cd) extension which prevents duplicate attestations from active attestors to be inserted into the mempool.
- Senders which are not in the active attestor set still have their call included (even if it is invalid) so they end up paying a runtime fee as a protection against DOS.

## Fixes

- Prevents active attestors from paying for duplicate `commit_attestation` calls as a result of submission races. Multiple attestors race to submit the same attestation for the sake of redundancy: we don't want those who lose the race to be penalized for participating in network security, and we don't want this action to be a slow drain of attestor wallets.


pallets/attestation/src/extensions.rs
```diff
@@ -0,0 +1,137 @@
...
+{
+    #[cfg(feature = "std")]
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "PrevalidateAttestationCommit")
+    }
+
+    #[cfg(not(feature = "std"))]
+    fn fmt(&self, _: &mut core::fmt::Formatter) -> core::fmt::Result {
+        Ok(())
+    }
```
Trantorian1: These impls do not have to be separate, `write!` is also defined for non-std usage in `core`
https://doc.rust-lang.org/stable/core/macro.write.html

pallets/attestation/src/extensions.rs
```diff
@@ -0,0 +1,137 @@
...
+
+            // Cheap storage read: the active attestor set per chain.
+            let active_attestors = ActiveAttestors::<T>::get(chain_key)
+                .into_iter()
+                .collect::<BTreeSet<_>>();
+            if !active_attestors.contains(who) {
+                return Err(TransactionValidityError::Invalid(
+                    InvalidTransaction::Custom(VALIDITY_ERROR_NOT_ACTIVE_ATTESTOR),
+                ));
+            }
```
Trantorian1: This makes it free for non-attestors to spam the chain. We should still forward such calls so they can pay an inclusion fee.

pallets/attestation/src/tests.rs
```diff
@@ -7543,3 +7543,155 @@ mod migrate_attestors_count_v1_to_v2 {
...
+                PrevalidateAttestationCommit::<Test>::new().validate(&ATTESTOR_2, &call, &info, 0);
+
+            assert_eq!(
+                result,
+                Err(TransactionValidityError::Invalid(
+                    InvalidTransaction::Custom(VALIDITY_ERROR_NOT_ACTIVE_ATTESTOR)
+                ))
+            );
+        })
+    }
```
Trantorian1: This should not cause a transaction rejection, for the reasons mentioned prior.

pallets/attestation/src/tests.rs
```diff
@@ -7543,3 +7543,155 @@ mod migrate_attestors_count_v1_to_v2 {
...
+
+    #[test]
+    fn rejects_duplicate_attestation() {
```
Trantorian1: This should also test to make sure that attestors are refunded any expenses if they loose the submission race.

pallets/attestation/src/tests.rs
```diff
@@ -7543,3 +7543,155 @@ mod migrate_attestors_count_v1_to_v2 {
...
+                None,
+                None,
+            );
+
+            // First commit makes the attestation a checkpoint, so reusing the
+            // same header_number triggers the post-checkpoint duplicate path.
+            assert_ok!(Attestation::commit_attestation(
+                attestor.attestor_origin.clone(),
+                attestation.clone()
+            ));
```
Trantorian1: This should also assert to make sure the attestation is in storage in order to validate the assumptions of the test at this point.

pallets/attestation/src/tests.rs
```diff
@@ -7543,3 +7543,155 @@ mod migrate_attestors_count_v1_to_v2 {
...
+            let info = call.get_dispatch_info();
+
+            let result = PrevalidateAttestationCommit::<Test>::new().validate(
+                &attestor.attestor_id,
+                &call,
+                &info,
+                0,
+            );
+
+            assert!(result.is_ok(), "expected Ok, got {result:?}");
```
Trantorian1: It is less error-prone to validate against on-chain storage.

pallets/attestation/src/tests.rs
```diff
@@ -7543,3 +7543,155 @@ mod migrate_attestors_count_v1_to_v2 {
...
+
+            let result =
+                PrevalidateAttestationCommit::<Test>::new().validate(&ATTESTOR_2, &call, &info, 0);
+
+            assert!(
+                result.is_ok(),
+                "unrelated calls should pass through: {result:?}"
+            );
+        })
+    }
```
Trantorian1: This test is not needed.

Trantorian1 (changes requested):
Looks promising, but we need more tests to validate that this indeed does not charge _any_ funds, not just that it rejects transactions.

pallets/attestation/src/extensions.rs
DylanVerstraete: TX is never included, so they don't need to pay I think. Obviously they'll use some resources, but I think it's negligible

pallets/attestation/src/extensions.rs
DylanVerstraete: Good call — collapsed into a single `core::write!` impl in bc14b55b, with a comment noting that `core::write!` works in both `std` and `no_std`.

pallets/attestation/src/extensions.rs
DylanVerstraete: Holding on this one for now per Dylan's reply above — the call still costs the spammer the resources of being signed and gossiped, and the CPU we burn validating the signature is bounded by mempool admission limits and is comparable to the work we'd do running the call to its on-chain `AttestorNotActive` failure (just earlier and without a fee debit on the *legitimate* attestor flow). Forwarding non-attestor calls would re-introduce the original drain vector for any attestor whose stash signs a stale or malformed commit (e.g. during the same chilled-window race we're trying to close), which is what the extension exists to prevent.

If you'd rather make this configurable (e.g. `ForwardForFee` mode) I'm happy to wire that as a follow-up, but I'd prefer to keep the default closed. Leaving the thread open for you to weigh in.

pallets/attestation/src/tests.rs
DylanVerstraete: Tied to the thread above on `extensions.rs:126` — keeping the rejection (and therefore this assertion) for now, pending your call there. The test now also asserts the caller's free balance is unchanged on rejection (`validate_without_charging`), which is the property that motivates the rejection in the first place.

pallets/attestation/src/tests.rs
DylanVerstraete: Added a dedicated `race_loser_pays_no_fee` test in bc14b55b that:

1. Spins up two active attestors (winner + loser) for the same chain.
2. Lets the winner commit the attestation on-chain (asserting the digest lands in `Attestations` storage).
3. Snapshots the loser's free balance, runs the loser's submission through the SignedExtension, asserts it is rejected as `Stale`, and asserts the loser's free balance is *unchanged* — i.e. the txpool admission path costs them nothing.

The other existing rejection tests now also flow through a `validate_without_charging` helper that pins the same balance-unchanged invariant, so all rejection paths are covered.

pallets/attestation/src/tests.rs
DylanVerstraete: Done in bc14b55b — added an `assert!(Attestations::<Test>::contains_key(chain_key, digest))` precondition right after the first commit, so the duplicate-rejection branch is exercised against actual on-chain state rather than the local closure's expectations.

pallets/attestation/src/tests.rs
DylanVerstraete: Updated in bc14b55b — the passthrough test now asserts against `ActiveAttestors::<Test>::get(chain_key).contains(&attestor_id)` and `!Attestations::<Test>::contains_key(chain_key, digest)` before calling `validate`, so the test reads the same on-chain state the extension does.

pallets/attestation/src/tests.rs
DylanVerstraete: Removed in bc14b55b — agreed, the `is_sub_type` short-circuit for non-`commit_attestation` calls is trivial enough that the test was just exercising rustc.

DylanVerstraete:
Pushed two follow-up commits per the chat thread:

**bc14b55b — review fixes from @Trantorian1's pass**
- Single `core::write!` `Debug` impl, no more `cfg(std)` split.
- Duplicate-rejection test now asserts the original attestation is in storage before validating the duplicate.
- Passthrough test now asserts against `ActiveAttestors` / `Attestations` storage rather than the local closure.
- New `validate_without_charging` helper threaded through every rejection test, asserting the caller's free balance is untouched on `Err` from `validate`.
- New end-to-end `race_loser_pays_no_fee` test: two active attestors, winner commits, loser's submission is rejected as `Stale` and balance is unchanged. This is the 'no funds are charged' invariant @Trantorian1 asked to test.
- Dropped `ignores_unrelated_calls` (the `is_sub_type` short-circuit was trivial).
- Holding on the 'forward non-attestor calls so they pay' suggestion per Dylan's reply — happy to make it configurable as a follow-up if you'd rather, but I think the default-closed posture is the safer one for this PR's scope.

**76c6c5ed — attestor-side handling for the new reject path**
The runtime extension now rejects `commit_attestation` at txpool admission for chilled / stale cases. Substrate maps those to JSON-RPC 1010 `Invalid Transaction`, surfaced through subxt as `RpcError::ClientError`. The existing submit loop in `attestor/src/worker/validation/mod.rs` treated *every* `Err` from `sign_and_submit_then_watch_default` as a transport failure and called `reconnect`, so race-loser/chilled submissions would loop indefinitely against perfectly healthy RPC nodes (this matches the connection loop @eliotmcnab1 was seeing). Inlined the prevalidation match per @eliotmcnab1's note (no `is_txpool_prevalidation_reject` helper): we detect the 1010 message, log it as an info-level skip, and return `Ok(())` for that height. All other RPC errors flow through the existing reconnect path unchanged.

Local status:
- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p pallet-attestation --no-deps --all-targets` — clean.
- `cargo clippy -p attestor --no-deps` — clean.
- `cargo test -p pallet-attestation` — 228 passed, 0 failed.
- `cargo check -p creditcoin3-runtime` — clean.
- `cargo check -p creditcoin3-node` still fails to link `librocksdb-sys` in this sandbox; will validate on CI.

pallets/attestation/src/extensions.rs
```diff
@@ -69,14 +69,10 @@ impl<T: Config + Send + Sync> Debug for PrevalidateAttestationCommit<T>
...
-    #[cfg(feature = "std")]
     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
-        write!(f, "PrevalidateAttestationCommit")
-    }
-
-    #[cfg(not(feature = "std"))]
-    fn fmt(&self, _: &mut core::fmt::Formatter) -> core::fmt::Result {
-        Ok(())
+        // `core::write!` works in both `std` and `no_std` builds, so we don't
+        // need to split the impl on the `std` feature flag.
```
Trantorian1: Useless comment.

pallets/attestation/src/tests.rs
```diff
@@ -7574,6 +7574,26 @@ mod prevalidate_attestation_commit_extension {
         RuntimeCall::Attestation(crate::Call::commit_attestation { attestation })
     }
 
+    /// Runs `validate` and asserts that the caller's free balance is
+    /// untouched, capturing the property reviewers care about: a rejected
+    /// transaction must not drain the would-be sender. `validate` itself is
+    /// pure, but pinning the invariant in tests guards us against accidental
+    /// future side-effects in this extension.
```
Trantorian1: Useless comment.

attestor/attestor/src/worker/validation/mod.rs
```diff
@@ -988,6 +988,25 @@ impl WorkerAttestationValidation {
...
+                // by the runtime's `PrevalidateAttestationCommit` SignedExtension
+                // (chilled attestor or stale/duplicate digest). Substrate maps
+                // those to JSON-RPC error code 1010 with an `"Invalid
+                // Transaction"` message, surfaced through subxt as
+                // `RpcError::ClientError`. Such rejections are permanent for
+                // this height — the wallet was *not* charged (that's the whole
+                // point of the extension) and reconnecting would just spin
+                // until backoff, masking the real outcome. Skip softly.
+                Err(subxt::Error::Rpc(subxt::error::RpcError::ClientError(ref client_err)))
+                    if client_err.to_string().contains("Invalid Transaction") =>
```
Trantorian1: This seems very brittle. Is there no other way to match on the error type then to perform a string search?

Trantorian1 (changes requested):

pallets/attestation/src/extensions.rs
DylanVerstraete: Removed in ce284258.

pallets/attestation/src/tests.rs
DylanVerstraete: Removed in ce284258.

attestor/attestor/src/worker/validation/mod.rs
DylanVerstraete: Agreed, fixed in ce284258. Replaced the string match with a structured downcast through subxt's re-exported jsonrpsee:

```rust
use subxt::ext::jsonrpsee::core::client::Error as JsonrpseeClientError;
const POOL_INVALID_TX: i32 = 1010; // sc_rpc_api::AUTHOR_BASE (1000) + 10

// ...
Err(err) => {
    if let subxt::Error::Rpc(subxt::error::RpcError::ClientError(boxed)) = &err {
        if let Some(JsonrpseeClientError::Call(obj)) =
            boxed.downcast_ref::<JsonrpseeClientError>()
        {
            if obj.code() == POOL_INVALID_TX {
                // skip softly
            }
        }
    }
    // existing reconnect path
}
```

The boxed error inside `subxt::Error::Rpc(RpcError::ClientError(_))` is canonically a `jsonrpsee::core::client::Error`, and substrate emits `POOL_INVALID_TX = AUTHOR_BASE (1000) + 10 = 1010` from `sc-rpc-api/src/author/error.rs` for *every* txpool admission failure — including any `SignedExtension` rejection. Matching the JSON-RPC code is the canonical hook, so this should be stable across substrate releases as long as that constant doesn't change.

Trantorian1 (dismissed):

Trantorian1 (changes requested):
Working on this locally. A lot of the tests don't make sense or don't work as intended on closer inspection. Needs manual intervention.

pallets/attestation/src/extensions.rs
```diff
@@ -0,0 +1,117 @@
...
+        if let Some(Call::commit_attestation { attestation }) = call.is_sub_type() {
+            let chain_key = attestation.chain_key();
+
+            let active_attestors = ActiveAttestors::<T>::get(chain_key)
+                .into_iter()
+                .collect::<BTreeSet<_>>();
+            if active_attestors.contains(who) && Pallet::<T>::check_duplicate(attestation) {
+                return Err(TransactionValidityError::Invalid(InvalidTransaction::Stale));
+            }
+        }
```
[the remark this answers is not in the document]
Trantorian1: This behavior has changed as it has been deemed better for non-attestors to pay for inclusion of invalid calls as a protection against DOS.

pallets/attestation/src/tests.rs
```diff
@@ -7543,3 +7543,222 @@ mod migrate_attestors_count_v1_to_v2 {
...
+                &attestor_2.attestor_id,
+                &call,
+                &info,
+                0,
+            );
+
+            assert_eq!(
+                res,
+                Err(TransactionValidityError::Invalid(InvalidTransaction::Stale))
+            );
```
atodorov, line 7697: Q: is this error going to propagate upwards ? If yes, it's probably cleaner to just call `commit_attestation()` inside `assert_err()` and assert on the TransactionValidityError directly.

pallets/attestation/src/tests.rs
```diff
@@ -7543,3 +7543,222 @@ mod migrate_attestors_count_v1_to_v2 {
...
+                &account.attestor_id,
+                &call,
+                &info,
+                0,
+            );
+
+            assert!(
+                res.is_ok(),
+                "Attestation from an invalid source must still be inserted into the mempool"
+            );
```
atodorov, line 7740: Q: do we pay fees here? Not asserted explicitly.

atodorov (commented):
LGTM, added couple comments wrt tests.

pallets/attestation/src/tests.rs
Trantorian1: The error doesn't actually take place in `commit_attestation`, but _prior_ to the attestation dispatch to `commit_attestation` by the node mempool, hence it is not sufficient to call `commit_attestation`. We validate the filtering behavior of [`PrevalidateAttestationCommit`](https://github.com/gluwa/creditcoin3/pull/908/changes#diff-a6ec5058aab39c46685a26b6019341b9c00dfd3cd103ba5ecebc9d07876858cd) instead.

pallets/attestation/src/tests.rs
Trantorian1: Fees are not payed during mempool validation, but as part of runtime inclusion, which is being checked [here](https://github.com/gluwa/creditcoin3/blob/81603bc41891d9cc73cc87f5460071ff72d2661e/pallets/attestation/src/tests.rs#L7752-L7756).

atodorov (approved):

Trantorian1 (approved):

beqaabu (approved):

PR #910, opened 2026-05-01
feat(api): key per block range overrides

DylanVerstraete:

Adds an ordered list of **fallback RPC URLs** per source chain in `proof-gen-api-server`. The eth client always tries the primary `eth_rpc_url` first, then walks `eth_rpc_fallback_urls` in declaration order whenever the primary returns "not found" or a transport error for a block fetch / receipts fetch / tx-hash lookup.

This replaces the earlier per-block-range override approach (initial commit on this branch), which had a hard limitation: tx-hash lookups have no block number to route on, so a tx mined in the "archive" range would 404 against the recent-only primary even though the archive endpoint had it.

## Why

Each source chain is in practice served by two very different access patterns:

- **Hot path** (recent blocks, near tip): high QPS, latency-sensitive, cheap on most providers.
- **Cold path** (deep historical blocks for back-fills / catch-up + tx-hash lookups for old txs): low QPS but requires archive support, metered separately, 5–50× more expensive.

A single `eth_rpc_url` forces a tradeoff: pay for archive on every call, or 404 on old data. With an ordered fallback list the primary stays cheap and recent-only, and the (typically more expensive) archive endpoint only gets queried when the primary can't answer.

Because tx-hash lookups carry no block-number information, the fallback walk is the *only* way to make `/proof-by-tx` work for old transactions — which was exactly the failure mode that motivated this change.

## How it works

- `ChainConfig` gains `eth_rpc_fallback_urls: Vec<String>`.
- `eth::Client` gains `new_with_fallbacks` and `new_with_cache_and_fallbacks` (cache-aware variant). On startup each fallback URL is connected and its `chain_id` is verified against the primary — mismatch fails fast.
- Block-keyed and tx-hash calls (`get_eth_block`, `get_receipts`, `get_tx_position_by_hash`) now walk providers via `providers_with_labels()`:
  - Order is `[primary, fallback[0], fallback[1], ...]`.
  - First `Ok(Some(_))` wins; remaining providers are not queried.
  - If at least one provider returns `Ok(None)`, the result is "not found" (peers that errored are demoted to `WARN`).
  - If every provider errors, the **first** error is returned and the rest are logged at `WARN`.
- Tip-side calls (`subscribe`, `get_last_block`) stay on the primary only — fallbacks are a historical-routing concern.
- `reconnect()` re-connects every fallback URL too and re-asserts `chain_id` parity, so a recovered primary can't silently keep using a stale fallback socket.
- Cache integration: `block_cache::get_block` short-circuits on hit; on miss, it calls `try_fetch_block`, which transparently uses the fallback walk. Cached entries store the merged result regardless of which provider answered.

## Logging

- Per-call logs are silent when the primary answers (the boring path), so we don't spam logs for deployments with no fallbacks.
- `INFO` log line emitted when a fallback serves a request, with provider label (`fallback[i]@<base-url>?…`), block number / tx hash, and the index/position. Operators can confirm at a glance which provider handled which request.
- `WARN` for demoted errors (peer errored but another said `not found`, or peer errored but another succeeded).
- `ERROR` only when **every** provider failed.
- All logged URLs go through `eth::redact_url_query`, which scrubs both query strings *and* secret-shaped path segments:
  - Anything after `?` is replaced with `?…` (covers `?key=...` style — Google Cloud, Infura legacy).
  - Any path segment that looks like an opaque token — ≥ 16 chars of `[A-Za-z0-9_]`, no hyphens, contains both a digit and a letter — is replaced with `…` (covers Chainstack-style trailing path keys, Alchemy `/v2/<KEY>`, etc.). Length is not leaked.
  - Human-readable identifiers (`v1`, `projects`, `ethereum-sepolia`, `cc3-testnet-rpckey-2`) stay intact — they're either short, hyphenated, or pure-alpha/pure-digit, so the heuristic ignores them.
  - The previous duplicate copy in `proof-gen-api-server` is removed; both crates now share the canonical eth-side helper to prevent drift.

## Validation (startup, no network I/O for the structural part)

- Each URL must be non-empty (whitespace is trimmed).
- Each URL must be unique within a chain's fallback list (duplicates would just hit the same endpoint twice and silently double the cost on a miss).
- Then per-URL: `chain_id` parity check against the primary at connect time.
- Errors include the offending `chain_key` and the offending fallback index.

## Usage

YAML (`config.example.yaml`):

```yaml
chains:
  - chain_key: 1
    eth_rpc_url: "https://recent.example/v2/<KEY_RECENT>"
    eth_rpc_fallback_urls:
      - "https://archive.example/v2/<KEY_ARCHIVE>"
```

- `eth_rpc_fallback_urls` is fully optional. Empty list ≡ pre-PR behavior (primary-only).
- Any number of fallbacks is allowed; they are tried in declared order.
- Available **only** via the multi-chain YAML path. The legacy single-chain CLI/env mode (`--eth-rpc-url`) stays single-URL by design.

## Tests

- `proof-gen-api-server::config` — 9 unit tests: empty list, empty/whitespace URL rejection, duplicate URL rejection, whitespace trimming, declaration order preservation, YAML round-trip with and without fallbacks, YAML invalid-fallback parse error.
- `eth::provider_lookup_tests` — covers the `merge_provider_lookup` decision matrix and `redact_url_query`:
  - At least one `None` → `NotFound` and peer errors demoted to warnings.
  - `None` with no errors → simple `NotFound`.
  - All errors → first error promoted, rest demoted to warnings.
  - Empty input panics (defensive — caller always passes non-empty).
  - Redaction: `?key=...` strip; short paths pass through; Chainstack-style trailing hex token redacted; Alchemy `/v2/<KEY>` redacted; googleapis path tokens kept intact while `?key=` is stripped; pure-letter and pure-digit long segments are kept (not key-shaped); inner secret segment with surrounding fixed segments is redacted in place.

## Notes / things reviewers may want to push back on

- Fan-out is **sequential**, not parallel. Trade-off: minimizes calls to expensive archive endpoints (fast 90% case never reaches them), but adds latency on the not-found-on-primary path. Parallel fan-out is a future option if latency on cold-path lookups becomes a concern.
- Primary-only writes / subscriptions are intentional — fallbacks are a *read-side historical safety net*, not a load balancer.
- No per-URL health / circuit breaker yet. If a fallback is consistently unreachable it'll keep getting tried. Worth revisiting once we have observability on fallback-served-rate.

---

- [x] All GitHub Actions report PASS
  - [x] Coverage tools report all newly added lines as covered
  - [x] The positive scenario is exercised
  - [x] Negative scenarios are exercised, e.g. assert on all possible errors
  - [ ] Assert on events triggered if applicable (n/a — no on-chain events)
  - [ ] Assert on changes made to storage if applicable (n/a)
- [x] Modified behavior/functions - try to make sure above test items are covered

### DIFF
--- a/common/eth/src/lib.rs
+++ b/common/eth/src/lib.rs
@@ -292,32 +292,15 @@ pub enum ConnectionTransport {
     Ws(WsConnect),
 }
 
-/// One per-block-range RPC URL override.
+/// One ordered fallback RPC provider attached to a [`Client`].
 ///
-/// Block-keyed RPC requests (block fetch, receipts) whose `block_number`
-/// falls within the inclusive range `[from_block, to_block]` are routed to
-/// `url` instead of the [`Client`]'s default URL. Either bound may be `None`
-/// to mean "open-ended on that side".
-///
-/// All override URLs must point to a node that reports the same `chain_id`
-/// as the default URL — this is verified when the [`Client`] is constructed.
-///
-/// Constraints (enforced at construction time):
-/// * Each override must specify at least one of `from_block` or `to_block`
-///   (otherwise it would silently shadow the default URL).
-/// * If both bounds are set, `from_block <= to_block`.
-/// * Overrides must not overlap each other.
+/// The [`Client`] always tries its primary URL first; on `Ok(None)` or
+/// transport errors it walks `fallback_providers` in declared order and
+/// uses the first non-empty answer. All fallback URLs must point to a
+/// node that reports the same `chain_id` as the primary URL — this is
+/// verified when the [`Client`] is constructed.
 #[derive(Debug, Clone)]
-pub struct RpcRangeOverride {
-    pub from_block: Option<u64>,
-    pub to_block: Option<u64>,
-    pub url: String,
-}
-
-#[derive(Debug, Clone)]
-pub(crate) struct RangeProvider {
-    pub(crate) from: u64,
-    pub(crate) to: u64,
+pub(crate) struct FallbackProvider {
     pub(crate) url: Url,
     pub(crate) provider: AlloyProvider,
 }
@@ -327,9 +310,10 @@ pub struct Client {
     url: Url,
     private_key: Option<String>,
     rpc_provider: AlloyProvider,
-    /// Block-range RPC overrides, sorted by `from`. Empty = no overrides
-    /// (legacy single-URL behavior). See [`RpcRangeOverride`].
-    range_providers: Vec<RangeProvider>,
+    /// Ordered fallback RPC providers. Empty = no fallbacks (single-URL
+    /// behavior). Each provider is tried in declared order whenever the
+    /// primary returns `Ok(None)` or errors. See [`FallbackProvider`].
+    fallback_providers: Vec<FallbackProvider>,
     // what chain id is implied here? Maybe need to define internal chain ids for different attestation chains
     // and not rely on ethereum chain ids?
     chain_id: u64,
@@ -377,37 +361,38 @@ impl Client {
             url,
             private_key: private_key.map(|s| s.to_owned()),
             rpc_provider,
-            range_providers: Vec::new(),
+            fallback_providers: Vec::new(),
             chain_id,
             #[cfg(feature = "block_cache")]
             cache: None,
         })
     }
 
-    /// Build a [`Client`] with per-block-range RPC URL overrides.
+    /// Build a [`Client`] with ordered fallback RPC URLs.
     ///
-    /// `url` is the default URL used for tip-related calls (subscriptions,
-    /// `eth_blockNumber`, `eth_getTransactionByHash`) and for any block whose
-    /// number is not covered by an entry in `overrides`.
+    /// `url` is the primary URL: tried first for every operation, and the
+    /// only URL used for tip-related calls (subscriptions, `eth_blockNumber`,
+    /// `eth_chainId`).
     ///
-    /// All override URLs are connected to at startup; each must report the
-    /// same `chain_id` as `url` and the override ranges must satisfy the
-    /// validity rules documented on [`RpcRangeOverride`].
+    /// `fallback_urls` are tried in declaration order whenever the primary
+    /// returns `Ok(None)` or a transport error for a block fetch / tx-hash
+    /// lookup. Each fallback URL is connected to at startup and must report
+    /// the same `chain_id` as `url`.
     ///
-    /// When `overrides` is empty this is equivalent to [`Client::new`].
-    pub async fn new_with_overrides(
+    /// When `fallback_urls` is empty this is equivalent to [`Client::new`].
+    pub async fn new_with_fallbacks(
         url: &str,
-        overrides: &[RpcRangeOverride],
+        fallback_urls: &[String],
         private_key: Option<&str>,
     ) -> anyhow::Result<Self> {
         let (url, rpc_provider, chain_id) = Self::init_rpc(url).await?;
-        let range_providers = Self::init_range_providers(chain_id, overrides).await?;
+        let fallback_providers = Self::init_fallback_providers(chain_id, fallback_urls).await?;
 
         anyhow::Ok(Self {
             url,
             private_key: private_key.map(|s| s.to_owned()),
             rpc_provider,
-            range_providers,
+            fallback_providers,
             chain_id,
             #[cfg(feature = "block_cache")]
             cache: None,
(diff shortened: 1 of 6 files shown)


atodorov:
See failure at https://github.com/gluwa/creditcoin3/actions/runs/25435092836/job/74614267857?pr=910

didac-gluwa:
I cannot agree with this PR, seems unnecessarily complicated and doesn't really solve any problem. For example: What if all fallback RPCs also fail? 

We already have the reconnecting logic in https://github.com/gluwa/creditcoin3/pull/917

If we have so many issues with RPCs we should solve it in that side of the connection not complicate our code with the assumption that RPCs are gonna be unreliable

DylanVerstraete:
> I cannot agree with this PR, seems unnecessarily complicated and doesn't really solve any problem. For example: What if all fallback RPCs also fail?
> 
> We already have the reconnecting logic in #917
> 
> If we have so many issues with RPCs we should solve it in that side of the connection not complicate our code with the assumption that RPCs are gonna be unreliable

It's more so that we can specify an archive endpoint as a fallback; if the default key can't do a transaction lookup, the fallback key is used. This decreases load on actual expensive platforms (infura, chainstack,..) and offloads all recent lookups to google (free api keys)

DylanVerstraete:
> See failure at https://github.com/gluwa/creditcoin3/actions/runs/25435092836/job/74614267857?pr=910

seems flaky because I restarted and it passed?

common/eth/src/lib.rs
```diff
@@ -353,13 +402,78 @@ impl Client {
     pub async fn reconnect(&mut self) -> Result<(), Error> {
         let (url, rpc_provider, chain_id) = Self::init_rpc(self.url.as_ref()).await?;
 
+        // Reconnect each fallback against its own URL too, otherwise a
+        // recovered primary would silently keep using a stale fallback
+        // socket until that fallback hit a fault on its own.
+        let mut new_fallbacks = Vec::with_capacity(self.fallback_providers.len());
+        for fp in &self.fallback_providers {
+            let (fp_url, fp_provider, fp_chain_id) = Self::init_rpc(fp.url.as_ref()).await?;
```
BradleyOlson64: Right now the whole reconnection attempt fails if we fail to reconnect one of our fallbacks.

Could it be better to allow the reconnection to succeed if our primary is healthy?

We could still log a loud error or continue attempting to reconnect the fallbacks separately.

proof-gen-api-server/bin/server.rs
BradleyOlson64, line 168: Maybe default to 1 here instead?

proof-gen-api-server/config.example.yaml
```diff
@@ -17,6 +17,22 @@ chains:
     # archiver_url: "http://localhost:8080"  # optional per chain
     # block_confirmation_depth: 0  # Blocks to lag behind chain tip for reorg protection.
```
BradleyOlson64: Should have a suggested value other than 0 for this. Perhaps 32. 

I'm going to wire the defaults to come from our source chain configs in pallet supported chains.

BradleyOlson64 (dismissed):
A couple thoughts, but looks good 👍

proof-gen-api-server/bin/server.rs
DylanVerstraete: Done in 9ffba0ec9438d55964d0b19060a3d7231baaa148 — `CHAIN_KEY` now defaults to `1` in legacy single-chain mode, and the comment in `config.example.yaml` was updated to match.

proof-gen-api-server/config.example.yaml
DylanVerstraete: Bumped to `block_confirmation_depth: 32` in 9ffba0ec9438d55964d0b19060a3d7231baaa148 with a comment noting it as a suggested value for Ethereum-style L1 chains and that the omitted default of 0 is only safe for local/dev. Happy to drop this entirely once defaults are wired from `pallet-supported-chains`.

common/eth/src/lib.rs
DylanVerstraete: Addressed in 9ffba0ec9438d55964d0b19060a3d7231baaa148: fallback reconnect is now best-effort. If a fallback fails to reconnect (transport error or chain_id mismatch), we keep the previous provider in place, log a loud `error!`, and let the next primary-failure path retry it. A misbehaving backup no longer takes down a healthy primary on reconnect.

common/eth/src/lib.rs
```diff
@@ -353,13 +402,84 @@ impl Client {
...
+            ));
+        }
+        out
+    }
+
+    /// Returns the configured fallback URLs in declaration order. Useful
+    /// for startup logging.
+    pub fn fallback_urls(&self) -> Vec<&Url> {
+        self.fallback_providers.iter().map(|fp| &fp.url).collect()
+    }
```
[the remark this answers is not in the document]
DylanVerstraete: Already removed — see `34373c2d chore: remove dead code` (`pub fn fallback_urls` getter is gone).

common/eth/src/lib.rs
```diff
@@ -353,13 +402,102 @@ impl Client {
...
+    ) -> anyhow::Result<Vec<FallbackProvider>> {
+        let mut providers: Vec<FallbackProvider> = Vec::with_capacity(fallback_urls.len());
+        for (idx, raw_url) in fallback_urls.iter().enumerate() {
+            let (url, provider, chain_id) =
+                Self::init_rpc(raw_url.as_ref()).await.with_context(|| {
+                    format!(
+                        "Failed to connect to fallback RPC URL #{idx} ({})",
+                        redact_url_query(raw_url),
+                    )
+                })?;
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in c30088e86b16ae6f714965fec6d0295b131c4984 — the `info!` in `init_rpc` now runs the URL through `redact_url_query`, so primary and fallback connects (`init_fallback_providers` / `reconnect`) no longer leak path-embedded API keys at INFO level.

common/eth/src/lib.rs
```diff
@@ -494,28 +635,133 @@ impl Client {
...
+            {
+                Ok(Some(receipts)) => {
+                    if label != "primary" {
+                        info!(
+                            provider = %label,
+                            block_number = number,
+                            "receipts fetched via fallback provider"
+                        );
+                    }
+                    return Ok(receipts);
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in aa0d94ce0c4c18a813f794b3055f1ccea6a21e07 — `get_receipts`, `get_eth_block`, and `get_tx_position_by_hash` now drain the accumulated `errors` vec on `Ok(Some(_))` and log each entry at `WARN` (`provider` / `served_by` / `error` fields) before returning. The 'peer errored but another succeeded' path is now actually wired up, so a flaky primary masked by a healthy fallback still shows up in operator logs.

beqaabu (dismissed):

proof-gen-api-server/bin/server.rs
```diff
@@ -161,11 +161,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
         c
     } else {
-        // Legacy single-chain mode: chain key from CHAIN_KEY env (default 2).
+        // Legacy single-chain mode: chain key from CHAIN_KEY env (default 1).
         let chain_key = env::var("CHAIN_KEY")
             .ok()
             .and_then(|s| s.parse().ok())
-            .unwrap_or(2);
+            .unwrap_or(1);
```
atodorov: I think this is the reason all examples failed:

```
Error: Wrong chain for chain_key 1. Source chain endpoint chain id: 31337, Supported chain id: 1
```
when starting prover.

Is the change of this default strictly necessary ?

atodorov (commented):
https://github.com/gluwa/creditcoin3/actions/runs/25547171455/job/74988938859?pr=910
```
Run git apply --allow-empty /var/tmp/new-metadata.diff
error: cli/creditcoin.json: patch does not apply
error: the patch applies to 'common/cc-client/artifacts/metadata.scale' (1e81682af7d3c72df09aa1288478ae8ee0c68285), which does not match the current contents.
error: common/cc-client/artifacts/metadata.scale: patch does not apply
Error: Process completed with exit code 1.
```

Probably needs a rebase.

atodorov (approved):

common/eth/src/lib.rs
```diff
@@ -548,42 +820,422 @@ impl Client {
...
+            if looks_like_secret_segment(seg) {
+                "…"
+            } else {
+                seg
+            }
+        })
+        .collect::<Vec<_>>()
+        .join("/");
+
+    format!("{redacted_path}{query_marker}")
```
beqaabu, line 1066: hmm is there a chance these keys have `-` in them? if so we would be just logging them. the comment below says it doesnt happen in practice but I kinda doublt it
> // Hyphenated multi-word identifiers (e.g. `cc3-testnet-rpckey-2`,
    // `ethereum-sepolia`) are never secrets in practice — providers don't
    // shape API keys that way. This single rule covers all the
    // human-readable path tokens we've seen in proof-gen configs.

beqaabu (commented):
overall looks good, just one Q

common/eth/src/lib.rs
DylanVerstraete: probably not, never seen it

beqaabu (approved):

PR #918, opened 2026-05-05
chore: add pruning attestations after checkpoint created event in pro…

beqaabu:
…of gen cache

# Description of proposed changes

<describe what this PR is about and why we want it>

---

Practical tips for PR review & merge:

- [ ] All GitHub Actions report PASS
- [ ] Newly added code/functions have unit tests
  - [ ] Coverage tools report all newly added lines as covered
  - [ ] The positive scenario is exercised
  - [ ] Negative scenarios are exercised, e.g. assert on all possible errors
  - [ ] Assert on events triggered if applicable
  - [ ] Assert on changes made to storage if applicable
- [ ] Modified behavior/functions - try to make sure above test items are covered
- [ ] Integration tests are added if applicable/needed

PR #924, opened 2026-05-06
fix(proof-gen-api): bracket query with mixed cache halves

DylanVerstraete:
## Problem

After #918 landed, `CheckpointReached` triggers `prune_attestations_at_or_below(checkpoint_height)`, which drops every attestation `<= checkpoint_height` from the proof-gen-api cache. The pruning itself is correct — the on-chain pallet evicts those entries from `AttestationRemovalQueues` once they exceed `AttestationRetentionDuration`, and serving proofs anchored on stale attestation digests would surface as `"Continuity proof does not match attestation or checkpoint"` at verify time.

The pruning leaves a structural gap, though. Once a checkpoint at block `H` lands, a query for the very first attestation block strictly above `H` (and before the next checkpoint lands) sees:

```
attestation_cache: lower=None,  upper=<query block itself>
checkpoint_cache:  lower=H,     upper=None
```

`build_continuity` tries the attestation cache, then falls back to the checkpoint cache, then bails. Each cache alone fails to bracket the query, so the request is rejected even though the two halves together describe a valid `(lower, upper)` anchor pair.

Production dump that motivated this:

```
min_query=max_query=10797810
checkpoint cache: greatest_strictly_before=Some(10797800), smallest_at_or_after=None
attestation cache: greatest_strictly_before=None,           smallest_at_or_after=Some(10797810)
```

## Fix

Add a third fallback arm `get_mixed_boundaries` to `build_continuity`. It reads both caches and returns the first viable cross-cache pair:

1. `(checkpoint_lower, attestation_upper)` — preferred because the attestation upper is tighter
2. `(attestation_lower, checkpoint_upper)` — symmetric fallback

Soundness: both caches store `(block_number, on_chain_digest)` pairs that anchor the same digest sequence. A checkpoint at block `H` and an attestation at block `H` reference the same on-chain root. `build_proof_from_roots` only requires:
- `lower < min_query` (anchor we extend from)
- `upper >= max_query` (anchor we verify against)

It does not care which cache the anchor came from, and it already reconciles the upper digest after building (`helpers.rs:127-145`), so a faulty cross-cache pair would be rejected by the same path that catches a faulty same-cache pair.

No pallet, retention, or prune-semantics changes. Read-only on both caches.

## Why not loosen the prune?

Considered and rejected. The prune doc-comment already explains it: any attestation we keep past the on-chain queue eviction window points at a digest the verifier no longer recognizes. Mirroring `AttestationRetentionDuration` exactly would couple the off-chain cache to per-chain mutable governance state, race the pallet's eviction, and only paper over the bug — wherever the post-prune attestation floor sits, the same `range(..floor)` empty-set issue still exists at that floor block. The fix here is independent of retention and closes the gap rather than reducing its frequency.

## Tests

Six new regression tests in `helpers.rs`:

- `split_bracket_each_cache_alone_fails` — sanity check that the production shape genuinely defeats both single-cache lookups
- `split_bracket_mixed_lookup_combines_cp_lower_with_att_upper` — exact reproduction of the dump (block `10_797_810`)
- `split_bracket_mixed_lookup_falls_back_to_att_lower_with_cp_upper` — symmetric case
- `split_bracket_mixed_lookup_prefers_attestation_upper_when_both_have_upper` — tie-breaker (tighter upper wins)
- `split_bracket_mixed_lookup_returns_none_when_neither_half_pair_exists` — no-anchor case
- `split_bracket_mixed_lookup_does_not_mutate_caches` — read-only invariant; locks out future "backfill" refactors

All 22 tests in the module pass; `cargo fmt` and `cargo clippy --tests -D warnings` clean.

### DIFF
--- a/proof-gen-api-server/src/services/continuity_service/helpers.rs
+++ b/proof-gen-api-server/src/services/continuity_service/helpers.rs
@@ -26,7 +26,10 @@ impl ContinuityService {
     /// Build a continuity proof.
     ///
     /// Flow:
-    /// 1. Resolve checkpoint boundaries from local cache (attestation first, then checkpoint)
+    /// 1. Resolve checkpoint boundaries from local cache. Try attestation
+    ///    bounds first (tightest), then checkpoint bounds, then a mixed
+    ///    cross-cache bracket as a final fallback for the steady-state gap
+    ///    that follows a checkpoint prune (see [`ContinuityService::get_mixed_boundaries`]).
     /// 2. Build proof from eth provider roots (archiver or chain)
     pub(crate) async fn build_continuity(
         &self,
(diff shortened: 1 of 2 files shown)


beqaabu (approved):
tested this locally and it works fine for the edgecase aswell as the other cases

atodorov:
> tested this locally and it works fine for the edgecase aswell as the other cases

Can you add the instructions as a comment and/or ticket. IMO it could be useful to add this as a follow-up and run the tool maybe even against live targets.

beqaabu:
> > tested this locally and it works fine for the edgecase aswell as the other cases
> 
> Can you add the instructions as a comment and/or ticket. IMO it could be useful to add this as a follow-up and run the tool maybe even against live targets.

opened https://gluwa.atlassian.net/browse/CSUB-2040

PR #937, opened 2026-05-07
feat(attestation): forward_patch_checkpoints extrinsic and recovery s…

DylanVerstraete:
## Why

The archiver and attestor network produced incorrect roots because we did not respect the source chain's finalization delay — checkpoints got committed against blocks that later reorged. The bad ladder is now on-chain and proofs built against it are wrong; the proof-gen-api-server cannot recover on its own. We need an operator-side surgical tool to overwrite the bad checkpoint ladder with the correct one and clear the stale attestation pipeline that fed it.

## What this PR adds

**1. New operator extrinsic `attestation.forward_patch_checkpoints(chain_key, wipe_suffix, checkpoints)`** (call index 29, `OperatorsOrigin`):

- Overwrites or inserts a batch of `AttestationCheckpoint`s (bounded by `MaxCheckpointsImportedPerCall`).
- When `wipe_suffix = true`, removes every checkpoint strictly above the batch tip from `Checkpoints` and `CheckpointBuckets`, bounded by `MAX_CHECKPOINT_SUFFIX_WIPE_TOTAL = 4096` per dispatch (guarded by `CheckpointSuffixWipeTooLarge`).
- Updates `LastCheckpoint`: with `wipe_suffix` it always tracks the batch tip; without it, only advances when the batch tip is higher than the current tip, so partial backfill batches can't accidentally rewind the head.
- Emits `CheckpointReached` per inserted height plus a single `ForwardCheckpointPatchApplied { chain_key, wiped_suffix, tip_block_number }`.
- Refuses to run when checkpoint pruning / clearing / bucket-clearing is already in progress for the chain (`CheckpointMaintenanceInProgress`) and on empty input (`EmptyCheckpointPatch`).
- All-or-nothing: wrapped in `#[transactional]`.

**2. Stale-attestation purge before any rewrite (`purge_attestations_for_forward_patch`)**

This is the key fix for the broken-roots incident. Before the new ladder is written we clear, for the target `chain_key`:

- `Attestations` — bounded `clear_prefix` loop, batch 128 × max 512 iterations = up to 65 536 entries per dispatch; if more remain, the call aborts with `TooManyAttestationsForForwardPatchClear` so the operator can split the work.
- `CheckpointingQueues`, `AttestationRemovalQueues`, `LastDigest`.

Without this, attestors would keep advertising a digest chain anchored on the bad ladder and the patched checkpoints would immediately be contradicted again. Bonding ledger entries and authorized-attestor sets are intentionally left intact — attestors resume committing under the patched ladder once the recovery batches are in.

**3. Recovery script: `attestor/scripts/ForwardPatchCheckpoints.js`**

Documented two-phase recipe matching the runtime semantics:

1. *Anchor + wipe* — single tx with `wipeSuffix: true` and one trusted checkpoint at `--anchor-height`; this deletes every checkpoint above the anchor and clears the attestation pipeline.
2. *Forward ladder* — submit ascending batches of size 100 from `checkpoints.json` for heights above the anchor; only the final batch sets `wipeSuffix: true` so anything beyond the imported tip is removed.

Retries with finalized-status confirmation, sudo or direct operator signing (`USE_SUDO=0`), `--dev` flag for short timing on devnet.

## Files touched

- `pallets/attestation/src/lib.rs` — new dispatchable, error variants, event variant, `WeightInfo` entry.
- `pallets/attestation/src/impls.rs` — `do_forward_patch_checkpoints`, `purge_attestations_for_forward_patch`, bounds constants.
- `pallets/attestation/src/benchmarking.rs` — benchmark for the new extrinsic.
- `pallets/attestation/src/weights.rs` — generated weight entry.
- `pallets/attestation/src/tests.rs` — coverage (see below).
- `attestor/scripts/ForwardPatchCheckpoints.js` — recovery script.

## Tests

`pallets/attestation/src/tests.rs` — new coverage:

- `forward_patch_checkpoints_overwrites_and_keeps_higher_last_when_no_wipe`
- `forward_patch_checkpoints_wipe_suffix_updates_tip` (also confirms higher checkpoints are deleted)
- Empty / oversize-suffix / maintenance-in-progress error paths
- `TooManyAttestationsForForwardPatchClear` guard

## Notes for reviewers

- High-risk per Cursor's auto-summary; the safety surface is `OperatorsOrigin` + the explicit bounds (`MAX_CHECKPOINT_SUFFIX_WIPE_TOTAL`, attestation clear loop bounds, `MaxCheckpointsImportedPerCall`).
- Recovery is intentionally heavy (state writes scale with batch + suffix size + attestations purged); use the script's chunking, don't hand-craft a giant batch.
- Clean merge against `usc-dev`; no rebase needed.

## Checklist

- [x] All GitHub Actions report PASS (in flight at time of writing — will confirm before merge)
  - [x] Positive scenarios exercised
  - [x] Negative scenarios exercised (empty batch, oversize suffix, maintenance-in-progress, too-many-attestations)
  - [x] Asserts on emitted events (`ForwardCheckpointPatchApplied`, `CheckpointReached`)
  - [x] Asserts on storage changes (`Checkpoints`, `CheckpointBuckets`, `LastCheckpoint`, `Attestations`, `LastDigest`)
- [ ] Integration test against a live recovery flow (manual via `ForwardPatchCheckpoints.js` on devnet)


pallets/attestation/src/impls.rs
```diff
@@ -1251,6 +1260,110 @@ impl<T: Config> Pallet<T> {
...
+            block_number: batch_max,
+            digest: *tip_digest,
+        };
+        let new_last = match LastCheckpoint::<T>::get(chain_key) {
+            Some(lc) if !wipe_suffix && lc.block_number > batch_tip.block_number => lc,
+            _ => batch_tip,
+        };
+        LastCheckpoint::<T>::insert(chain_key, new_last.clone());
+
+        Self::deposit_event(Event::<T>::ForwardCheckpointPatchApplied {
```
atodorov, line 1358: Q: do cc3-indexer and/or prover need to listen to this event ?

pallets/attestation/src/lib.rs
```diff
@@ -1089,6 +1101,32 @@ pub mod pallet {
...
+        /// Clears **[`Attestations`]**, **[`CheckpointingQueues`]**, **[`AttestationRemovalQueues`]**, and
+        /// **[`LastDigest`]** for this `chain_key` first so stale attestations cannot contradict the patched
+        /// ladder (see [`crate::pallet::Pallet::purge_attestations_for_forward_patch`]).
+        ///
+        /// Does **not** unregister attestors or alter bonding ledger entries — attestors resume committing
+        /// attestations after recovery.
+        ///
+        /// When `wipe_suffix` is true, every checkpoint strictly above the batch tip is removed in this
+        /// dispatch (bounded by [`crate::impls::MAX_CHECKPOINT_SUFFIX_WIPE_TOTAL`]).
+        #[pallet::call_index(29)]
```
atodorov: nitpick: since this is call index 29 it would be great if it's actually right after index 28, not in betwene 18-19.

atodorov (commented):
Added couple of questions. 

Need to wait for this to regenerate metadata.

Trantorian1 (dismissed):

beqaabu (dismissed):

atodorov (dismissed):
Please rebase. This PR causes lots of CI job failures across the board but more importantly the new matadata patch cannot be applied:

```
error: cli/creditcoin.json: patch does not apply
error: the patch applies to 'common/cc-client/artifacts/metadata.scale' (ef9bd849da8a7a89ac15767ed42644cd7414ff06), which does not match the current contents.
error: common/cc-client/artifacts/metadata.scale: patch does not apply
```

which leads me to believe this is out of date.

pallets/attestation/src/impls.rs
DylanVerstraete: Might need to patch the indexer, currently nothing really relies on indexer data and it's a one off, not sure what to do

For the prove, we just restart that component anyway

Trantorian1 (dismissed):

pallets/attestation/src/impls.rs
atodorov: As long as we can resync indexer so that its internal data matches the on-chain state after patching these checkpoints IMO it's OK to have it as a follow-up item. 

However from what I can see the events will be coming in reverse order. First there will be a bunch of `CheckpointReached` events and the indexer will trigger on them.  It will either update existing records, or crash on them or more likely just add duplicates.

And secondly there's the `ForwardCheckpointPatchApplied` event which comes at the end, so even if indexer wants to remove some old records how does it know whether they are trully old ones or haven't just been updated/inserted based on event trigger 10 lines above. 

Maybe you need to emit events when purching the attestations and that would be the trigger for indexer, not necessarily the `ForwardCheckpointPatchApplied` event.

beqaabu (dismissed):

cc3-indexer/src/mappings/mappingHandlers.ts
```diff
@@ -364,6 +364,9 @@ export async function handleEventCheckpointReached(event: SubstrateEvent): Promi
     const chainKeyStr = chainKey.toString();
     const chainKeyNumber = BigInt(chainKeyStr);
 
+    // Replace any prior indexer rows for this checkpoint height (on-chain overwrites are unique per height).
+    await remove_checkpoints_at_checkpoint_height(chainKeyNumber, checkpoint.blockNumber);
+
```
atodorov, line 373: You already have remove_attestations_above_height() and remove_checkpoints_above_height() functions which will remove exactly what we need. (maybe call it with checkpoint.blockNumber-1).

Then on every subsequent iteration they won't remove anything b/c records have already been removed (assuming we import the new checkpoints in lowest -> highest order).

cc3-indexer/src/mappings/mappingHandlers.ts
```diff
@@ -1329,6 +1332,40 @@ export async function handleForcedElection(event: SubstrateEvent): Promise<void>
...
+            data: [chainKey, wipedSuffix, tipBlockNumber],
+        },
+    } = event;
+
+    const chainKeyNumber = BigInt(chainKey.toString());
+    const wipedSuffixBool = codecToBool(wipedSuffix);
+    const tipBlock = BigInt(tipBlockNumber.toString());
+
+    // Mirrors `purge_attestations_for_forward_patch`: chain clears all attestations for this chain key.
+    await remove_all_attestations_for_chain(chainKeyNumber);
```
atodorov: Why not remove_attestations_above_height() ?

beqaabu (dismissed):

cc3-indexer/src/mappings/mappingHandlers.ts
```diff
@@ -1329,6 +1332,40 @@ export async function handleForcedElection(event: SubstrateEvent): Promise<void>
...
+
+    // When suffix was wiped, chain removed checkpoints strictly above the batch tip.
+    if (wipedSuffixBool) {
+        await remove_checkpoints_above_height(chainKeyNumber, tipBlock);
```
atodorov, line 1359: The comparison in this function is
```
            if (BigInt(checkpoint.blockNumber.toString()) > revertHeight) {
```

Not sure if you want to retain checkpoint at tipBlock or not!

cc3-indexer/src/mappings/mappingHandlers.ts
DylanVerstraete: https://github.com/gluwa/creditcoin3/pull/937#discussion_r3201214045

cc3-indexer/src/mappings/mappingHandlers.ts
DylanVerstraete: yes we do

cc3-indexer/src/mappings/mappingHandlers.ts
atodorov: The whole point of these comments was that there are functions which already perform these removals. You don't need to create new ones for the same task. 

If you want to remove all attestations that's fine by me, just call `remove_attestations_above_height(chainKeyNumber, 0) ` for example.

cc3-indexer/src/mappings/mappingHandlers.ts
DylanVerstraete: remove_checkpoints_above_height(chainKey, checkpoint.blockNumber − 1) runs on every CheckpointReached, not only forward-patch. That drops every stored checkpoint ≥ that height. That matches a wiped suffix when the batch is emitted low→high, but it’s wrong for forward_patch_checkpoints with wipe_suffix == false, where the runtime keeps checkpoints above batch_max. So we only clear above the tip when ForwardCheckpointPatchApplied has wiped_suffix, and only replace rows at each emitted height otherwise.

cc3-indexer/src/mappings/mappingHandlers.ts
```diff
@@ -1329,6 +1336,42 @@ export async function handleForcedElection(event: SubstrateEvent): Promise<void>
     await forcedElection.save();
 }
 
+export async function handleEventForwardCheckpointPatchApplied(event: SubstrateEvent): Promise<void> {
```
atodorov, line 1339: Need to whitelist this handler, otherwise the CI job will report it wasn't executed. 

I've added CSUB-2041 to exercise it via integration tests.

.github/check-cc3-indexer-handlers.sh
```diff
@@ -10,24 +10,61 @@ if [ -z "$MONITOR_FILE" ]; then
     exit 2
 fi
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+HANDLER_WHITELIST_FILE="$SCRIPT_DIR/cc3-indexer-handler-whitelist.txt"
+
 HANDLERS_FROM_SOURCE=$(grep handler: cc3-indexer/datasources.ts | tr -d ' ",' | tr -d "'" | cut -f2 -d: | sort | uniq)
```
atodorov: Gees, you could have just filtered this handler out on this line. The rest is really an overkill. 

If it fails we can update. If it doesn't leave it as-is and I will revert and fix it as part of CSUB-2041.

beqaabu (approved):

atodorov (approved):
Debugging/fixing handleTransactionVerified in #922. 

The rest LGTM.

PR #958, opened 2026-05-13
fix(audit): direct-link attestations

Trantorian1:

Local attestation validation now supports attestation intervals of 1.

## Fixes

- [audit item 6](https://github.com/chainlight-io/2026-q1-ctc-usc-phase2-audit-issues/issues/6)

### DIFF
--- a/common/streams/attestation/src/lib.rs
+++ b/common/streams/attestation/src/lib.rs
@@ -267,9 +267,11 @@ impl StreamAttestation {
         );
 
         let continuity_proof = attestor_primitives::block::ContinuityProof::from_blocks(blocks);
-
-        let prev_digest = (!continuity_proof.is_empty())
-            .then(|| continuity_proof.compute_continuity_digest(block_first));
+        let prev_digest = if self.attestation_interval.get() == 1 {
+            Some(self.attestation_prev.digest)
+        } else {
+            Some(continuity_proof.compute_continuity_digest(block_first))
+        };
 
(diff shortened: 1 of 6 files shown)


DylanVerstraete (approved):

common/streams/attestation/src/lib.rs
```diff
@@ -267,9 +267,11 @@ impl StreamAttestation {
...
 
         let continuity_proof = attestor_primitives::block::ContinuityProof::from_blocks(blocks);
-
-        let prev_digest = (!continuity_proof.is_empty())
-            .then(|| continuity_proof.compute_continuity_digest(block_first));
+        let prev_digest = if self.attestation_interval.get() == 1 {
+            Some(self.attestation_prev.digest)
+        } else {
+            Some(continuity_proof.compute_continuity_digest(block_first))
+        };
```
didac-gluwa, line 274: Is this distinction really necessary? If our interval is only 1 the compute will only have a single root which will end up computing the same digest as the `self.attestation_prev.digest` no?

I guess it is a sort of optimization to just skip it. Maybe I would add a comment about it

didac-gluwa (approved):

common/streams/attestation/src/lib.rs
Trantorian1: If the interval is 1 the continuity proof is actually empty, since attestations link end-to-end. In that case we can't retrieve the `prev_digest` from the continuity proof directly.

PR #976, opened 2026-05-15
fix(attestation): enforce BLS public key uniqueness per chain

creditcoinprotoclaw:
## Summary

Closes a quorum-collapse bug in `pallet-attestation` where multiple attestor controller accounts could register the same BLS public key on the same chain, letting one signer satisfy the threshold-of-`N` quorum on their own.

## The attack

`start_attesting` (`pallets/attestation/src/impls.rs::425`) only verifies a BLS proof-of-possession for the supplied public key; it does **not** enforce that the key is unique across attestors on the same chain. The same `(bls_pubkey, PoP)` pair can therefore be registered by N distinct controller accounts.

`validate_attestation`:
- `eligible_attestors` is a `BTreeSet<AccountId>` — counts logical identities, not BLS signers.
- `gather_attestor_public_keys` returns a `Vec<BlsPublicKey>` — duplicates allowed.
- `aggregate_public_keys(...)` sums them. BLS aggregation is linear, so aggregating the same key `k` three times yields `3k`, and one signer holding `s` (with `s*G = k`) can produce a single signature `σ` whose triple `3σ` verifies under `3k`.

With the default `TargetSampleSize = 3` (runtime/src/lib.rs:881) and `calculate_threshold = 3` (primitives/attestor/src/lib.rs:349-352), three colluding controller accounts aliased to a single BLS private key satisfy the quorum with one real signer. Downstream `BlockProver` users rely on the quorum reflecting independent signers, so this is high severity.

### 1. Registration-time uniqueness gate

New storage `BlsKeyOwner: (chain_key, BlsPublicKey) -> AccountId`. `start_attesting` now:
- rejects a key already owned by a different controller with `Error::BlsKeyAlreadyRegistered`;
- on key rotation by an idle attestor, releases the old slot before claiming the new one;
- treats re-asserting the same key for the same controller as idempotent.

The slot is **retained through retirement** so attestations referencing the retired controller can still be verified for the unbond delay; it is released when the retired entry is purged in `withdraw_unbonded` / `clear_retired_bls_entry_for_controller`, and in `on_supported_chain_removed`.

### 2. Validation-time dedup

`validate_attestation` now collects the gathered BLS public keys into a `BTreeSet<BlsPublicKey>`, requires `unique_keys >= threshold`, and aggregates only over the deduplicated set. Catches pre-migration storage, migration drift, and retired-key edge cases. New error `Error::InsufficientUniqueSigners`.

### 3. Storage migration `v2 -> v3`

`MigrateBlsKeyUniquenessV2ToV3`:
- iterates `Attestors`, sorts by `(chain_key, attestor_id)`;
- the lowest-`AccountId` attestor wins the `BlsKeyOwner` claim per `(chain_key, bls_pubkey)`;
- "losers" have their `bls_public_key` cleared, status forced to `Idle`, and are removed from `ActiveAttestors`;
- each demotion emits `Event::DuplicateBlsKeyDetectedDuringMigration` for ops visibility;
- bumps `STORAGE_VERSION` to 3.

Wired into the runtime's `SingleBlockMigrations` chain after the existing V1 -> V2.

### Migration policy — please review

I chose **auto-clear losers** rather than emit-only-with-an-operator-extrinsic. Reasoning:

- This is a security migration that fixes a quorum-poisoning bug. Leaving duplicates in storage keeps the attack live until ops intervene per-attestor.
- The runtime-level dedup check already rejects attestations from a colliding set, so the auto-clear only changes which attestor can re-register the key going forward. Operationally the losers must re-register fresh keys regardless.
- Deterministic ordering by `AccountId` makes the resolution reproducible and reviewable; without registration timestamps in storage there is no better tiebreaker.

Affected attestors are **not slashed** and retain bond/ledger/staking state — only their BLS key claim and active set membership are reset. They re-run `attest` with a freshly generated unique BLS keypair.

If the team prefers the operator-extrinsic alternative (collisions logged + event-only, with a `force_clear_duplicate_bls_key` gated on `OperatorsOrigin`), I can swap the policy in this PR — the migration scaffolding stays the same.

## Tests

Added `mod bls_key_uniqueness` covering:

- registration rejection when a second attestor reuses an active BLS key on the same chain;
- per-chain independence (same key registered on a different chain succeeds);
- idle-attestor key rotation: old slot released, new slot claimed;
- `validate_attestation` rejects an attestation built from three distinct controllers aliased to one BLS key (with a real one-signer aggregate signature) with `InsufficientUniqueSigners`;
- happy path: three attestors with three distinct keys produces a valid attestation;
- migration v2 -> v3: collision case keeps the lowest-AccountId attestor and clears the loser; noop case when already at v3.

All 240 pallet-attestation tests pass locally (`cargo test -p pallet-attestation --lib`).

## Build status

- `cargo check -p pallet-attestation` ✅
- `cargo check -p pallet-attestation --features try-runtime` ✅ (includes new `pre_upgrade` / `post_upgrade` for the migration)
- `cargo test -p pallet-attestation --lib` ✅ — 240 passed
- `cargo clippy -p pallet-attestation --all-targets -- -D warnings` ✅
- `cargo clippy -p creditcoin3-runtime --all-targets -- -D warnings` ✅
- `cargo fmt --all` ✅

Linking the full workspace test binary may exceed sandbox limits; relying on CI for the cross-pallet test run.

## Reviewer attention

- Migration policy choice (see above).
- `BlsKeyOwner` retention through retirement: please double-check the cleanup points in `clear_retired_bls_entry_for_controller`, `purge_retired_bls_keys_for_stash`, and `on_supported_chain_removed` for any path I missed where a BLS key is dropped from `Attestors` without dropping the owner map entry.
- The dedup loop in `validate_attestation` re-encodes each `PublicKey` to bytes for set membership. `BlsPublicKey` is `[u8; 48]` so it implements `Ord` directly; this is the simplest correct approach but if you'd rather hash or compare PublicKey objects directly, happy to switch.


pallets/attestation/src/tests.rs
```diff
@@ -8010,3 +8010,358 @@ mod prevalidate_attestation_commit_extension {
...
+                Attestation::attest(
+                    RuntimeOrigin::signed(att2.attestor_id),
+                    SUPPORTED_CHAIN_KEY,
+                    att1.public_key,
+                    att1.signature,
+                ),
+                Error::<Test>::BlsKeyAlreadyRegistered
+            );
+
+            // BLS-key-owner unchanged, att2 still keyless.
```
atodorov: I don't get this comment. The assertion below is for `att1` but not for `att2`.  perhaps you are missing some extra assertions here ?

pallets/attestation/src/tests.rs
```diff
@@ -8010,3 +8010,358 @@ mod prevalidate_attestation_commit_extension {
...
+                active.push(attestor_id);
+            }
+        });
+    }
+
+    #[test]
+    fn second_attestor_cannot_reuse_bls_key_on_same_chain() {
+        ExtBuilder.build_and_execute(|| {
+            let att1 = Attestor::new(STASH_1, ATTESTOR_1);
+            let att2 = Attestor::new(STASH_2, ATTESTOR_2);
```
atodorov, line 8068: Q: how come 2 attestors where both stash & attestor ID is different can have the same bls key ? 

Is this an error in test setup or am I not understanding something ?

DylanVerstraete: Where do you see this? as far as I can see, the attest call which registers the BLS key is properly asserted on duplication insertion. L8088

pallets/attestation/src/tests.rs
DylanVerstraete: can move the comment below line 8102

atodorov: Follow up question: 

will 

`                BlsKeyOwner::<Test>::get(SUPPORTED_CHAIN_KEY, att2.public_key),` also be None in this case ?

pallets/attestation/src/tests.rs
```diff
@@ -8010,3 +8010,358 @@ mod prevalidate_attestation_commit_extension {
...
+
+            assert_ok!(Attestation::validate_attestation(
+                SUPPORTED_CHAIN_KEY,
+                &signed
+            ));
+        })
+    }
+
+    #[test]
+    fn migration_v2_to_v3_keeps_first_attestor_and_clears_duplicates() {
```
atodorov: Missing a test scenario for applying migration where duplicate keys are for different chains. Should not result in any duplicate detection or changes to the active attestor set.

pallets/attestation/src/migrations.rs
```diff
@@ -294,6 +297,228 @@ impl<T: Config> OnRuntimeUpgrade for MigrateAttestorsCountV1ToV2<T> {
...
+
+        // Clear losers: drop their BLS key, force-idle, evict from `ActiveAttestors`,
+        // emit an event for ops visibility.
+        for (chain_key, loser_id, bls_public_key, winner_id) in losers.into_iter() {
+            Attestors::<T>::mutate(chain_key, &loser_id, |maybe_attestor| {
+                if let Some(attestor) = maybe_attestor {
+                    attestor.bls_public_key = None;
+                    attestor.status = AttestorStatus::Idle;
+                }
+            });
```
atodorov: IMO this shoud also emit a chill event oterwise indexer's views will become outdated.

pallets/attestation/src/migrations.rs
```diff
@@ -294,6 +297,228 @@ impl<T: Config> OnRuntimeUpgrade for MigrateAttestorsCountV1ToV2<T> {
...
+            ActiveAttestors::<T>::mutate(chain_key, |active| {
+                if let Some(pos) = active.iter().position(|a| a == &loser_id) {
+                    active.swap_remove(pos);
+                }
+            });
```
atodorov: This 2 blocks even appear to be the same as `do_chill_attestor_immediate()`. Better reuse existing code if possible so that it's always consistent.

atodorov (changes requested):

pallets/attestation/src/impls.rs
```diff
@@ -868,12 +913,50 @@ impl<T: Config> Pallet<T> {
...
+        // deduped set so the aggregate verification check itself does not accept replayed
+        // contributions.
+        let mut deduped_public_keys: Vec<PublicKey> =
+            Vec::with_capacity(attestor_public_keys.len());
+        let mut seen_keys: BTreeSet<BlsPublicKey> = BTreeSet::new();
+        for pk in attestor_public_keys.iter() {
+            let pk_bytes: BlsPublicKey = pk.as_bytes()[..].try_into().map_err(|_| {
+                log::error!("Unexpected BLS public key encoding length");
+                Error::<T>::InvalidBlsPublicKey
+            })?;
```
didac-gluwa: Why not just compare the key bytes instead of trying to cast it to a `BlsPublicKey` I know that this code should never fail in theory but still.

DylanVerstraete (approved):

didac-gluwa (approved):

PR #977, opened 2026-05-15
feat(cc-client): better reconnection logic

Trantorian1:

Refactors `cc_client::Client` and related  CC3 event streaming to simplify reconnection logic.

## Changes

- [`cc_client::Client` is now an `Arc`-backed data structure](https://github.com/gluwa/creditcoin3/pull/977/changes#diff-a66c9305ca76dd867784b9ab5b242120032f9ab2db94e9789eb03ef07d865680): clones of the client are shallow clones, which allows a single shared instance to handle all related reconnection logic.
- The `OnlineClient<SubstrateConfig>` api is now retrieved by reference instead of being cloned, which lets the client manage its lifetime across reconnections.
- `cc_client::Client::reconnect` implements a type-safe pattern which ensures reconnections can only be triggered off connection errors.
- event subscription logic in `cc_client::attestations` has been removed in favor of  `stream::cc3::StreamCC3`.

## Improvements

- `streams::cc3::StreamCC3` has been simplified and is now resistant against skipped blocks as well as connection drops via backfilling.

## Cleanup

- `cc_client::Error` dead error variants have been removed.
- `cc_client::attestations` has been moved to [`cc_client::events`](https://github.com/gluwa/creditcoin3/pull/977/changes#diff-fc4fa58b3b07590f5eca21f85e84c925e1f0fd2aae15380c6230375c3465b5d5)


common/cc-client/src/lib.rs
DylanVerstraete, line 1: Every call site now does use `arc_swap::access::Access as _;`. That's ~25 imports here. should be a single use at the top of the file, or pub use re-export so call sites just do client.api().load() without ceremony. cosmetic, but it's noisy in the diff.

common/cc-client/Cargo.toml
```diff
@@ -35,7 +35,7 @@ vrf = { workspace = true }
 serde = { workspace = true }
 sp-core = { workspace = true, default-features = false }
 thiserror = { workspace = true }
-tokio = { workspace = true }
+tokio = { workspace = true, features = ["signal"] }
```
DylanVerstraete, line 36: Where is `signal` used?

common/cc-client/src/lib.rs
```diff
@@ -1035,6 +1038,15 @@ impl Client {
     }
 }
 
+impl ClientInner {
+    async fn new(url: &str) -> Result<Self, Error> {
+        let rpc = RpcClient::from_insecure_url(&url).await?;
```
DylanVerstraete: `from_insecure_url(url)`

DylanVerstraete (changes requested):
Majority looks good, few comments

atodorov (dismissed):
Please see all the compilation issues and failed CI jobs.

common/cc-client/src/lib.rs
Trantorian1: This is actually due to a conflict between `arc_swap::access::Access` and `arc_swap::ArcSwap`. Both expose a `map` method, however for `Client::api` we need to be referring to `arc_swap::ArcSwap::map`. Unfortunately, `arc_swap::access::Access` is also implemented for `arc_swap::ArcSwap`, so if we were to import that globally the `arc_swap::access::Access::map` method would shadow the `arc_swap::ArcSwap::map` method -truly, some wonderful type design :slightly_smiling_face:

common/cc-client/Cargo.toml
Trantorian1: This was in use in `cc-client::attestations` for `tokio::signal::ctrl_c` but might no longer be in use since that file has been refactored.

Trantorian1:
> Please see all the compilation issues and failed CI jobs.

Yep, still very early draft, will be addressing CI once all changes have been fully implemented.

Trantorian1 (commented):
Finished  self-review.

common/cc-client/src/lib.rs
```diff
@@ -305,14 +308,22 @@ impl Client {
...
             )
             .await?
-            .ok_or(Error::FailedToGetBabeVrf(two_epoch_ago))?;
+            .ok_or(Error::ConnectionError(Reconnect(subxt::Error::Rpc(
+                // Babe randomness is no persisted by substrate in storage an might be missing for
+                // the first two epochs after node crash.
+                RpcError::RequestRejected(format!(
+                    "Failed to get Babe VRF at epoch {two_epoch_ago}"
+                )),
+            ))))?;
```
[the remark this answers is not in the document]
Trantorian1: This is intentionally as a way to keep trying to fetch  babe randomness until enough epochs have passed without crashing the attestor.

atodorov (commented):
Please rebase and/or restart failing CI jobs.

common/streams/cc3/src/lib.rs
```diff
@@ -1,151 +1,119 @@
...
-                                            match block.events().await {
-                                                Ok(events) => Ok(StreamEvents::new(block_number, events, chain_key)),
-                                                Err(err) => Err(Error::Subxt(err)),
-                                            }
-                                        }
-                                    })
-                                    .boxed();
-
-                                Ok::<_, Error>(stream)
+                    Err(Error::Client(err)) => yield Err(err),
```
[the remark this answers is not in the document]
Trantorian1: This is incorrect. The stream only yields errors it _cannot_ recover from.

common/cc-client/src/api.rs
```diff
@@ -36,65 +50,34 @@ impl<'a> ReconnectingRuntimeApi<'a> {
...
-            Ok::<_, crate::Error>(runtime_api)
-        }
-    };
-    tokio::select! {
-        retry = tokio_retry::Retry::spawn(strategy, reconnect) => {
-            Ok(retry.expect("Unbounded retry cannot error"))
-        }
-        _ = tokio::signal::ctrl_c() => {
-            Err(Interrupt::Stop)
+                    self.runtime_api = runtime_api;
```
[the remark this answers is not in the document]
Trantorian1: Of course it should keep spinning until the runtime api call succeeds, that is the goal off the design. However this cannot lead to stale handles as any subsequent call error results in a reconnection as well.

common/cc-client/src/lib.rs
```diff
@@ -137,36 +139,41 @@ impl Client {
...
+                Ok(inner) => {
+                    self.inner.store(Arc::new(inner));
+                    tracing::warn!("Reconnected to CC3!");
+                    break;
+                }
+                Err(err) => {
+                    tracing::warn!(?err, "Failed to reconnect to CC3");
+                }
+            }
+        }
```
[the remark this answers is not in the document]
Trantorian1: This was deemed out-of-scope as per prior reviews.

common/streams/cc3/src/lib.rs
```diff
@@ -1,151 +1,121 @@
...
-                                                Err(err) => Err(Error::Subxt(err)),
-                                            }
-                                        }
-                                    })
-                                    .boxed();
-
-                                Ok::<_, Error>(stream)
+                    Err(Error::EndOfStream) => {
+                        let blocks = config.cc3.api().load().blocks();
+                        finalized = match blocks.subscribe_finalized().await {
```
beqaabu, line 83: Is immediate resubscription okay here with a backoff or a reconnection to the client? is it possible we get stuck in a loop?

common/streams/cc3/src/lib.rs
```diff
@@ -1,151 +1,121 @@
...
-                        events = retry.await.expect("Unbounded retry cannot error");
+                            latest = block.number();
+                            for (block_number, events) in backfill.drain(..).rev() {
+                                yield Ok(StreamEvents::new(
+                                    block_number as attestor_primitives::Height,
+                                    events,
+                                    config.chain_keys.clone()
+                                ));
+                            }
+                        }
```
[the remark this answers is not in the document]
Trantorian1: This is false. `subxt`'s `api().load().blocks().at_latest()` calls [`at_or_latest`](https://github.com/paritytech/subxt/blob/9573b55e5f20114bf7b0b118c9bba85556216447/subxt/src/blocks/blocks_client.rs#L57-L61) under the hood, which evaluates to [`last_finalized_block_ref`](https://github.com/paritytech/subxt/blob/9573b55e5f20114bf7b0b118c9bba85556216447/subxt/src/blocks/blocks_client.rs#L72-L75). Therefore, `latest` is by default initialized to the latest _finalized_ block and **not** the _chain tip_.

attestor/attestor/src/worker/validation/mod.rs
```diff
@@ -1054,37 +1069,4 @@ impl WorkerAttestationValidation {
...
-
-    async fn reconnect(&mut self, err: Error) -> Result<(), Interrupt<Error>> {
-        tracing::warn!(?err, "CC3 connection lost");
-
-        let strategy = tokio_retry::strategy::ExponentialBackoff::from_millis(100)
-            .max_delay(std::time::Duration::from_millis(5_000))
-            .map(tokio_retry::strategy::jitter);
-        let reconnect = || {
-            tracing::warn!("Reconnecting to CC3...");
-
```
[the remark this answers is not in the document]
Trantorian1: This is incorrect, as the runtime call being made is re-generated after each connection, and `call` is itself stateless. The call made to `wait_for_finalized_success` will therefore always be up-to-date.

common/cc-client/src/lib.rs
```diff
@@ -44,84 +37,93 @@ use vrf::{make_proof_of_inclusion, Error as VrfError, ProofOfInclusion};
...
-/// The live subxt RPC connection lives behind an [`ArcSwap`] so that a single
-/// `Arc<Client>` shared across many tasks can be reconnected in place: once
-/// any holder calls [`Client::reconnect`], every other holder picks up the
-/// fresh subxt `RpcClient` / `OnlineClient` on its next call. The previous
-/// design swapped fields on a value-cloned `Client`, which left other
-/// `Arc<Client>` holders pinned to a stale (and often `RestartNeeded`)
-/// connection until the process was restarted.
+/// The live subxt RPC connection lives behind an [`ArcSwap`] so that a single [`Client`] shared
+/// across many tasks can be reconnected in place: once any holder calls [`Client::reconnect`],
+/// every other holder picks up the fresh subxt `RpcClient` / `OnlineClient` on its next call.
```
beqaabu, line 160: hmm im not sure this comment is true anymore. Since `fn reconnect` has `&mut self` every time a Client is cloned, every task holds an exclusive reference to self, together with their own delay. each task will call `reconnect` on their own clone of the original client, opening a new ws connection every time against a node that is already kinda down

common/cc-client/src/lib.rs
```diff
@@ -181,26 +188,12 @@ impl Client {
...
-    pub fn legacy(&self) -> LegacyRpcMethods<SubstrateConfig> {
-        self.inner.load().legacy.clone()
-    }
-
-    /// Snapshot of the live subxt `RpcClient`. Used internally for raw RPC
-    /// requests such as attestation submission.
-    fn rpc(&self) -> RpcClient {
-        self.inner.load().rpc.clone()
+    pub fn api(&self) -> impl arc_swap::access::Access<OnlineClient<SubstrateConfig>> + '_ {
+        self.inner.map(|inner: &ClientInner| &inner.api)
```
beqaabu, line 282: what happens to the api if a reconnection happens while it is still being used? the `ArcSwap` will update `inner` but will the same be true for the api?

beqaabu (commented):
looks good overall, just a couple of questions

common/streams/cc3/src/lib.rs
Trantorian1: This is actually okay since, if we do get a reconnection error, this will trigger the reconnection branch of the match statement, which will use the exponential backoff from the cc client.

common/cc-client/src/lib.rs
Trantorian1: Yes and no: while there can be contentious connection acquisition during downtime, there ends up only being a single connection alive as all other ones will be dropped when swapping out the inner arc.

common/cc-client/src/lib.rs
Trantorian1: That's exactly why this returns a reference and not a clone. Values held by `ArcSwap` are still ref-counted, so the old api can still be used even as it is invalid: it will just be returning errors which should be handled with `Client::reconnect`

beqaabu (dismissed):

common/cc-client/src/lib.rs
```diff
@@ -44,84 +37,93 @@ use vrf::{make_proof_of_inclusion, Error as VrfError, ProofOfInclusion};
...
-/// atomically replaces the live connection for *every* `Arc<Client>`
-/// holder at once.
-struct ClientInner {
-    rpc: RpcClient,
-    api: OnlineClient<SubstrateConfig>,
-    legacy: LegacyRpcMethods<SubstrateConfig>,
+#[derive(Debug)]
+/// Type-safe reconnection wrapper which prevents us from shooting ourselves in the foot by
+/// reconnecting on a non-recoverable error.
+pub struct Reconnect(subxt::Error);
```
didac-gluwa, line 82: Good idea on this pattern! The naming could be better I think, though I'm having a hard time finding a better one, maybe `ReconnectToken`? Or `RecoverableRpcError`?

common/cc-client/src/lib.rs
```diff
@@ -137,36 +139,41 @@ impl Client {
...
 
-    /// Open a fresh subxt connection (RPC + `OnlineClient` + `LegacyRpcMethods`)
-    /// against `url`. Used by both [`Client::new`] and [`Client::reconnect`].
-    async fn build_inner(url: &str) -> Result<ClientInner, Error> {
-        let rpc = RpcClient::from_insecure_url(url.to_owned()).await?;
-        let api = OnlineClient::<SubstrateConfig>::from_rpc_client(rpc.clone()).await?;
-        let legacy = LegacyRpcMethods::<SubstrateConfig>::new(rpc.clone());
-        Ok(ClientInner { rpc, api, legacy })
+    /// Atomically replace the live subxt connection with a freshly-opened one.
+    pub async fn reconnect(&mut self, Reconnect(err): Reconnect) {
```
didac-gluwa: For this I know right now we don't need it, but have you considered splitting the method in two?
* One would be `force_reconnect` without any parameter
* The other would be `reconnect_from` which would call `force_reconnect`

Trantorian1: Having a  `force_reconnect` kind of defeats the whole purpose of type-checking the `Reconnect` type though. Personally, I feel that would just be more error-prone.

didac-gluwa (dismissed):

common/cc-client/src/lib.rs
DylanVerstraete: This one will set in a permanent retry loop on shutdown until the process is killed.

DylanVerstraete (changes requested):
`reset_connection_delay()` placement looks backwards. It's called in the `FailedToCreateProofOfInclusion` ("not selected") early-return arm, but I don't see it on either the successful babe-randomness fetch or the successful submission path. Shouldn't the delay reset specifically when we prove the connection is healthy by completing a real operation? Right now, after several consecutive `MajorityNotReached`/duplicate-race outcomes, the delay never resets, and the next genuine reconnect starts from whatever level it last grew to

DylanVerstraete (commented):
Quick observation on the cc3 stream backfill (`common/streams/cc3/src/lib.rs`): the parent-walk loop fires sequential `blocks.at(parent_hash)` requests as fast as the source chain RPC can answer them. The walk is inherently serial (each step needs the prior block's parent_hash), but there's no minimum interval between fetches. So, a long gap (post-outage recovery, large in-stream gap) bursts the RPC at whatever rate the network allows.                                                                                                                                                        
 
The archiver handles this differently (`max_fetch_tasks` concurrency limit at `archiver/src/config.rs:34`), but its workload is concurrent, so the knob is concurrency, not RPS. For a serial parent, the equivalent knob is "minimum interval between fetches."

DylanVerstraete (approved):

beqaabu (approved):

didac-gluwa (approved):

PR #978, opened 2026-05-15
fix(proof-gen-api-server): gate startup checkpoint cache during revert pruning

creditcoinprotoclaw:
## Summary

Companion to #975. That PR added a bucket-granular gate on `pallet_attestation::Pallet::checkpoint_if_stable` so the `block-prover` precompile cannot trust checkpoint digests in pivots still pending async prune after a `revert_to`. The off-chain prover API has the symmetric problem on its startup path: it warms its in-memory checkpoint cache by iterating raw `Checkpoints` storage, which during an in-flight reversion still contains the stale post-revert digests.

Those digests would then sit in the cache and be used as continuity-proof trust anchors until the next `BlockAttested` / `CheckpointReached` / `RevertedAttestationChainTo` event rewrote it. Within that window the prover API would happily serve continuity proofs anchored at exactly the checkpoints the operator's `revert_to` was meant to invalidate — until the on-chain verifier rejects them at submission time with `Continuity proof does not match attestation or checkpoint`.

## Bug

`proof-gen-api-server/src/services/continuity_service/mod.rs` populates `checkpoint_cache` from `cc_provider.get_checkpoints_for_chain(chain_key)`. That helper in `cc-client` iterates raw `pallet_attestation::Checkpoints` storage via subxt, which during the async pruning window after `do_revert_to` still holds stale entries in pivots above `checkpoint_height` (see #975 for the full description of the on-chain window).

The runtime event path is already correct: `RevertedAttestationChainTo` triggers `revert_caches`, which `split_off`s entries above the revert height. The bug is purely on the startup load.

## Fix

Mirror the on-chain gate on the startup load.

* **`attestor-primitives`**: hoist `CHECKPOINT_BUCKET_SIZE` (was a private const inside `pallet-attestation`) plus a small `checkpoint_pivot` helper, so off-chain consumers can compute the same pivot as the pallet without depending on the pallet crate. `pallet-attestation` re-exports the constant so existing call sites keep working.

* **`cc-client`**: add `get_stable_checkpoints_for_chain(chain_key)`:

  1. Pins a single block ref via `api.blocks().at_latest().await?` so `CheckpointPruningStates` and the `Checkpoints` iteration read against the same chain state. Without the pin a revert event landing between the two reads could leave us with a pruning state from block N and checkpoints from block N+k.
  2. Fetches the chain's `CheckpointPruningState` at that hash.
  3. Iterates `Checkpoints` at the same hash and drops any entry whose pivot is `>= state.next_pivot`.

  Heights in the sync-cleaned revert bucket and in already-drained buckets stay readable. When no reversion is in flight the set matches `get_checkpoints_for_chain` exactly. Same semantics as `Pallet::checkpoint_if_stable` from #975.

* **`continuity::rpc::CcRpcProvider`**: add the trait method with rustdoc pointing back to the on-chain gate. `CcClient` impl delegates to the new `cc-client` helper. Mock provider delegates to the unfiltered set by default for tests that do not model a reversion.

* **`proof-gen-api-server`**: startup cache population now calls `get_stable_checkpoints_for_chain` instead of `get_checkpoints_for_chain`. Runtime event handling (`RevertedAttestationChainTo` -> `revert_caches`) was already correct and is unchanged.

## Why startup, not the hot path

Grepping `proof-gen-api-server` for chain RPC reads against `Checkpoints` storage: the only path that touches raw storage is the startup load. All other reads go through the in-memory `checkpoint_cache`, which is updated reactively by `BlockAttested` / `CheckpointReached` / `RevertedAttestationChainTo` events. Those event handlers already do the right thing.

## Tests

* `proof-gen-api-server/src/services/continuity_service/helpers.rs::tests::startup_uses_stable_checkpoints_and_drops_stale_pivots`
  * Injects a `CcRpcProvider` that returns different sets from `get_checkpoints_for_chain` (unfiltered: includes stale digests above the revert height) and `get_stable_checkpoints_for_chain` (filtered: only stable-pivot entries).
  * Boots `ContinuityService` with the custom provider.
  * Asserts the startup checkpoint cache contains *exactly* the stable set and that each modeled stale-pivot height is absent. This pins the startup code path to the gated read; if anyone routes startup back through `get_checkpoints_for_chain` the test fails.

Local test results:

* `proof-gen-api-server` lib tests: 39 passed (+1 new).
* `continuity` lib tests: 4 passed.
* `pallet-attestation` lib tests: 233 passed (unchanged behavior; only the const home moved).
* `cargo fmt --all` clean.
* `cargo clippy` clean on every touched crate (`attestor-primitives`, `pallet-attestation`, `cc-client`, `continuity`, `proof-gen-api-server`).

## Diff overview

| File | Change |
| --- | --- |
| `primitives/attestor/src/lib.rs` | Add `CHECKPOINT_BUCKET_SIZE` const and `checkpoint_pivot` helper. |
| `pallets/attestation/src/lib.rs` | Re-export `CHECKPOINT_BUCKET_SIZE` from `attestor-primitives`. |
| `common/cc-client/src/lib.rs` | New `get_stable_checkpoints_for_chain` with pinned-block read + pruning-state gate. |
| `common/continuity/src/rpc.rs` | New `CcRpcProvider::get_stable_checkpoints_for_chain` trait method + `CcClient` impl. |
| `common/continuity/src/mocks.rs` | Mock impl: delegate to unfiltered set (default behavior for tests not modeling a reversion). |
| `proof-gen-api-server/src/services/continuity_service/mod.rs` | Startup cache population uses the gated read. |
| `proof-gen-api-server/src/services/continuity_service/helpers.rs` | New regression test + helper `CcRpcProvider` modeling an in-flight reversion. |

### DIFF
--- a/common/cc-client/src/lib.rs
+++ b/common/cc-client/src/lib.rs
@@ -27,9 +27,9 @@ use cc3::runtime_types::{
 };
 
 use attestor_primitives::{
-    block::ContinuityProof, Attestation as RpcAttestation, AttestationCheckpoint, AttestationData,
-    AttestorId, AttestorStatus, BlsPublicKey, BlsSignature, ChainEncodingVersion, ChainKey, Digest,
-    SignedAttestation,
+    block::ContinuityProof, checkpoint_pivot, Attestation as RpcAttestation, AttestationCheckpoint,
+    AttestationData, AttestorId, AttestorStatus, BlsPublicKey, BlsSignature, ChainEncodingVersion,
+    ChainKey, Digest, SignedAttestation,
 };
 use supported_chains_primitives::SupportedChain;
 use vrf::{make_proof_of_inclusion, Error as VrfError, ProofOfInclusion};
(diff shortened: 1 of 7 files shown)


didac-gluwa (approved):

beqaabu (approved):

DylanVerstraete:
Closing because irrelevant at this point when we have this in place https://github.com/gluwa/creditcoin3/blob/ddcac8b5e09109956e73aa4b5991efbd75f3027a/proof-gen-api-server/src/events/mod.rs#L317

PR #986, opened 2026-05-20
Added metadata auto update worker to cc-client

didac-gluwa:
Adds a `ClientRuntimeUpdater` sidecar task to the CC Client which constatly monitors the rpc and updates the metadata.

Also added a log line for the production worker in the attestor which reports the current `spec_version` and `transaction_version` of the cc client it is subscribed to.

The following tests were perfomed with a local node + anvil + zombienet:

* Upgrading just the runtime version -> **Works, attestors continue attesting without issue**
* Changing order of events in pallet -> **Works, attestors continue attesting without issue**
* Adding new extra value at the end of the `BlockAttested` event -> **Works**
   * The subxt decodes the `BlockAttested` event correctly
   * The new value is of course not available to the attestors since the struct definition they have doesn't contain it
* Changing order of values in BlockAttested event -> **Fails (expected)**
   * Attestors miss the event entirely (likely because of signature mismatch), **they don't error out**
   * After correcting the signature and making another update the attestors **needed to be restarted to continue attesting normally**
   * That is related to the `BlockAttested` events that they all missed meaning that when they produced the next attestations, those were not correctly linked to the previous one
   * This is just an extreme example that is very unlikely to happen, since the attestor events and related extrinsics are treated with extra care when we modify them
   
In conclusion, **any updated that don't affect the surface API of the attestors will not affect them**, meaning attestor binaries will only need upgrading on changes to that surface API, for all other runtime upgrades they will work without issue


attestor/attestor/src/worker/production/mod.rs
```diff
@@ -178,6 +178,13 @@ impl super::Worker for WorkerAttestationProduction {
         loop {
             let can_attest = self.can_attest.load(std::sync::atomic::Ordering::Acquire);
 
+            let runtime_version = self.cc3.api().runtime_version();
+            let spec_version = runtime_version.spec_version;
+            let transaction_version = runtime_version.transaction_version;
+
+            // Added for ease of debugging and visibility into the production loop, especially around runtime upgrades.
+            tracing::info!(spec_version = %spec_version, transaction_version = %transaction_version, can_attest, "Handling next production event");
```
BradleyOlson64, line 186: Seems like this will print very frequently. Maybe that's not such a bad thing though 🤷‍♂️

common/cc-client/src/lib.rs
```diff
@@ -88,8 +88,19 @@ struct ClientInner {
...
 
+// Manual Drop implementation to ensure that the updater task is aborted when the client is dropped.
+impl Drop for ClientInner {
+    fn drop(&mut self) {
+        self.updater_handle.abort();
+    }
+}
+
+// Timeout for the runtime metadata updater task to wait before retrying after an error or stream end.
+const UPDATER_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
```
BradleyOlson64, line 102: Perhaps rename to `UPDATER_FREQUENCY`?

BradleyOlson64 (dismissed):

attestor/attestor/src/worker/production/mod.rs
didac-gluwa: Yeah, I though about it but I think the extra logs are worth it

beqaabu, line 186: this will log on for every finalized block and attestation, it will be too noisy. can we either make it a debug or make it less frequent?

common/cc-client/src/lib.rs
```diff
@@ -153,7 +164,34 @@ impl Client {
...
-        Ok(ClientInner { rpc, api, legacy })
+
+        // In order to keep metadata up to date, we need to continuously poll the node for updates.
+        // This is done by spawning a background task that calls `perform_runtime_updates` in a loop.
+        // If the stream ends or an error occurs, we log the error and retry after a short delay.
+        let updater = api.updater();
+        let updater_handle = tokio::spawn(async move {
+            loop {
+                match updater.perform_runtime_updates().await {
+                    Ok(()) => info!(
```
beqaabu, line 175: I think this one should also be a warning, because `Ok(())` returned from `perform_runtime_updates` means it the runtime version sub termiaated with no more items, which usually doesnt happen, so it deserves a warn

common/cc-client/src/lib.rs
```diff
@@ -153,7 +164,34 @@ impl Client {
...
+                        "runtime metadata updater stream ended, retrying in {}s",
+                        UPDATER_TIMEOUT.as_secs()
+                    ),
+                    Err(e) => warn!(
+                        "runtime metadata updater stopped: {e}, retrying in {}s",
+                        UPDATER_TIMEOUT.as_secs()
+                    ),
+                }
+
+                tokio::time::sleep(UPDATER_TIMEOUT).await;
```
atodorov, line 185: I am wondering if 5 seconds isn't too small of a timeout. Is this going to be called by every single attestor process ? If so we may end-up bombarding our RPCs with too many requests.

common/cc-client/src/lib.rs
```diff
@@ -153,7 +164,34 @@ impl Client {
...
         let legacy = LegacyRpcMethods::<SubstrateConfig>::new(rpc.clone());
-        Ok(ClientInner { rpc, api, legacy })
+
+        // In order to keep metadata up to date, we need to continuously poll the node for updates.
+        // This is done by spawning a background task that calls `perform_runtime_updates` in a loop.
+        // If the stream ends or an error occurs, we log the error and retry after a short delay.
+        let updater = api.updater();
+        let updater_handle = tokio::spawn(async move {
+            loop {
+                match updater.perform_runtime_updates().await {
```
atodorov, line 174: Q1) Whare is this updater coming from? I am actually not familiar with what it does. 

2) This flow needs to be exercised as part of CI jobs to make sure it doesn't crash. Let me know if you need pointers in the right direction. 

3) Once metadata is updated, don't we need to restart all processes that are using said metadata ? 

3.1) What if there's some sort of backwards incompatibility, i.e. you need to recompile the binary - how do we prevent that in the first place, how do we detect it and how do we stop processes from automatically updating themselves and crashing unrecoverably. (Correct me if I am not understanding something here). Looks like a pretty significant risk to me.

atodorov (changes requested):

atodorov:
> The following tests were perfomed with a local node + anvil + zombienet:

It's a good idea to codify all of these as new/part of existing CI jobs for starters.

atodorov:
> 
>     * Changing order of events in pallet -> **Works, attestors continue attesting without issue**
> 

I think we have a check to fail if that ever happens b/c potentially other downstream clients (of polkadot-js api) may break.  Thanks for testing it but IMO that should be relegated to a lower priority item for testing.

>     * Adding new extra value at the end of the `BlockAttested` event -> **Works**
>       

What if you add a new field in the middle of this struct ? 

>     * Changing order of values in BlockAttested event -> **Fails (expected)**
>       

hmmm, sounds like something we should start checking for/safeguarding during the development flow (for all events) so as to not introduce unnecessary risks of breakages.  Do you mind explaining this in a bit more details?


>       * Attestors miss the event entirely (likely because of signature mismatch), **they don't error out**
>       * After correcting the signature and making another update the attestors **needed to be restarted to continue attesting normally**
>       * That is related to the `BlockAttested` events that they all missed meaning that when they produced the next attestations, those were not correctly linked to the previous one
>       * This is just an extreme example that is very unlikely to happen, since the attestor events and related extrinsics are treated with extra care when we modify them

^^^ There's not much that stops us from breaking these though. I'd prefer to not rely on "extra care" but rather have tools in place. Are we concerned only about BlockAttested events or any other as well ? Which extrinsics (or what mode of breaking them) should we be looking for ? 


> 
> 
> In conclusion, **any updated that don't affect the surface API of the attestors will not affect them**, meaning attestor binaries will only need upgrading on changes to that surface API, for all other runtime upgrades they will work without issue

Is there any way to detect this programmatically ? Sounds like it would be of great help if we could do it.

common/cc-client/src/lib.rs
didac-gluwa: 1) The updater comes from subxt itself it just opens a subscrition to the rpc for metadata updates
2) Not sure it's worth it, you would need to compile two separate runtimes and then do the sudo call
3) Depends on the API surface those processes are touching, see PR description
3.1) In that case whatever process is likely to fail/crash but that would already have happened after we did a runtime upgrade, this PR is only to avoid updating some components when we do runtime upgrades that don't affect the API surface

common/cc-client/src/lib.rs
didac-gluwa: Hmm in theory this worker receives push events from the RPC so it should not bombard it, will increase the retry to 30s anyways

didac-gluwa:
Keeping as draft for now, since it won't be in scope soon

common/cc-client/src/lib.rs
atodorov: > 2. Not sure it's worth it, you would need to compile two separate runtimes and then do the sudo call

I don't support the conviction that it's not worth testing but I agree there's disproportionally high cost in testing this specific feature compared to others.

PR #1005, opened 2026-05-29
Internal Audit - Phase 2 - Point 2

didac-gluwa:
Addresses issue from internal audit phase 2 - point 2

2.  `RandomnessByEpochIndex` storage is unbounded

**NOTE**: I haven't added a storage migration since I don't believe its necessary, after this PR is merged both current devent and testnet chains storage will stop growing, but it will remain at a bigger size than what it will be on mainnet, which again I don't believe warrants a storage migration to clean it up on those chains.


atodorov:
> **NOTE**: I haven't added a storage migration since I don't believe its necessary, after this PR is merged both current devent and testnet chains storage will stop growing, but it will remain at a bigger size than what it will be on mainnet, which again I don't believe warrants a storage migration to clean it up on those chains.

Both Testnet & Devnet are used for testing purposes for things like forks & try-runtime which fetch all storage items from the existing chain. IMO if we can clean up what isn't strictly necessary will still be a benefit.

pallets/randomness/src/lib.rs
```diff
@@ -72,6 +89,26 @@ pub mod pallet {
 
                 RandomnessByEpochIndex::<T>::insert(epoch_index, randomness);
 
+                let max_epoch_history = T::MaxEpochHistory::get();
+                let prune_to = epoch_index.checked_sub(max_epoch_history);
+                let prune_from = last_seen_epoch_index
```
BradleyOlson64, line 96: Say we have a node shut down for quite a while. And we have epoch randomness from a long time ago in storage. Does this `prune_from` actually capture those old entries?

didac-gluwa: It should since `last_seen_epoch` will contain the latest epoch before the node went down

BradleyOlson64 (dismissed):

DylanVerstraete (commented):
I'd actually prefer a storage migration, since the pruning state is going to be dead code anyway..

didac-gluwa:
> I'd actually prefer a storage migration, since the pruning state is going to be dead code anyway..

Added migration!

atodorov (dismissed):

didac-gluwa:
Superseded by https://github.com/gluwa/creditcoin3/pull/1033

PR #1013, opened 2026-06-01
fix(chain-info): avoid panic on CheckpointBuckets/Checkpoints desync (Certik N8)

DylanVerstraete:
The bucket-search precompiles unwrapped Checkpoints::get assuming every CheckpointBuckets entry has a matching checkpoint. The prefix-clear paths in clear_or_revert.rs advance the two maps' cursors independently after a chain removal, so a bucket entry can briefly outlive its checkpoint. An unwrap there would trap runtime WASM and halt block import (DoS). Skip orphan entries and continue the pivot search instead. Adds regression tests covering both find_highest_attested_before and find_lowest_attested_after.

### DIFF
--- a/precompiles/chain-info/src/lib.rs
+++ b/precompiles/chain-info/src/lib.rs
@@ -231,28 +231,31 @@ where
                 let mut items_processed = 0_u64;
 
                 // We search the checkpoint bucket for the highest checkpoint below the target height.
-                maybe_highest =
+                let max_below =
                     CheckpointBuckets::<Runtime>::iter_key_prefix((chain_key, block_pivot))
                         .inspect(|_| {
                             items_processed += 1;
                         })
                         .filter(|block_number| *block_number < target_height)
-                        .max_by_key(|block_number| *block_number)
-                        .map(|block_number| {
-                            handle.record_cost(GAS_STORAGE_LOOKUP)?;
+                        .max_by_key(|block_number| *block_number);
 
-                            // At this point, we know the checkpoint exists. So we can safely unwrap.
-                            let digest =
-                                Checkpoints::<Runtime>::get(chain_key, block_number).unwrap();
-
-                            <EvmResult<HeightHashResult>>::Ok(HeightHashResult {
-                                height: block_number,
-                                hash: digest,
-                                is_attestation: false,
-                                exists: true,
-                            })
-                        })
-                        .transpose()?;
+                // `CheckpointBuckets` is expected to mirror `Checkpoints` (writes are paired).
+                // Treat any desync as a soft miss and fall through to the next pivot rather
+                // than panic — the prefix-clear paths in `clear_or_revert.rs` advance the two
+                // maps' cursors independently after chain removal.
+                maybe_highest = if let Some(block_number) = max_below {
+                    handle.record_cost(GAS_STORAGE_LOOKUP)?;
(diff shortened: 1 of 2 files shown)


creditcoinprotoclaw (commented):
Reviewed by Claw 🦞 — `LGTM` with one follow-up flag.

### Summary
Kills two `Checkpoints::get(...).unwrap()`s in `find_highest_attested_before` / `find_lowest_attested_after` that would panic the runtime if a `CheckpointBuckets` entry briefly outlives its matching `Checkpoints` row. Verified the desync premise: `on_init_clear_checkpoints` and `on_init_clear_buckets` in `pallets/attestation/src/clear_or_revert.rs` advance the two maps with independent `clear_prefix` cursors, so an interleaved block can absolutely observe orphan buckets. Skipping rather than panicking is the right call for a view precompile — a panic here traps WASM and halts block import, which is a DoS, exactly as the description says.

### Correctness
- New control flow: take the max/min `block_number` from the bucket prefix iter, then do **one** `Checkpoints::get`. If `None`, fall through to the next pivot via the existing outer loop. Semantics preserved for the happy path; failure mode is now "soft miss" instead of "trap". 👍
- Gas accounting still records `GAS_STORAGE_LOOKUP` for the `Checkpoints` read on the orphan branch and the per-item iteration cost is unchanged. No free-lookup hole. 👍
- One nit: on a desync the search continues to the next pivot, but the orphan bucket entry itself is still the max-below / min-above within its own bucket. That means we won't find a *lower* (or higher) consistent checkpoint **within the same pivot** that the orphan shadowed. In practice the desync window is tiny (one block of clearing progress) and the next pivot will almost certainly contain the right answer, so this is fine — just worth being explicit about in the comment.

### Tests
- Both regression tests construct the exact orphan-bucket condition via `CheckpointBuckets::insert` without a matching `Checkpoints::insert`, then assert the search falls through to the next pivot. Clean and targeted.
- Pre-PR: the same test bodies would `.unwrap()` panic. Post-PR: pass. Verified locally:
```
cargo test -p pallet-evm-precompile-chain-info -- skips_orphan_bucket_entry
running 2 tests
test tests::find_lowest_attested_after_skips_orphan_bucket_entry ... ok
test tests::find_highest_attested_before_skips_orphan_bucket_entry ... ok
```
Full crate suite: `23 passed; 0 failed`.

### Follow-up (out of scope, not a blocker)
`is_height_attested` also iterates `CheckpointBuckets` (lines ~446–456). It doesn't `.unwrap()` so there's no panic, but during the same desync window it can report `found_prev_checkpoint = true` for an orphan height — a brief false positive. Worth a separate ticket to either cross-check `Checkpoints` there too or document the eventual-consistency. Same comment for any other call site that treats a bucket hit as proof of attestation.

### CI
`check-version` and `rebase-check` are red — those are repo-meta jobs (version bump / target-branch sync), not the substantive checks. The Rust pipeline (`cargo-test`, `cargo-clippy`, `cargo-fmt`) was still in progress at review time; locally everything builds and the new tests pass. Worth a glance once the run finishes.

### Verdict
`LGTM` once `check-version` / `rebase-check` are addressed. Solid, surgical fix for a real DoS surface.

didac-gluwa (approved):

beqaabu (approved):

PR #1014, opened 2026-06-01
fix(attestation): reject continuity proof tail at block 0 (Certik N10)

DylanVerstraete:
A continuity proof with roots.len() >= header_number drives the tail block number to 0, making `tail_block_number - 1` underflow (panic under overflow-checks, wrap otherwise). Guard with checked_sub and reject such proofs as InvalidAttestationContinuityProofTail, with a regression test.

### DIFF
--- a/pallets/attestation/src/continuity.rs
+++ b/pallets/attestation/src/continuity.rs
@@ -122,8 +122,14 @@ impl<T: Config> Pallet<T> {
             return Err(Error::<T>::InvalidAttestationContinuityProofTail.into());
         };
 
-        if tail_prev_header_number != tail_block_number - 1 {
-            error!("❌ Continuity proof tail prev digest points to an attestation/checkpoint with header number {}, but expected {}", tail_prev_header_number, tail_block_number - 1);
+        // `tail_block_number` is attacker-influenced (header_number.saturating_sub(roots.len()));
+        // a degenerate proof can drive it to 0. Reject that explicitly instead of computing
+        // `tail_block_number - 1`, which would underflow (panics under overflow-checks, wraps otherwise).
+        let expected_tail_prev_header_number = tail_block_number
+            .checked_sub(1)
+            .ok_or(Error::<T>::InvalidAttestationContinuityProofTail)?;
+        if tail_prev_header_number != expected_tail_prev_header_number {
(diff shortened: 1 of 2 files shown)


beqaabu (approved):

didac-gluwa (approved):

creditcoinprotoclaw (commented):
🦞 Review

**Verdict:** ✅ ship it. Clean, minimal, well-tested fix.

### Correctness

- `continuity.rs` L125-130: correct guard. `tail_block_number` comes from `ContinuityProof::start_block_number` which uses `saturating_sub` ([`primitives/attestor/src/block.rs#L326-327`](https://github.com/gluwa/creditcoin3/blob/audit_continuity_fix/primitives/attestor/src/block.rs#L326-L327)), so the `0` case is genuinely reachable on attacker input. Mapping the failure to `InvalidAttestationContinuityProofTail` matches the surrounding tail-validation arm — consistent error UX for the caller.
- The regression test asserts `tail_block_number == 0` before constructing the attestation, so it really does exercise the new branch (not some earlier validation).

### Minor / out of scope

- 🔵 nit: the genesis-empty-proof branch above (~L74) also does `prev_header_number + 1` on an untrusted-derived value. Not exploitable in practice (no chain reaches `u64::MAX` blocks) but same class of arithmetic-on-untrusted-input. Worth a follow-up hardening sweep, not this PR.
- 🔵 nit (PR description, cosmetic): under `overflow-checks = false` (release), `0u64 - 1` wraps to `u64::MAX`, which the existing `!=` check on the next line *would* still reject — just via a confusing mismatch message. So the user-visible bug is mostly the debug-build panic; in release the chain was safe-by-accident. The fix is still right (debug-build hardening + clean code), just worth phrasing precisely.
- 🔵 nit: in the test, the implicit coupling between `progress_to_block(5)` and `header_number = 5u64` could use a one-liner comment for the next reader. Skip if you don't care.

### CI

- `integration-test-blockchain`: failing test is `substrate-transfer.test.ts` — unrelated to attestation continuity. Looks pre-existing/flaky; worth confirming against `main` but not blocking this PR.
- `rebase-check`: branch needs rebase on base. Standard housekeeping.

### Audit framing

This closes Certik N10. Fix matches the audit's recommended remediation pattern (explicit reject of degenerate tail rather than trusting arithmetic), and the regression test pins the exact zero-tail scenario. Good closure.

PR #1016, opened 2026-06-01
Add source chain accounts to balance check

didac-gluwa:
Adds a sepolia source chain account (Traffic Simulator account) to balance checks

### DIFF
--- a/scripts/usc-audit-automation/config-testnet.json
+++ b/scripts/usc-audit-automation/config-testnet.json
@@ -25,6 +25,17 @@
           "name": "NEW_LENDER_ accnt in 1Password. Used for testing"
         }
       ]
+    },
+    {
+      "name": "Ethereum Sepolia",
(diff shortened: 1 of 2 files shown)


atodorov (commented):
Can we see an example execution of this before merge ?

didac-gluwa:
> Can we see an example execution of this before merge ?

There you go!

image

atodorov (approved):

atodorov (changes requested):
Image

This error message is misleading - the chain prefix I mean. I would expect it to be `Ethereum Sepolia` like in the config file.

didac-gluwa:
> Image
> 
> This error message is misleading - the chain prefix I mean. I would expect it to be `Ethereum Sepolia` like in the config file.

Fixed!

image

BradleyOlson64 (approved):

beqaabu (approved):

atodorov (approved):

PR #1027, opened 2026-06-02
Certik N21: Ensure chain is supported before verifying

BradleyOlson64:
I refrain from updating runtime version in this pr, since that's been a source of conflicts with other folks' concurrent work in the past. Can add that if needed.

- Adds a check verifying that a source chain is supported in the block prover's `get_checkpoint`

- Removes dead code `last_checkpoint`

- Simplifies some logic in `verify_continuity_chain`

- Removes unused traits `AttestationProvider` and `CheckpointProvider`


precompiles/block-prover/src/lib.rs
didac-gluwa: This comment needs to be updated to reflect that we now charge gas thrice in the happy path

didac-gluwa (changes requested):

precompiles/block-prover/src/lib.rs
BradleyOlson64: Comment updated 👍

precompiles/block-prover/src/lib.rs
```diff
@@ -358,7 +362,29 @@ where
     /// Currently unused but kept for potential future optimizations.
     #[allow(dead_code)]
```
DylanVerstraete: if this is actually unused, remove it

precompiles/block-prover/src/continuity.rs
```diff
@@ -138,13 +141,7 @@ where
         // since they're always used together
         let they_match = attestation_matches || checkpoint_matches;
 
-        // Special case: If the continuity chain ends at query height and query height
```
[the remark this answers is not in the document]
BradleyOlson64: Not an issue. Attestations are removed in a single block.

precompiles/block-prover/src/lib.rs
BradleyOlson64: Addressed 👍

DylanVerstraete (dismissed):

precompiles/block-prover/src/lib.rs
```diff
@@ -374,16 +368,22 @@ where
...
+    /// Gas: charges `GAS_STORAGE_LOOKUP` thrice. Once to check that the chain is
+    /// supported, once for the `CheckpointPruningStates` guard read, and once for
+    /// the `Checkpoints` read itself.
     ///
     /// Returns `Ok(Some(..))` / `Ok(None)` on success; gas recording failures surface as `EvmResult` errors.
     fn get_checkpoint(
         handle: &mut impl PrecompileHandle,
         chain_key: u64,
         block_number: u64,
     ) -> EvmResult<Option<H256>> {
```
atodorov, line 380: A similar function also exists in `pallets/attestation/src/lib.rs` and it doesn't check whether chain_key is supported or not, just calls `Checkpoints::<T>::get(chain_key, block_number)` which I guess will return None too.  Not sure if we want to do anything about that or not, or whether we want to reuse the existing implementation instead of providing an alternative one.

precompiles/block-prover/src/lib.rs
```diff
@@ -374,16 +368,22 @@ where
...
+        // Charge for the supported chains lookup
+        handle.record_cost(GAS_STORAGE_LOOKUP.saturating_mul(1))?;
+        if SupportedChains::<Runtime>::get(chain_key).is_none() {
+            return Ok(None);
+        }
```
atodorov, line 385: This is missing test coverage. Can be a unit test.

precompiles/block-prover/src/lib.rs
```diff
@@ -374,16 +368,22 @@ where
...
     ///
-    /// Gas: charges `GAS_STORAGE_LOOKUP` twice, once for the `CheckpointPruningStates`
-    /// guard read and once for the `Checkpoints` read itself. Both reads happen
-    /// unconditionally so the pessimistic cost matches the worst case.
+    /// Gas: charges `GAS_STORAGE_LOOKUP` thrice. Once to check that the chain is
+    /// supported, once for the `CheckpointPruningStates` guard read, and once for
+    /// the `Checkpoints` read itself.
     ///
     /// Returns `Ok(Some(..))` / `Ok(None)` on success; gas recording failures surface as `EvmResult` errors.
     fn get_checkpoint(
```
atodorov, line 376: Q: what/who calls this function ? Looks a bit like dead code to me. If that's the case why not remove it altogether ?

atodorov (changes requested):

precompiles/block-prover/src/lib.rs
BradleyOlson64: This gets called as `Self::get_checkpoint` in `precompiles/block-prover/src/continuity.rs`

precompiles/block-prover/src/lib.rs
BradleyOlson64: Good catch Alex! 

It looks like the traits `AttestationProvider` and `CheckpointProvider` were both no longer in use anywhere.

That includes the alternative definition of `get_checkpoint`. Both now removed in this pr.

precompiles/block-prover/src/lib.rs
BradleyOlson64: Unit test created 👍

DylanVerstraete (approved):

atodorov (approved):

didac-gluwa (approved):

beqaabu (approved):

PR #1028, opened 2026-06-03
fix(attestation): reject cross-stash clearing of retired BLS protection (Certik N11)

DylanVerstraete:
register_attestor accepts an arbitrary attestor_id, so any caller could register a controller retired by a different stash and prematurely clear that stash's retired BLS-key protection (releasing the BlsKeyOwner claim before the unbond delay elapses). Guard the clear on stash ownership and reject mismatches with ControllerRetiredByAnotherStash; the self-service re-registration path is unchanged. Adds a regression test.


didac-gluwa (dismissed):

pallets/attestation/src/tests.rs
```diff
@@ -4605,6 +4605,81 @@ fn purge_retired_bls_does_not_remove_other_stash_bls_row_on_index_mismatch() {
...
+        assert!(crate::RetiredAttestorBlsKeys::<Test>::contains_key(
+            SUPPORTED_CHAIN_KEY,
+            ATTESTOR_1
+        ));
+        assert_eq!(
+            BlsKeyOwner::<Test>::get(SUPPORTED_CHAIN_KEY, a1.public_key),
+            Some(ATTESTOR_1)
+        );
+
+        // The owning stash may still re-register its own controller, clearing its own protection.
```
atodorov, line 4666: Missing a test scenario where STASH_2 attempts registration after the unbonding period has elapsed. 

This comment
>     /// (the `BlsKeyOwner` protection that must persist until that stash's unbond delay elapses).

makes me think that once the unbonding period is over another stash should be able to register the same attestor ID, no ? 

Regardless of which case it is, we need to have 2 test scenarios covering the state before/after unboding period has elapsed.

atodorov (changes requested):
Overall LGTM. Needs one more test scenario.

beqaabu (dismissed):

atodorov (approved):

beqaabu (approved):

PR #1037, opened 2026-06-04
cli: use latest checkpoint instead of latest attestation in prover-check

creditcoinprotoclaw:
Per Alex's request: in `cli/src/scripts/prover-check.ts` replace the use of the latest attestation with the latest checkpoint when seeding `LAST_SOURCE_BLOCK`.

### What changed

- Replaced `chainInfoPrecompile.getLatestAttestedHeightAndHash(chainKey)` with a direct call to the chain-info precompile method `get_latest_checkpoint_height_and_hash`.
- `LAST_SOURCE_BLOCK` and the checkpoint-pair iteration upper bound now derive from the latest checkpoint height instead of the latest attestation height.
- Updated the log line to reflect the new source.

### Why

Attestations lag behind checkpoints, so seeding with the latest checkpoint exercises a more recent slice of the chain by default.

### Implementation note

`@gluwa/usc-sdk` (0.12.2) does not yet expose a wrapper for `get_latest_checkpoint_height_and_hash` on `PrecompileChainInfoProvider`, so the script instantiates an `ethers.Contract` directly using:
- `chainInfo.CHAIN_INFO_PRECOMPILE_ADDRESS` (already exported)
- the `chain_info.json` ABI shipped with the SDK package (same import style already used for `evmV1DecoderAbi.json`).

If/when the SDK gains a `getLatestCheckpointHeightAndHash` helper, this script can be migrated to use it.

### Verification

- `npx prettier --check cli/src/scripts/prover-check.ts` ✅
- `npx eslint cli/src/scripts/prover-check.ts` ✅ (no new warnings)
- `npx tsc --noEmit -p cli` produces only the two pre-existing errors (`utils` export + `evmV1DecoderAbi.json` resolution) that were already present on `usc-dev` before this change; no new diagnostics were introduced.

### DIFF
--- a/cli/src/scripts/prover-check.ts
+++ b/cli/src/scripts/prover-check.ts
@@ -1,6 +1,7 @@
 import { mkdirSync, writeFileSync } from 'fs';
 import { blockProver, chainInfo, proofGenerator, utils } from '@gluwa/usc-sdk';
 import EvmV1DecoderABI from '@gluwa/usc-sdk/dist/utils/evmV1DecoderAbi.json';
+import ChainInfoABI from '@gluwa/usc-sdk/dist/chain-info/chain_info.json';
 import { Contract, WebSocketProvider } from 'ethers';
 import { createClient } from 'graphqurl';
 import axios from 'axios';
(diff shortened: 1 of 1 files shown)


atodorov (approved):
LGTM https://github.com/gluwa/creditcoin3/actions/runs/26961592432

didac-gluwa (approved):

PR #1047, opened 2026-06-08
fix(chain-info-precompile): route raw checkpoint reads through checkpoint_if_stable

DylanVerstraete:
Use `checkpoint_if_stable` in chain info precompile

### DIFF
--- a/precompiles/chain-info/src/lib.rs
+++ b/precompiles/chain-info/src/lib.rs
@@ -11,8 +11,8 @@ use sp_std::vec::Vec;
 
 use attestor_primitives::{ChainId, ChainKey};
 use pallet_attestation::{
-    AttestationChainGenesisBlockNumber, Attestations, CheckpointBuckets, Checkpoints,
-    LastCheckpoint, LastDigest, Pallet as PalletAttestationPoc, CHECKPOINT_BUCKET_SIZE,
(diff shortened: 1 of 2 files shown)


atodorov (approved):

precompiles/chain-info/src/lib.rs
```diff
@@ -252,8 +252,14 @@ where
 
                 maybe_highest = None;
                 for block_number in candidates {
```
atodorov, line 254: Note: from what I can see none of these precompile functions check whether chain_key is supported or not, see PR #1027 for example. Not sure if we want to be adding such checks or not (should be a follow up PR though).

beqaabu (approved):

didac-gluwa (changes requested):
Indexer test needs to adjusted

atodorov:
> Indexer test needs to adjusted

Nope. That's a flaky behavior - picking up values from the previous state - see attempt #2 - all good.

didac-gluwa (approved):

precompiles/chain-info/src/lib.rs
DylanVerstraete: Yes we should probably add this, will create a followup PR

PR #1083, opened 2026-06-23
feat: writeability

DylanVerstraete:
## USC Write-Ability (Creditcoin L1 → destination chain)

Adds the **write-ability** layer to USC: a dApp on Creditcoin L1 can publish a message that is
attested by the validator/attestor set, relayed to a destination chain, and (optionally)
acknowledged back on Creditcoin via a trust-minimized native proof.

> Base branch is `attestor_v2` (the failing `target-branch-check` reflects that and resolves on
> retarget). Running live on **cc3-usc-dev**: 12 attestors across 3 clusters attest Sepolia
> (chain_key 7), with delivery, acks, and bridge claims flowing end to end.

### End-to-end flow

```
dApp ──publishMessage──▶ Outbox (Creditcoin L1)
                          │  emits MessagePublished
        attestors ────────┘  each ECDSA-signs messageHash, gossips a vote
                              on {chain_key}/message-votes/v1
        message-relayer ──▶  aggregates 2N/3+1 unique signers
                          │
                          ▼
   Inbox.deliverMessage (destination) ──▶ EOAValidator.validateVotes ──▶ dApp.receiveMessage
                          │  emits MessageDelivered
   (optional ack) relayer watches MessageDelivered ──▶ fetches native USC delivery proof
                          ▼
   AcknowledgmentValidator.submitAcknowledgment (Creditcoin) ──▶ block-prover precompile
   (merkle inclusion + continuity) + EvmV1Decoder ──▶ Outbox.acknowledgeMessage
```

`messageHash = keccak256(abi.encode(messageId, emitter, destinationChainKey, creditcoinChainId, payload))`
(raw 32-byte hash, no EIP-191), shared byte-for-byte across attestor, relayer, and the Solidity Inbox.

### What's in the PR

**Attestor (`attestor/attestor/src/tasks/write_ability`)**
- On-chain Outbox resolution (chain-info precompile → `IOutboxFactory.getOutbox`) with dynamic
  activation when the factory/outbox is registered later.
- EVM `MessagePublished` listener + finality gate, ECDSA signing, message-vote gossip on a new
  topic over the **existing** attestor swarm (with the same bounded publish-retry queue as
  block-attestation votes), in-memory aggregator, anti-abuse (allowlist + ecrecover + dedup +
  caps), reobservation liveness recovery, metrics, and on-chain attestor-set hot-reload.
- **On-chain governance gating**: the local `enabled` flag is the operator opt-in; a registered
  `WriteAbilityConfigs` entry is authoritative for `message_attestation_enabled` and the `bytes32`
  chain key. New CLI: `--writeability` / `--cc3-eth-url`.

**Shared crate (`common/write-ability`)**
- Canonical `hash` / `envelope` / `abi` / `protocol` re-exported by both attestor and relayer so
  they cannot diverge on `messageHash` or the vote wire format (golden-vector tested).

**Runtime / pallets / precompiles**
- `pallet-supported-chains`: `WriteAbilityConfig` + `OutboxFactories` storage (zero-value inputs
  rejected), extrinsics + events, storage migration, runtime API; chain-info precompile
  `get_outbox_factory_address(uint64)`.

**Demo / deploy harness (`usc-messaging`)**
- `SimpleDApp` / `TestDestination` demo contracts + `deploy.ts` that deploys the protocol
  contracts **from the `@gluwa/usc-contracts` package** (the Outbox/Inbox/EOAValidator/
  AcknowledgmentValidator sources and their Foundry tests live there since #0d59daac), publishes
  messages, and runs the dApp-level ack worker.

**CI**
- `.github/workflows/write-ability-e2e.yml` — full local stack (cc3 `--dev` + anvil + attestor
  zombienet + relayer built from `gluwa/usc-message-relayer`) publishing a message and asserting
  delivery **including the proof-based ack round-trip** (`Outbox.messages(id).acknowledged == true`).

### Extracted to sibling repos (formerly on this branch)

- **`gluwa/usc-message-relayer`** — the off-chain relayer: vote aggregation + delivery, the
  proof-based ack submitter, and the permissionless bridge-claim submitter ("relayer on both
  sides"). CI clones + builds it for the e2e here.
- **`gluwa/usc-bridge-poc`** — the wCTC bridge PoC UI (live demo of lock → attest → deliver →
  auto-claim).
- **`@gluwa/usc-contracts`** — protocol Solidity sources + tests.

### Testing
- Rust unit/abuse tests: `common/write-ability`, `attestor::write_ability` (aggregator, ingest,
  signing, reobservation), pallet + precompile suites.
- The write-ability e2e workflow exercises the full happy path + ack on every relevant PR.
- Live on cc3-usc-dev with the Sepolia attestor set (9-of-12 quorum) since 2026-07.

### Status / out of scope
- Attestors are authorized-operator-only for v1; OutboxFactory deploy + attestor-set sync onto
  `EOAValidator` are ops tasks per network.
- **Fees/quoting deferred to Phase 2** (Quoter / RelayerFeeVault design tracked in
  usc-write-ability-research; the attestor and relayer here never validate fees).
- Live multi-node attestor e2e deferred (the single-runner e2e is the regression gate);
  Outbox *re-registration* mid-run requires an attestor restart (TODO in `resolver.rs`).
- `spec_version` intentionally kept at 126 on this branch.

https://claude.ai/code/session_01Bv5GVkV7UMs87ui9FoYJ6K


attestor/attestor/src/tasks/write_ability/listener.rs
```diff
@@ -0,0 +1,166 @@
...
+    confirmation_depth: u64,
+    last_seen: &mut u64,
+    tx: &mpsc::Sender<IndexedMessage>,
+) -> Result<()> {
+    let tip = provider.get_block_number().await?;
+    let to_block = tip.saturating_sub(confirmation_depth);
+    if to_block <= *last_seen {
+        return Ok(());
+    }
+    let from_block = *last_seen + 1;
```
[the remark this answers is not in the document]
DylanVerstraete: Not reachable in practice, so leaving as-is. Block 0 is the genesis block, and the Outbox is a runtime factory-created contract — it cannot emit `MessagePublished` in block 0, so there's no event there to miss. Both trigger conditions require an Outbox log in a block where the Outbox doesn't exist yet:

- `head_before_resolve == 0` can only occur if the chain is at genesis when the attestor boots, but then `resolve` returns `None` (no Outbox registered) and `watch()` is never entered.
- `start_block: 0` is an explicit backfill-from-genesis setting, where again no Outbox logs exist.

Expressing "scan from block 0 inclusive" correctly would mean changing the `last_seen` cursor from `u64` to `Option<u64>`, rippling into the public `poll_once` signature and the e2e test — disproportionate churn for a genesis edge that can hold no Outbox events.

attestor/attestor/src/tasks/write_ability/ingest.rs
```diff
@@ -0,0 +1,255 @@
...
+/// Validate and (if admissible) count an incoming message-vote frame.
+pub fn validate_and_count(
+    state: &MessageVoteState,
+    our_chain_key: u64,
+    bytes: &[u8],
+) -> Acceptance {
+    let Ok(vote) = MessageVote::decode_bytes(bytes) else {
+        tracing::warn!("⛔ undecodable message vote — rejecting");
+        return Acceptance::Reject;
+    };
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in commit 00aca2390 — the change is in `common/write-ability/src/envelope.rs` (the file this `ingest.rs` line calls into), not here, which is why it didn't auto-resolve. `MessageVote::decode_bytes` (and `ReobservationRequest::decode_bytes`) now use `decode_all`, so trailing bytes are rejected. Added a `rejects_trailing_bytes` regression test.

attestor/attestor/src/tasks/write_ability/listener.rs
```diff
@@ -0,0 +1,166 @@
...
+            start_block = start,
+            "⏮️ no persisted Outbox cursor; starting message-attestation scan from configured block"
+        );
+        start.saturating_sub(1)
+    } else {
+        provider
+            .get_block_number()
+            .await
+            .context("failed to read Creditcoin L1 chain head")?
+    };
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in commit 4eefd8668 — the change is in `tasks/write_ability/mod.rs` (the `run` loop), not `listener.rs`, which is why it didn't auto-resolve. The chain head is now captured *before* the resolve-retry loop and passed as the default scan origin (`cfg.start_block.or(Some(head_before_resolve))`), so messages published on the Outbox during the resolve window are covered instead of being skipped by a post-resolution head.

attestor/attestor/src/tasks/write_ability/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+    // Re-sign unconditionally (the per-`message_id` cooldown above is the bound). We must NOT skip
+    // just because our *own* aggregator already saw local quorum: the requester is the relayer, and
+    // the whole reason it asked is that it is missing votes the attestor mesh may have settled among
+    // itself. Re-gossiping is idempotent at the relayer (it dedups), so the worst case is harmless.
+    tracing::info!(
+        %message_id,
+        message_hash = %indexed.message_hash,
+        "♻️ re-signing reobserved message"
+    );
+    produce_vote(state, signer, our_address, chain_key, indexed);
```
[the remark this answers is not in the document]
DylanVerstraete: Good catch — fixed in 5143ee1a9. `reobserve()` now enforces the same finality bound the listener signs under (`block_height + confirmation_depth <= tip`) before scanning for the log, and ignores pre-final requests. This mattered more than the usual gate-consistency nit: the destination Inbox validates quorum signatures, not source-chain state, so quorum signatures extracted over a publish that subsequently reorged away would still deliver — for the bridge, a release without the corresponding escrow. Legitimate reobservation targets are stalled *old* messages, so ignoring pre-final requests costs nothing (the relayer re-requests on its own cadence).

attestor/attestor/src/tasks/write_ability/aggregator.rs
```diff
@@ -0,0 +1,247 @@
...
+        };
+        if !entry.signers.insert(signer) {
+            return VoteOutcome::Duplicate;
+        }
+        entry.last_update = now;
+        let reached_threshold = !entry.completed && entry.signers.len() >= self.threshold;
+        if reached_threshold {
+            entry.completed = true;
+        }
+        VoteOutcome::Accepted { reached_threshold }
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 14f463227. `set_threshold` now re-evaluates every tracked aggregate: a raised quorum un-latches below-quorum completions (with a fresh TTL window — the old latch also kept them in the completed retention class, surviving TTL and evicted last) so the signal re-fires when the new quorum is genuinely met; a lowered quorum completes already-sufficient entries and returns them so the set watcher surfaces the milestone (no later vote transition exists to fire it). Both directions pinned by new unit tests. Worth noting the blast radius was local observability only — relayers count delivery quorum independently against the on-chain validator and already re-evaluate on set changes — but the aggregator's signal is now honest against the live threshold.

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +47,106 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+
+    const [messageId, emitterAddress, requiresAck, payload] = event.args;
+    logger.info(`MessagePublished: messageId=${messageId}, emitter=${emitterAddress}, requiresAck=${requiresAck}`);
+
+    // Keyed by messageId so handleMessageAcknowledged can load-and-update the same record.
+    // outboxId references the OutboxContract created by handleOutboxCreated (same lowercased address).
+    const message = OutboxMessage.create({
+        id: messageId,
+        outboxId: event.address ? event.address.toLowerCase() : '',
+        emitter: emitterAddress,
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 151b585c2 — an absent `event.address` now skips loudly (mirroring the surrounding `args`/`transactionHash` guards) instead of saving an OutboxMessage with `outboxId: ''` dangling outside every OutboxContract relation. Defensive-only in practice (EVM logs always carry the emitter address), but the failure mode is now a visible error log rather than a silently unqueryable record.

attestor/attestor/src/tasks/p2p/mod.rs
```diff
@@ -276,8 +335,13 @@ pub async fn run(
...
                 // the next retry tick.
-                if !could_broadcast && can_broadcast && !retry_queue.is_empty() {
-                    flush_retry_queue(&mut swarm, &topic, &mut retry_queue);
+                if !could_broadcast && can_broadcast {
+                    if !retry_queue.is_empty() {
+                        flush_retry_queue(&mut swarm, &topic, &mut retry_queue);
+                    }
+                    if let Some(mv_topic) = &mv_topic {
+                        flush_mv_retry_queue(&mut swarm, mv_topic, &mut mv_retry_queue);
+                    }
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 001e1da82 — one mesh hint per topic (`can_broadcast` / `can_broadcast_mv`), each retry queue flushed on its own topic's mesh edge. Agreed on the severity: both failure modes were bounded (a flush into a peerless mv mesh just stays queued; a missed edge deferred to the 30s retry tick — delay, never loss, since the publish-error requeue + periodic tick backstop everything), but the per-topic edge is the correct trigger and costs one extra in-memory `mesh_peers` lookup per event.

attestor/attestor/src/tasks/write_ability/attestor_set.rs
```diff
@@ -0,0 +1,129 @@
...
+                    *guard = set.clone();
+                    old
+                };
+                // Re-evaluates every tracked aggregate against the new quorum; a lowered threshold
+                // can push already-collected messages over it, in which case the milestone must be
+                // surfaced here — there is no later vote transition to fire it.
+                let newly_completed = state
+                    .aggregator
+                    .lock()
+                    .set_threshold(new_threshold, std::time::Instant::now());
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 9e825cd09. Added `VoteAggregator::retain_signers`: on hot-reload the aggregator drops removed signers' signatures across every tracked entry (un-completing entries that fall below quorum, mirroring `set_threshold`'s raised-quorum semantics), and the hot-reload path calls it *before* `set_threshold` so completion is re-evaluated against pruned counts. Regression test covers discount + re-open + single re-fire with current signers only.

.github/workflows/write-ability-e2e.yml
```diff
@@ -0,0 +1,295 @@
...
+      # The relayer lives in its own repository (public) since the extraction — build it from
+      # source here so the e2e still exercises attestor + relayer together. Pinned to `main`;
+      # cross-repo protocol drift is caught by the shared write-ability golden vectors on both
+      # sides. (Alternative once image publishing is wired: docker cp the binary out of
+      # gluwa/usc-message-relayer:main.)
+      - name: Check out usc-message-relayer
+        uses: actions/checkout@v6
+        with:
+          repository: gluwa/usc-message-relayer
+          path: usc-message-relayer
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 9e825cd09 — `ref: main` added to the checkout so the pin the comment promises is actually enforced.

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +47,113 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+        publishedAt: BigInt(event.blockNumber),
+        publishedTimestamp: eventTimestamp(event),
+        publishedTxHash: event.transactionHash,
+        acknowledged: false,
+        acknowledgedAt: undefined,
+        acknowledgedTimestamp: undefined,
+        acknowledgedTxHash: undefined,
+    });
+
+    await message.save();
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 9e825cd09. `handleMessagePublished` now loads the existing record first and skips the replay — publish fields are immutable per messageId, so there is nothing to update, and the acknowledged state set by `handleMessageAcknowledged` is preserved.

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +47,113 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+        factoryId,
+        createdAt: BigInt(event.blockNumber),
+        createdTimestamp: eventTimestamp(event),
+        createdTxHash: event.transactionHash,
+    });
+    await outbox.save();
+
+    // Spin up a dynamic datasource that indexes this Outbox's messages. `{ address }` is spread
+    // into the 'Outbox' template's processor.options, binding it to this instance.
+    await createOutboxDatasource({ address });
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 9e825cd09. `handleOutboxCreated` now checks `OutboxContract.get(address)` and skips when the Outbox is already registered, so a replayed `OutboxCreated` log can no longer register a duplicate dynamic datasource (which would have made every later `MessagePublished` on that Outbox fire once per duplicate).

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +47,132 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 75528a642. The save-then-register ordering meant a `createOutboxDatasource` failure after `outbox.save()` committed would strand the Outbox: the entity-existence idempotency guard would trip on retry and skip registration forever, so its MessagePublished / MessageAcknowledged events were never indexed. Reordered to register the datasource first and save the entity last, so the saved `OutboxContract` is a 'fully processed' marker — a failed registration leaves no entity and the retry re-runs both steps cleanly. prettier + eslint clean.

cc3-indexer/src/mappings/mappingHandlers.ts
```diff
@@ -196,6 +197,32 @@ export async function handleSupportedChainRegistered(event: SubstrateEvent): Pro
...
+
+    logger.info(`OutboxFactoryRegistered: factory=${address}, chainKey=${chainKey.toString()} at block ${blockNumber}`);
+
+    const factory = OutboxFactory.create({
+        id: address,
+        chainKey: BigInt(chainKey.toString()),
+        registeredAt: blockNumber,
+        registeredTimestamp: event.block.timestamp ? BigInt(event.block.timestamp.getTime()) : BigInt(0),
+    });
+    await factory.save();
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in ef34e1e74. `handleOutboxFactoryRegistered` now does an `OutboxFactory.get(address)` guard and skips if the factory is already recorded, so a repeat `set_outbox_factory_addr` update or a reprocessed block no longer re-`create`/`save`s a duplicate id. The entity only backs display + `OutboxContract.factory` resolution, so keeping the first registration is sufficient.

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +47,139 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+
+    const outbox = OutboxContract.create({
+        id: address,
+        chainKey,
+        factoryId,
+        createdAt: BigInt(event.blockNumber),
+        createdTimestamp: eventTimestamp(event),
+        createdTxHash: event.transactionHash,
+    });
+    await outbox.save();
```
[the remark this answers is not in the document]
DylanVerstraete: This and the earlier 'Outbox datasource never retried' finding are the two opposite horns of the same dilemma, so I want to be precise rather than just flip the order again.

Checked the installed `@subql/node-core`: the handler-facing `createDynamicDatasource` is invoked at `indexer.manager.js:58` **without a tx**, and `dynamic-ds.service.ts` always `setNewDynamicDatasource` (no dedup). So datasource registration commits **immediately, outside** the block's entity transaction, while `entity.save()` is buffered and commits at block end. Given that:

- **Ordering is equivalent.** In both create-first and save-first, the datasource persists the moment `createDynamicDatasource` is called and the entity persists at block commit. The only duplicate window is: datasource persisted → block-commit (entity flush) fails → retry re-creates. That window exists identically in either order — swapping to save-first would just re-introduce the earlier finding without closing this one.
- **The stated trigger (reprocessed block / reorg replay) is handled** by the `OutboxContract.get` guard: on replay of a *committed* block the entity exists and we skip, so no duplicate.
- The residual mid-block-flush-crash duplicate is a **framework-level limitation** of non-transactional dynamic datasources in this SubQuery version, not fixable at the handler layer (there's no handler-visible API to query already-registered datasources, and `createDynamicDatasource` neither dedups nor joins the block tx).

Keeping create-first as it matches SubQuery's documented dynamic-datasource pattern. Resolving as a known framework limitation with the replay path guarded.

attestor/attestor/src/tasks/write_ability/mod.rs
```diff
@@ -0,0 +1,577 @@
...
+                    %validator, %err, attempt,
+                    next_retry_secs = backoff.as_secs(),
+                    "attestor-set read failed at startup; retrying"
+                );
+                tokio::time::sleep(backoff).await;
+                backoff = (backoff * 2).min(STARTUP_SET_FETCH_MAX_BACKOFF);
+            }
+        }
+    }
+    None
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in c8b99f002. build_state no longer resolves the OnChainValidator set inline — it starts the set **empty** (ingest rejects every vote, signer ∉ set, until it's populated) and returns immediately, so Attestor::run spawns block-attestation / p2p / production without waiting on the destination RPC. The already-spawned attestor_set::watch task populates the set + aggregator threshold on its first tick (fires immediately, off the core-startup path) and re-polls every 30s. That per-poll retry also recovers from a startup RPC blip *without* a restart, which supersedes the removed ~94s bounded-retry (I dropped fetch_attestor_set_with_retry + its constants). Static sets still resolve inline; clear misconfigs (empty static set, missing destination_eth_rpc_url) still disable message attestation. Build/clippy/fmt clean, 26 write-ability tests pass.

usc-messaging/scripts/launch-attestors.sh
```diff
@@ -0,0 +1,205 @@
...
+      i++;
+      while (i < lines.length && /^\s+-\s/.test(lines[i])) i++; // drop old entries
+      addrs.forEach((a, n) => out.push(`    - "${a}"  # zombie-${n}`));
+      continue;
+    }
+    i++;
+  }
+  fs.writeFileSync(path, out.join("\n"));
+' "$CONFIG" "$SET"
+echo "✍️  Updated $CONFIG  (write_ability.attestors)"
```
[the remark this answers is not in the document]
DylanVerstraete: Addressed in 1d53dd09d. This is the dev-only zombienet **demo launcher**, and you've correctly described a real limitation of it: it runs with random per-attestor seeds (no `--well-known-keys`), so each EVM signer address is only knowable after boot — hence the post-startup config rewrite, which static sets don't hot-reload. A restart wouldn't help (new random seeds → new addresses). Delivery still works (each attestor gossips its own vote; the relayer reaches quorum via `--attestor-set`), but attestor-side in-mesh aggregation is against the stale set for the run.

The correct fix isn't a launcher rewrite — it's to run the attestors in **OnChainValidator mode**: set `write_ability.validator_address` to the EOAValidator this script already syncs with the discovered addresses (lines 157-171), and the on-chain attestor-set watcher converges every node's active set within one poll interval, no restart. I've made that explicit + recommended in the launcher output rather than rewriting the random-seed demo flow (which I can't validate end-to-end here). Static-list demos remain fine for the delivery path. Resolving as a documented dev-demo limitation with the correct production-shaped path called out.

cc3-indexer/src/mappings/evmHandlers.ts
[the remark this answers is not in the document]
DylanVerstraete: Fixed by saving the OutboxContract before createOutboxDatasource. You are right that a dynamic datasource created mid-block handles the same block's later events, so a same-block MessagePublished must find the parent staged — save-first guarantees that (both writes land in the same per-block store transaction). This reconciles the three passes on these lines: SubQuery buffers .save() into the block tx, so save-first does NOT strand the datasource on a partial failure (if createOutboxDatasource throws, the whole block — including the save — rolls back and the retry re-runs both). The residual duplicate-on-crash window is identical in either order (the datasource-metadata write is the only non-transactional step; framework limitation) and the OutboxContract.get guard covers the common replay/reorg case. SubQuery sets the datasource start block to the current block automatically. prettier + eslint clean.

attestor/attestor/src/tasks/write_ability/reobservation.rs
```diff
@@ -0,0 +1,326 @@
...
+    };
+    if !final_enough {
+        tracing::warn!(
+            block = request.block_height,
+            ?finalized,
+            confirmation_depth,
+            "🔎 reobservation request targets a not-yet-finalized block — ignoring"
+        );
+        return Ok(None);
+    }
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 6c43960f6. Reobservation now shares the listener's exact finality envelope: a block is re-signable once it is EITHER GRANDPA-finalized OR buried under the governed `confirmation_depth` (the same fallback the listener uses when finality stalls or the `finalized` tag is missing). Extracted as a pure, unit-tested `reobs_final_enough` predicate so the two gates can't drift, with a test covering the stall-recovery case (block above the stalled finalized head, depth-buried).

atodorov (changes requested):
Needs rebase onto latest usc-dev, then all of its child PRs need rebase too so we can see CI workflows executing in the child PRs.

.github/workflows/write-ability-e2e.yml
```diff
@@ -0,0 +1,300 @@
...
+    uses: ./.github/workflows/remove-runner.yml
+    with:
+      RUNNER_VM_NAME: "${{ github.event.repository.name }}-${{ github.run_id }}-attempt-${{ github.run_attempt }}"
+      proxy: ${{ matrix.proxy }}
+    secrets: inherit
+
+  e2e:
+    needs:
+      - build
+    runs-on: ubuntu-24.04
```
atodorov, line 86: Already on ubuntu 26. Needs update.

.github/workflows/write-ability-e2e.yml
```diff
@@ -0,0 +1,300 @@
...
+      # wired: docker cp the binary out of a SHA-tagged gluwa/usc-message-relayer image.)
+      - name: Check out usc-message-relayer
+        uses: actions/checkout@v6
+        with:
+          repository: gluwa/usc-message-relayer
+          # Pin a full commit SHA (not a branch) so this creditcoin3 commit always builds against
+          # the *same* relayer code — `ref: main` made the e2e non-reproducible (the same commit
+          # could pass or fail as the relayer's default branch moved). Bump this deliberately when
+          # a relayer change needs to be exercised here.
+          ref: b5c71a3499a699d4e1ac180b85f420d6c7883aeb # usc-message-relayer main @ 2026-07-14 (#21)
```
atodorov: This will be difficult to maintain by hand and supposedly we'd want to be testing relayer with latest creditcoin3 and vice versa to catch issues quickly. Pinning an exect commit defeats this purpose.

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +63,197 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+// Outbox: MessagePublished(bytes32 indexed messageId, address indexed emitterAddress, bool requiresAck, bytes payload)
+type MessagePublishedArgs = [string, string, boolean, string];
+// Outbox: MessageAcknowledged(bytes32 indexed messageId)
+type MessageAcknowledgedArgs = [string];
+
+function eventTimestamp(event: { blockTimestamp?: Date }): bigint {
+    return event.blockTimestamp ? BigInt(event.blockTimestamp.getTime()) : BigInt(Date.now());
+}
+
+export async function handleOutboxCreated(event: FrontierEvmEvent<OutboxCreatedArgs>): Promise<void> {
```
atodorov, line 81: No test for this handler.

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +63,197 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+    // see same-block buffered writes).
+    if (!authenticated) {
+        const counter = await OutboxDatasourceCount.get(chainKey);
+        await OutboxDatasourceCount.create({ id: chainKey, count: (counter?.count ?? 0) + 1 }).save();
+    }
+
+    await createOutboxDatasource({ address });
+}
+
+export async function handleMessagePublished(event: FrontierEvmEvent<MessagePublishedArgs>): Promise<void> {
```
atodorov, line 189: No test for this handler.

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +63,197 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+        acknowledged: false,
+        acknowledgedAt: undefined,
+        acknowledgedTimestamp: undefined,
+        acknowledgedTxHash: undefined,
+    });
+
+    await message.save();
+}
+
+export async function handleMessageAcknowledged(event: FrontierEvmEvent<MessageAcknowledgedArgs>): Promise<void> {
```
atodorov, line 244: No test for this handler.

cc3-indexer/src/mappings/mappingHandlers.ts
```diff
@@ -196,6 +197,42 @@ export async function handleSupportedChainRegistered(event: SubstrateEvent): Pro
     await Promise.all([chainRegistered.save(), suportedChain.save(), newChain.save()]);
 }
 
+// USC write-ability: an operator registered an OutboxFactory for a chain key
+// (supportedChains.OutboxFactoryRegistered { chain_key, outbox_factory_addr }). Recorded for
+// display / to resolve OutboxContract.factory. Discovery of Outboxes does NOT depend on this event
+// — see the chain-wide OutboxCreated watch in datasources.ts (the factory creates Outboxes before
+// it is registered here, so following this event would miss them).
+export async function handleOutboxFactoryRegistered(event: SubstrateEvent): Promise<void> {
```
atodorov, line 205: No test for this handler.

pallets/supported-chains/src/lib.rs
```diff
@@ -288,6 +404,110 @@ pub mod pallet {
...
+        /// CTC (paid via `msg.value`); `Some(address)` in an ERC20 on Creditcoin's EVM
+        /// (attestcoin, pulled via `transferFrom`). A zero `amount` disables the fee. Only
+        /// accounts in the Operators membership (or root) can call this extrinsic.
+        ///
+        /// The value is read live by the EVM through the chain-info precompile
+        /// (`get_core_fee(uint64)`), so changes take effect on the next publish without any
+        /// contract redeploys — including a later native→attestcoin denomination switch.
+        #[pallet::call_index(4)]
+        #[pallet::weight(T::WeightInfo::set_core_fee())]
+        pub fn set_core_fee(
```
atodorov, line 487: Not seeing integration test for this extrinsic.

precompiles/chain-info/src/lib.rs
```diff
@@ -159,6 +177,65 @@ where
...
+            None => Ok(OutboxFactoryResult {
+                factory_addr: Address(H160::zero()),
+                exists: false,
+            }),
+        }
+    }
+
+    #[precompile::public("get_core_fee(uint64)")]
+    #[precompile::view]
+    fn get_core_fee(
```
atodorov, line 209: missing integration test for this function.

pallets/supported-chains/src/lib.rs
```diff
@@ -133,6 +177,34 @@ pub mod pallet {
...
+                    },
+                );
+            }
+
+            for (chain_key, outbox_factory_addr) in &self.outbox_factories {
+                // Mirror the `set_outbox_factory_addr` guard: a zero factory reads as "unregistered"
+                // to the attestor/relayer, so a chainspec must not seed one.
+                assert!(
+                    !outbox_factory_addr.is_zero(),
+                    "Genesis outbox_factories has a zero address for chain_key {chain_key:?}"
```
atodorov, line 204: Question: will this silently be a no-op when doing a runtime upgrade on an existing chain or will it cause a failure?

pallets/supported-chains/src/lib.rs
```diff
@@ -273,6 +383,12 @@ pub mod pallet {
 
             ChainIdAndNameToUniqKey::<T>::remove(item.chain_id, item.chain_name.clone());
 
+            OutboxFactories::<T>::remove(chain_key);
+
+            WriteAbilityConfigs::<T>::remove(chain_key);
+
+            CoreFees::<T>::remove(chain_key);
```
atodorov, line 390: All of these need to be asserted to in

2) cc3-indexer tests
3) cc3-indexer handle mappers

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+            ),
+            Error::<Test>::ZeroWriteAbilityChainKey
+        );
+
+        assert_eq!(WriteAbilityConfigs::<Test>::get(chain_key), None);
+    });
+}
+
+#[test]
+fn remove_chain_clears_write_ability_config() {
```
atodorov, line 794: IMO should be merged with the existing `remove_chain` test and just adding the assertion at the bottom of it. Will make it easier to read/trace intended behavior & test coverage.

Also merge remove_chain_clears_core_fee() into that too. 

Add assertion that OutboxFactories have been removed too.

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+            SupportedChain::set_write_ability_config(
+                RuntimeOrigin::signed(2),
+                chain_key,
+                [0u8; 32],
+                true,
+            ),
+            BadOrigin
+        );
+
+        assert_ok!(SupportedChain::set_write_ability_config(
```
atodorov, line 734: Breaks separation of concerns: this action & assertion are not supposed to be here. If you want to explicitly exercise this extrinsic with an operator account (and not root) then duplicate/parametrize set_write_ability_config_works().

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+            ),
+            Error::<Test>::ChainNotSupported
+        );
+
+        assert_eq!(WriteAbilityConfigs::<Test>::get(chain_key), None);
+    });
+}
+
+#[test]
+fn set_write_ability_config_should_error_when_chain_key_is_zero() {
```
atodorov, line 773: Incorrect name: should be "_when_write_ability_key_is_zero".

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+        let chain_key = 1;
+
+        assert_noop!(
+            SupportedChain::set_write_ability_config(
+                RuntimeOrigin::root(),
+                chain_key,
+                [0u8; 32],
+                true,
+            ),
+            Error::<Test>::ZeroWriteAbilityChainKey
```
atodorov, line 786: Q: How is that even working? According to the previous scenario `chain_key = 1` isn't supported and we should have seen `ChainNotSupported` error here?

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+        );
+    });
+}
+
+#[test]
+fn set_core_fee_works_native() {
+    ExtBuilder.build_and_execute(|| {
+        System::set_block_number(1);
+
+        let chain_key = 1;
```
atodorov, line 885: Something is completely off in all of these tests. Once you are using `chain_key = 1` as an unregistered chain and another time you are using the same value to represent an already supported chain and I'm not seeing where it gets registered. 

This seems completely wrong and is not obvious what's happening.

precompiles/chain-info/src/tests.rs
```diff
@@ -143,6 +261,36 @@ fn get_chain_by_key_returns_default_data_with_unknown_chain_key() {
         });
 }
 
+#[test]
+fn get_outbox_factory_address_returns_default_when_not_set() {
```
atodorov, line 265: nitpick: move under get_outbox_factory_address_works() - easier to read.

pallets/supported-chains/src/lib.rs
```diff
@@ -288,6 +404,110 @@ pub mod pallet {
...
+            token: Option<H160>,
+            amount: sp_core::U256,
+        ) -> DispatchResult {
+            T::OperatorsOrigin::ensure_origin(origin)?;
+
+            ensure!(
+                SupportedChains::<T>::contains_key(chain_key),
+                Error::<T>::ChainNotSupported
+            );
+            ensure!(token != Some(H160::zero()), Error::<T>::ZeroCoreFeeToken);
```
atodorov, line 499: This contradicts the behavior inside chain-info::get_core_fee().

The impementation requires that token is not the zero address so we pass None when we want native fees, however get_core_fee() will return a zero address in the same scenario.

precompiles/chain-info/src/lib.rs
```diff
@@ -159,6 +177,65 @@ where
...
+                // sentinel on the EVM side. The extrinsic rejects Some(H160::zero()), so a zero
+                // token here always means native, never a broken ERC20 config.
+                token: Address(fee.token.unwrap_or_default()),
+                amount: fee.amount,
+                exists: true,
+            }),
+            None => Ok(CoreFeeResult {
+                token: Address(H160::zero()),
+                amount: U256::zero(),
+                exists: false,
```
atodorov, line 234: Adding to my comment on line R499 in supported-chains/src/lib.rs:

The only distinction whether the return value of this function is indeed fee set in native chain currency vs. non-existing chain the the `exists` field. 

Given that anywhere else in extrinsics we error out when the chain_key isn't supported shouldn't we just error out from the precompile too? For other precompile functions we already revert when chain_key is unsupported.

precompiles/metadata/precompiles-creditcoin3-testnet.json
```diff
@@ -20,8 +20,8 @@
...
...sted);\n\n    /**\n     * @dev Get attestation bounds for a target height\n     * @notice Returns the parent (highest attested before) and child (lowest attested after) for a target height\n     * @param chainKey The chain key to query\n     * @param targetHeight The target height to find bounds for\n     * @return result Attestation bounds result\n     */\n    function get_attestation_bounds(uint64 chainKey, uint64 targetHeight)\n        external\n        view\n        returns (BoundsCheckResult memory result);\n\n    /**\n     * @dev Get attestation height by its digest\n     * @param chainKey The chain key to query\n     * @param digest The attestation digest to look up\n     * @return result Attestation height result\n     */\n    function get_attestation_height_for_digest(uint64 chainKey, bytes32 digest) external view returns (HeightResult memory);\n\n    /**\n     * @dev Get checkpoint by its digest/hash\n     * @param chainKey The chain key to query\n     * @param height The checkpoint height to look up\n     * @return result Checkpoint hash result\n     */\n    function get_checkpoint_for_height(uint64 chainKey, uint64 height) external view returns (HashResult memory);\n}",
```
atodorov, line 23: precompiles-creditcoin3-mainnet.json must also be updated.

cc3-indexer/abis/outbox.json
```diff
@@ -0,0 +1,21 @@
+[
```
atodorov, line 1: Q: where is the canonical source for this ABI and/or it's curresponding .sol file? 

We probably want to make sure this is kept in sync .

usc-messaging/contracts/src/TestDestination.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT
```
atodorov, line 1: Q: what's the purpose of this contract? I don't get it.

atodorov (changes requested):
Added initial set of comments around tests. There are also multiple failing CI jobs, most importantly: 
 
- cc3-indexer failures: https://github.com/gluwa/creditcoin3/actions/runs/29565283426/job/87839835228
- writeability e2e failures: https://github.com/gluwa/creditcoin3/actions/runs/29565283195/job/87839959757

attestor/attestor/src/tasks/write_ability/listener.rs
```diff
@@ -0,0 +1,449 @@
...
+                    // Finality is lagging but not stalled — stay at the finalized head.
+                    tracker.in_fallback = false;
+                    f
+                }
+            }
+            None => {
+                // Chain reports no finalized head (or the tag is unsupported / errored this poll).
+                tracker.in_fallback = true;
+                tip.saturating_sub(fallback_depth)
+            }
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 33f52d659. The no-finalized-head fallback now clamps `tip - depth` to `tracker.last_finalized`, so a transient `finalized`-read blip can't regress the scan below a known-final head. Mirrors the stall arm's `.max(f)`; added a unit test.

attestor/attestor/src/tasks/p2p/mod.rs
```diff
@@ -332,16 +446,102 @@ fn drain_pending_votes(
...
+                TokenBucket::new(REOBS_PER_PEER_CAPACITY, REOBS_PER_PEER_REFILL_PER_SEC, now),
+            );
+            self.order.push_back(source);
+        }
+        // unwrap: just ensured the entry exists.
+        if !self.per_peer.get_mut(&source).unwrap().try_take(now) {
+            return false;
+        }
+        self.global.try_take(now)
+    }
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 33f52d659. `admit` now refills both buckets, then consumes a token from each only when BOTH have one — a global-pressure denial no longer burns the peer's per-peer token.

DylanVerstraete:
bugbot run

DylanVerstraete:
bugbot run

DylanVerstraete:
closing in favor of https://github.com/gluwa/creditcoin3/pull/1215

PR #1086, opened 2026-06-23
fix(proof-gen-api): reject tx proofs for empty blocks

creditcoinprotoclaw:
## Problem

`/api/v1/proof/{chain_key}/{header_number}/{tx_index}` previously accepted `tx_index=0` against an empty block and returned an empty `TransactionMerkleProof`. The block-prover precompile rejects empty transaction bytes (`"Transaction data cannot be empty"` — see `precompiles/block-prover/src/verify.rs:207`), so those proofs could never verify on-chain — the API was returning payloads guaranteed to fail downstream.

## Fix

- New `ServiceError::EmptyBlockTxProof { height }`, mapped to **422 Unprocessable Entity** (same group as `BlockNotReady` / `UnsupportedBlockFormat` — request is well-formed but the block can't produce a transaction proof).
- `generate_merkle_proof()` returns `EmptyBlockTxProof` when `tx_bytes.is_empty()` instead of producing a useless proof. The downstream empty-branch handling in the `spawn_blocking` + cache assembly is dead now (the early return covers it), so the helper is simplified.
- New `ErrorType::EmptyBlockTxProof` for prometheus labels and route docstring update describing the new behavior.

## Behavior change for callers

Clients that previously sent `tx_index=0` for an empty block now get:

```
HTTP 422 EmptyBlockTxProof
{"code":"EmptyBlockTxProof","message":"block N is empty; tx proof unavailable (use a continuity-only proof)"}
```

If continuity-only attestations for empty blocks become a real use case, that's a separate endpoint — intentionally not adding one in this PR to keep the surgical fix focused.

## Audit item addressed

> "proof-gen-api-server allows tx_index=0 for empty blocks and returns an empty Merkle proof, but the block-prover precompile rejects empty transaction bytes, so the API can return proofs that cannot verify on-chain"

### DIFF
--- a/proof-gen-api-server/src/networking/routes/continuity.rs
+++ b/proof-gen-api-server/src/networking/routes/continuity.rs
@@ -24,13 +24,13 @@ type TxHashes = Vec<String>;
     params(
         ("chain_key" = u64, Path, description = "Chain identifier"),
         ("header_number" = u64, Path, description = "Block number"),
-        ("tx_index" = u64, Path, description = "Transaction index in the block (0 for empty blocks)")
+        ("tx_index" = u64, Path, description = "Transaction index in the block. Empty blocks have no transaction proof; request EmptyBlockTxProof (422) for those.")
     ),
     responses(
         (status = 200, description = "Continuity and merkle proof", body = SingleContinuityResponse),
         (status = 400, description = "Invalid parameters", body = ErrorResponse),
         (status = 404, description = "Block or transaction not found", body = ErrorResponse),
-        (status = 422, description = "Unprocessable: block not yet attested (BlockNotReady), or source RPC block/header inconsistent (UnsupportedBlockFormat)", body = ErrorResponse),
+        (status = 422, description = "Unprocessable: block not yet attested (BlockNotReady), source RPC block/header inconsistent (UnsupportedBlockFormat), or block contains no transactions (EmptyBlockTxProof)", body = ErrorResponse),
         (status = 503, description = "RPC unavailable", body = ErrorResponse)
     )
 )]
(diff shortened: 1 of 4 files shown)


DylanVerstraete (approved):

proof-gen-api-server/src/networking/routes/continuity.rs
```diff
@@ -24,13 +24,13 @@ type TxHashes = Vec<String>;
...
         ("header_number" = u64, Path, description = "Block number"),
-        ("tx_index" = u64, Path, description = "Transaction index in the block (0 for empty blocks)")
+        ("tx_index" = u64, Path, description = "Transaction index in the block. Empty blocks have no transaction proof; request EmptyBlockTxProof (422) for those.")
     ),
     responses(
         (status = 200, description = "Continuity and merkle proof", body = SingleContinuityResponse),
         (status = 400, description = "Invalid parameters", body = ErrorResponse),
         (status = 404, description = "Block or transaction not found", body = ErrorResponse),
-        (status = 422, description = "Unprocessable: block not yet attested (BlockNotReady), or source RPC block/header inconsistent (UnsupportedBlockFormat)", body = ErrorResponse),
+        (status = 422, description = "Unprocessable: block not yet attested (BlockNotReady), source RPC block/header inconsistent (UnsupportedBlockFormat), or block contains no transactions (EmptyBlockTxProof)", body = ErrorResponse),
```
atodorov, line 33: NOTE: this behaviour is actually what some of the prover check tools use to figure out whether a block has any transactions or not (without querying the source chain). 

Once deployed will have to refactor the code which relies on the current behavior to inspect the status code instead.

atodorov (approved):

PR #1089, opened 2026-06-23
fix(attestor/metrics): rollback reset + connected-peers gauge + gossipsub counter

creditcoinprotoclaw:
Three attestor observability fixes from the audit pass. Additive — no metric was renamed or removed, so existing prometheus dashboards keep working.

## 1. Reset production / lag metrics on rollback

The `RevertedAttestationChainTo` arm in the production worker updated `attestation_latest_cc3` / `attestation_local` internal state but left the prometheus gauges showing pre-revert heights and a stale lag measurement until the next forward step. Operators saw the chain at the wrong height and a spurious negative lag in dashboards.

Now we `set_attestation_local` / `set_attestation_finalized` to the revert height immediately and recompute eth/cc3 lag against the rolled-back local height. cc3 lag collapses to zero by definition after a revert.

## 2. Add `connected_peers` gauge (separate from Kademlia routing-table peers)

The existing `peer_to_peer{peer_to_peer=peers}` gauge tracks Kademlia routing-table inserts / evictions and never decremented on `ConnectionClosed`. Its docstring said "Active peer count" but the value really meant "known peers in the routing table".

The routing-table gauge stays so existing dashboards don't break; the new `connected_peers` gauge is bumped from `ConnectionEstablished` and decremented on `ConnectionClosed`, answering the more common "how many attestors am I actually talking to right now" question.

## 3. Add `gossipsub_messages_total` Counter alongside the legacy gauge

Gossipsub message counts are monotonic and belong in a Counter; the gauge representation is silently wrap-prone and harder to reason about with `rate()`. The existing gauge stays for backwards-compatible dashboards during a deprecation window; new dashboards should target `gossipsub_messages_total`.

## Audit items addressed

- "During rollback handling … only the internal `attestation_latest_cc3` and `attestation_local` states are reverted, while metrics remain unchanged…"
- "describes a metric as 'Active peer count', but the value actually reflects insertions and evictions in the Kademlia routing table … The metric is not decremented on `ConnectionClosed`."
- "stores both peer counts and Gossipsub message counts in the same `Gauge<u64>` family, even though message counts are monotonically increasing…"

## Deferred (separate follow-ups)

- Attestation generation latency metric description / start-time semantics (item 4): worth a separate look at whether to propagate the creation timestamp through `StreamAttestation` or just relabel.
- `StreamAttestation::new` `tip = 0` init (item 5): touches stream init lifecycle.

### DIFF
--- a/attestor/attestor/src/worker/p2p/mod.rs
+++ b/attestor/attestor/src/worker/p2p/mod.rs
@@ -584,6 +584,11 @@ impl WorkerP2P {
                 ..
             } => {
                 tracing::info!(%peer_id, connection = %connection_id, "🔗 Connection established");
+                // Track *currently connected* peers separately from the Kademlia routing-table
+                // gauge (increase_peer_count below), which only fires on routing-table
+                // inserts. The two values drift in normal operation — a peer can be in the
+                // routing table without an active connection.
+                self.metrics.note_peer_connected();
             }
 
             // Disconnected from an existing remote peer
@@ -593,6 +598,7 @@ impl WorkerP2P {
                 ..
             } => {
                 tracing::info!(%peer_id, connection = %connection_id, "⛓️‍💥 Connection closed");
+                self.metrics.note_peer_disconnected();
 
                 self.ping_failures.remove(&connection_id);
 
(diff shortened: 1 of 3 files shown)


creditcoinprotoclaw:
Closing — redoing on `attestor_v2` per Dylan. Items 1 and 3 (rollback metrics reset, gossipsub Counter split) are already in `attestor_v2`; the new PR only adds the `connected_peers` gauge.

PR #1106, opened 2026-06-25
Raise Utilized Compute Per Block to Match Polkadot Standard

BradleyOlson64:
We were under-utilizing our 15 second blocks, treating them as if they were limited to 6 seconds.

This pr fixes that, allowing USC readability to process 2.4x larger transactions without failing.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -429,20 +429,27 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for FindAuthorTruncated<F> {
     }
 }
 
-/// EVM gas available per block. Raised from 75_000_000 (2.4x) so that large
-/// cross-chain operations — e.g. verifying + decoding a Sepolia transaction — fit
-/// within a single block, which they previously could not.
+/// EVM gas available per block. The prod value was raised from 75_000_000 (2.4x)
+/// so that large cross-chain operations — e.g. verifying + decoding a Sepolia
+/// transaction — fit within a single (15s) block, which they previously could not.
 ///
-/// Kept UNIFORM across all profiles (not macro-gated) so that `fast-runtime`
-/// and `devnet` can reproduce the same oversized transactions as prod in tests.
+/// Macro-gated to track `WEIGHT_MILLISECS_PER_BLOCK`. `WeightPerGas` is derived as
+/// `NORMAL_DISPATCH_RATIO * WEIGHT_MILLISECS_PER_BLOCK * WEIGHT_REF_TIME_PER_MILLIS
+/// / BLOCK_GAS_LIMIT` (see below). The compute budget (`WEIGHT_MILLISECS_PER_BLOCK`)
+/// is itself profile-gated (5000/1666/1666 ms), so leaving this cap uniform made the
+/// derived gas->weight price diverge per profile (~20_833 ref_time/gas in prod vs
+/// ~6_941 under fast/devnet). That skew inflated EVM gas estimates ~2.9x on
+/// fast/devnet, which silently broke fee/funding assumptions calibrated against
+/// prod (e.g. accounts funded by the bridge example scripts ran out of gas).
 ///
-/// `WeightPerGas` is derived from this together with `WEIGHT_MILLISECS_PER_BLOCK`
-/// (see below). Because the compute budget is profile-gated while this limit is
-/// not, the derived gas->weight price intentionally differs per profile:
-/// ~20_833 ref_time/gas in prod vs ~6_941 under fast/devnet. That divergence is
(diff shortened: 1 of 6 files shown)


BradleyOlson64:
Confirmed that all CI failures aren't due to this pr's contents:

CI / hello-bridge / example (pull_request) -> Fails because of a test account which is out of funds

CI / integration-test-blockchain (pull_request) -> Fails because of a storage version change in pallet randomness. Not from this pr

runtime/src/lib.rs
```diff
@@ -411,7 +426,20 @@ impl<F: FindAuthor<u32>> FindAuthor<H160> for FindAuthorTruncated<F> {
...
+/// Kept UNIFORM across all profiles (not macro-gated) so that `fast-runtime`
+/// and `devnet` can reproduce the same oversized transactions as prod in tests.
+///
+/// `WeightPerGas` is derived from this together with `WEIGHT_MILLISECS_PER_BLOCK`
+/// (see below). Because the compute budget is profile-gated while this limit is
+/// not, the derived gas->weight price intentionally differs per profile:
+/// ~20_833 ref_time/gas in prod vs ~6_941 under fast/devnet. That divergence is
+/// expected and acceptable — `fast-runtime` validates behavior, not prod weight
+/// calibration.
+const BLOCK_GAS_LIMIT: u64 = 180_000_000;
```
atodorov: You need to adjust the hard-coded value of 75 mil in prover-check.ts or better yet read that from on-chain constants if possible.

atodorov (changes requested):

runtime/src/lib.rs
BradleyOlson64: Addressed 👍

atodorov (changes requested):
See failing CI jobs!

atodorov:
> CI / hello-bridge / example (pull_request) -> Fails because of a test account which is out of funds

Can you elaborate here please ? 

The account used for testing should be:
0x2C317383b8A99537c61042B63C4fB042Df51033a (NEW_LENDER_ accnt in 1Password. Used for testing): 10999.775716 CTC

which has plenty of funds.

BradleyOlson64:
>Can you elaborate here please ?

>The account used for testing should be:
>0x2C317383b8A99537c61042B63C4fB042Df51033a (NEW_LENDER_ accnt in 1Password. Used for testing): >10999.775716 CTC

>which has plenty of funds.

You were right here Alex. Seems that I needed to change `BLOCK_GAS_LIMIT` based on whether `fast-runtime` is being used. Should solve for any gas cost changes in tests.

atodorov (changes requested):
Needs a proper rebase to get rid of this commit:

Image

runtime/src/lib.rs
```diff
@@ -184,8 +184,23 @@ pub fn native_version() -> sp_version::NativeVersion {
...
+///   - prod   (15s block): 5000ms -> 33%
+///   - devnet ( 5s block): 1666ms -> 33%
+///   - fast   ( 5s block): 1666ms -> 33%
+///
+/// History: this was previously a flat 2000ms (the stock node-template default,
+/// whose comment referenced a 6s block). On Creditcoin's 15s prod block that
+/// was only ~13% — conservative. But it MUST remain macro-gated: a flat 5000ms
+/// would be 100% of a 5s `fast-runtime`/`devnet` slot, starving block import
+/// and stalling the chain. See also `BLOCK_GAS_LIMIT` below.
+pub const WEIGHT_MILLISECS_PER_BLOCK: u64 = prod_devnet_fast!(5_000, 1_666, 1_666);
```
atodorov, line 203: CC3 Devnet runs with 2000ms weight at the moment. Not sure if doing a runtime upgrade to 1666 will cause it to break (if that was block time it would). 

Have you tried forking & doing a runtime upgrade with this new change? If not we need clear confirmation that it breaks/doesn't break the chain.

BradleyOlson64: Just finished testing that runtime upgrade.

I upgraded from the latest usc-dev runtime to the runtime from this branch.

Both runtime builds used the flag `fast-runtime`, so the upgrade tested reducing compute per block by 1/6

Nothing crashed 👍

DylanVerstraete (dismissed):

atodorov (dismissed):

BradleyOlson64:
I believe we're waiting to merge this until after Attestor v2 lands.

Will rebase then

atodorov (approved):

BradleyOlson64:
Confirmed in local testing that CTC cost of EVM operations is the same before and after this change

DylanVerstraete (approved):

beqaabu (approved):

didac-gluwa (approved):

PR #1107, opened 2026-06-29
fix(prover-check): skip empty blocks instead of failing on 422

creditcoinprotoclaw:
## Problem

The `Check Prover` job [failed](https://github.com/gluwa/creditcoin3/actions/runs/28341144821/job/83955710791) on `cc3-devnet` (CK 1). Root cause:

The prover endpoint `/api/v1/proof/{chainKey}/{block}/0` now returns **HTTP 422** with body:

```json
{ "code": "EmptyBlockTxProof", "message": "block 11157959 is empty; tx proof unavailable (use a continuity-only proof)", "retriable": false }
```

for blocks that contain no transactions. Previously such blocks returned `200` with `txBytes === undefined`, which the script already handled by skipping. Now `axios.get` throws before reaching that guard, killing the whole run.

## Fix

- In `getProofForBlock`, catch the specific `422` / `EmptyBlockTxProof` response and return `null`.
- In the verification loop, skip the block when the response is `null` (mirrors the existing "no transactions in block" skip).
- All other errors are re-thrown, so genuine failures still fail the run.

## Verification

- `yarn check-format` ✅
- `yarn lint` ✅
- `yarn build` ✅

### DIFF
--- a/cli/src/scripts/prover-check.ts
+++ b/cli/src/scripts/prover-check.ts
@@ -39,8 +39,23 @@ function writeToDisk(dirPath: string, proofData: any) {
 
 async function getProofForBlock(apiUrl: string, chainKey: number, blockNumber: bigint) {
     const url = `${apiUrl}/api/v1/proof/${chainKey}/${blockNumber}/0`;
-    // NOTE: throws an exception in case of errors
-    return axios.get(url);
+    try {
+        // NOTE: throws an exception in case of errors
+        return await axios.get(url);
+    } catch (error) {
+        // The prover returns HTTP 422 with code 'EmptyBlockTxProof' for blocks
+        // that contain no transactions; there is no tx proof to verify, so we
+        // treat this as a skip rather than a hard failure. Any other error is
+        // re-thrown so genuine problems still surface and fail the run.
+        if (
+            axios.isAxiosError(error) &&
+            error.response?.status === 422 &&
+            error.response?.data?.code === 'EmptyBlockTxProof'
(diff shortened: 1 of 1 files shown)


atodorov (approved):
Example execution:
https://github.com/gluwa/creditcoin3/actions/runs/28364524301/job/84026996926
```
2026-06-29T10:12:52.2256016Z ... get proof for source chain block 11160971
2026-06-29T10:12:52.8255732Z     ... skipping verification. Empty block, no tx proof available
2026-06-29T10:12:52.8256638Z ... get proof for source chain block 11160978
2026-06-29T10:12:54.0079745Z     ..... trying to verify proof for 11160978 -> 0x4d102e727d396f9c309741cf42ec419ed3373d3986b577a877676f94cb420933
2026-06-29T10:12:54.0263377Z     ... proof verification: SUCCESS
2026-06-29T10:12:54.0456149Z     ... gasForVerification=58835
2026-06-29T10:12:54.0456756Z     ..... trying to decode proof
2026-06-29T10:12:54.1838249Z     ... decoded as type 2, gasForDecoding=622568
2026-06-29T10:12:54.1839244Z     ... totalGas (with 10% margin)=749543 (threshold=52500000)
2026-06-29T10:12:54.1839844Z ... get proof for source chain block 11160985
2026-06-29T10:12:54.7715692Z     ... skipping verification. Empty block, no tx proof available
```

DylanVerstraete (approved):

PR #1108, opened 2026-06-29
fix(audit): chain-info bucket-walk false-negative, scan cap, real force_election randomness

creditcoinprotoclaw:
Addresses three creditcoin3 audit findings (see Creditcoin Bug Submission sheet).

### #18 (Medium) — chain-info sparse-checkpoint false negative
`BUCKET_SEARCH_ATTEMPTS` was hardcoded to `5`, so `find_highest_attested_before` / `find_lowest_attested_after` gave up after 5 buckets and returned a false negative when valid checkpoints existed further away. Replaced with a **dynamic bound** derived from the real distance to `LastCheckpoint`: `ceil(distance / CHECKPOINT_BUCKET_SIZE) + 1`, clamped to a new `MAX_BUCKET_SEARCH_ATTEMPTS = 4096`. Applied symmetrically to the down-walk, up-walk, and `is_height_attested`. Per-bucket `GAS_STORAGE_LOOKUP` accounting is unchanged, so deep malicious searches still self-limit on gas.

### #21 residual (Medium) — unbounded `Attestations::iter_prefix` scans
The gas-per-item underpricing was already fixed (`GAS_STORAGE_LOOKUP = 2600` charged per item). This adds a defensive hard cap `MAX_ATTESTATION_SCAN_ITEMS = 100_000` via `.take(...)` on all 6 prefix scans (before `.inspect`, so gas still matches items actually scanned). No behavior change for normal stores.

### #26 (Medium) — force_election deterministic randomness
`force_election` previously passed a hardcoded `[0; 32]` to `do_start_election`. Now threads the **real epoch randomness** from `pallet_randomness` (the same source the scheduled path already consumes via `OnRandomnessUpdate`), via a new `Config::RandomnessProvider: RandomnessPalletProvider`. Selection semantics are unchanged (still activates all eligible attestors; no subsampling) — only the seed is now real instead of zero.

### Verification
- `cargo fmt --check`: clean
- `cargo clippy --all-targets -- -D warnings`: clean on touched crates
- `cargo check -p creditcoin3-runtime`: compiles (full runtime + WASM)
- Tests: chain-info **40 passed** (2 new bucket-distance tests); pallet-attestation **258 passed** (new `force_election_threads_real_epoch_randomness`)

Opened as **draft** for review. cc @gluwa/protocol


precompiles/chain-info/src/lib.rs
```diff
@@ -32,7 +32,41 @@ mod tests;
...
+/// Bounding the walk by the *real* checkpoint range (rather than a fixed constant) is what fixes
+/// the sparse-checkpoint false negative: a valid checkpoint far from the target is still reached.
+/// The result is `ceil(distance / CHECKPOINT_BUCKET_SIZE) + 1` (the `+1` covers the partial bucket
+/// at each end), clamped to `MAX_BUCKET_SEARCH_ATTEMPTS` so the loop is always bounded.
+fn bucket_search_attempts(target_height: u64, last_checkpoint_height: u64) -> u32 {
+    let distance = target_height.abs_diff(last_checkpoint_height);
+    // ceil(distance / CHECKPOINT_BUCKET_SIZE) + 1
+    let buckets = distance.div_ceil(CHECKPOINT_BUCKET_SIZE).saturating_add(1);
+    let buckets: u32 = buckets.try_into().unwrap_or(u32::MAX);
+    buckets.min(MAX_BUCKET_SEARCH_ATTEMPTS)
```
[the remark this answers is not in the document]
creditcoinprotoclaw: Good catch — fixed in b56056f7d. The downward walks (`find_highest_attested_before` and `is_height_attested`) walk from `target_height` toward genesis, so the bound is now derived from `target_height` (the real down-span), not `abs_diff(target_height, last_checkpoint_height)`. `bucket_search_attempts` was refactored to take the directional `span_blocks`: downward callers pass `target_height`; the upward `find_lowest_attested_after` passes `last_checkpoint_height - target_height`. Added two regression tests where the last checkpoint sits one block above the target (tiny abs_diff) while a valid checkpoint is ~8 buckets below — both now resolve correctly.

precompiles/chain-info/src/lib.rs
```diff
@@ -32,7 +32,51 @@ mod tests;
...
+/// Using `abs_diff(target_height, last_checkpoint_height)` for a *downward* walk is wrong: when the
+/// last checkpoint sits just above the target the distance is tiny, yet the nearest checkpoint
+/// below the target can be many buckets away, so the walk would stop early and false-negative.
+fn bucket_search_attempts(span_blocks: u64) -> u32 {
+    // ceil(span_blocks / CHECKPOINT_BUCKET_SIZE) + 1
+    let buckets = span_blocks
+        .div_ceil(CHECKPOINT_BUCKET_SIZE)
+        .saturating_add(1);
+    let buckets: u32 = buckets.try_into().unwrap_or(u32::MAX);
+    buckets.min(MAX_BUCKET_SEARCH_ATTEMPTS)
```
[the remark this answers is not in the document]
creditcoinprotoclaw: Fixed in 164b13e3e. You're right — clamping to a fixed 4096-bucket ceiling re-introduced the same false negative for targets above ~4_096_000 blocks. Removed the ceiling entirely: `bucket_search_attempts` now returns the full real span (`ceil(span/CHECKPOINT_BUCKET_SIZE)+1`, saturating only at `u32::MAX` to guard the loop-counter type). The walk is bounded by **gas** instead — every bucket iteration charges `GAS_STORAGE_LOOKUP`, so a pathologically deep walk exhausts gas and reverts (an honest failure) rather than silently returning "not attested". Added a regression test with a checkpoint 4097 buckets below the target (beyond the old ceiling) that is now found.

pallets/attestation/src/lib.rs
```diff
@@ -1308,14 +1315,22 @@ pub mod pallet {
...
+        /// Seeds the election with the real on-chain randomness for `epoch` (via
+        /// [`Config::RandomnessProvider`]), matching the scheduled election path
+        /// ([`OnRandomnessUpdate::on_new_epoch_randomness`]) instead of passing a hardcoded
+        /// zero seed. The current selection logic does not subsample by randomness, but threading
+        /// the real seed keeps forced and scheduled elections consistent and avoids advertising a
+        /// misleading all-zero randomness value.
         #[pallet::call_index(25)]
         #[pallet::weight(<T as Config>::WeightInfo::force_election())]
         pub fn force_election(origin: OriginFor<T>, epoch: u64) -> DispatchResult {
+            use randomness_primitives::provider::RandomnessPalletProvider;
```
atodorov: Why not import at top level ?

pallets/attestation/src/tests.rs
```diff
@@ -7591,6 +7591,38 @@ fn force_election_should_emit_forced_election_event() {
     })
 }
 
+// Regression for #26: `force_election` used to pass a hardcoded `[0; 32]` seed. It must now
+// thread the real on-chain randomness for the requested epoch via `Config::RandomnessProvider`.
+#[test]
+fn force_election_threads_real_epoch_randomness() {
```
atodorov: This test scenario is a bit useless as written b/c it compares the actual value to a pre-defined seed or against a hard-coded value.

pallets/attestation/src/tests.rs
```diff
@@ -7591,6 +7591,38 @@ fn force_election_should_emit_forced_election_event() {
...
+        // The provider the pallet uses for `force_election` exposes the seeded randomness,
+        // proving the wired source is real (not the old hardcoded zero seed).
+        assert_eq!(
+            <Test as Config>::RandomnessProvider::randomness_by_epoch_id(epoch),
+            seed
+        );
+        assert_ne!(
+            <Test as Config>::RandomnessProvider::randomness_by_epoch_id(epoch),
+            [0u8; 32]
+        );
```
atodorov: A better scenario is to iterate over several epoch and compare that the value returned by randomness_by_epoch_id() is different for each epoch.

Calling force_election() extrinsic is independent of checking the randomness seed value.

atodorov (changes requested):

pallets/attestation/src/lib.rs
creditcoinprotoclaw: Done in d2e172990 — moved `use randomness_primitives::provider::RandomnessPalletProvider;` to the top of the pallet module alongside the other `use`s.

pallets/attestation/src/tests.rs
creditcoinprotoclaw: Agreed — reworked in d2e172990. Dropped the hardcoded-seed comparison. The new `randomness_provider_returns_distinct_value_per_epoch` seeds several epochs and asserts `randomness_by_epoch_id()` returns a distinct, non-zero value for each (collision => fail), which actually exercises that the source is real per-epoch randomness.

pallets/attestation/src/tests.rs
creditcoinprotoclaw: Implemented exactly this in d2e172990: `randomness_provider_returns_distinct_value_per_epoch` iterates epochs [3..=7] and asserts each `randomness_by_epoch_id()` is distinct + non-zero, fully independent of the extrinsic. Kept `force_election_consumes_real_epoch_randomness` as a separate, minimal check that the extrinsic still succeeds while consuming the seed.

atodorov (dismissed):

didac-gluwa (dismissed):

DylanVerstraete:
closing in favor of https://github.com/gluwa/creditcoin3/pull/1280

PR #1120, opened 2026-07-01
fix(indexer): remove Attestors entity on unregister to sync GraphQL with chain

creditcoinprotoclaw:
## Problem

`unregisterAttestor` fully purges the attestor from chain storage via `Attestors::<T>::remove(chain_key, &attestor_id)` (`pallets/attestation/src/impls.rs:361`), but the indexer's `handleEventAttestorUnregistered` kept the `Attestors` entity and only flipped its `status` to `Idle`. As a result, unregistered attestors still appeared in the `attestors` GraphQL list, diverging from actual chain state.

## Fix

We cannot simply delete the `Attestors` entity: historical `MapAttestationAttestor` rows reference it via a **non-nullable** `attestor: Attestors!` relation, so removing the row would leave those maps with an unresolvable foreign key (and re-registration mints a new id, so old maps never re-link).

Instead, `handleEventAttestorUnregistered` now sets a new indexer-only `unregistered` status (`4`) on the `Attestors` entity (and bumps `lastUpdateBlockNumber`). This:

- keeps referential integrity for historical `MapAttestationAttestor` relations, and
- lets consumers filter out `status = 4` so the `attestors` list matches chain state.

The `AttestorUnregistered` history rows are still saved unchanged.

The integration test now asserts the entity is retained with `status = 4` after unregister.

### DIFF
--- a/cc3-indexer/src/mappings/mappingHandlers.ts
+++ b/cc3-indexer/src/mappings/mappingHandlers.ts
@@ -285,11 +285,10 @@ export async function handleEventAttestorUnregistered(event: SubstrateEvent): Pr
 
     const id = `${blockNumber}-${event.idx}`;
     const attestorEntity = await checkAndGetAttestor(id, attestor.toString(), chainKeyNumber);
-    attestorEntity.lastUpdateBlockNumber = blockNumber;
-    // Unregistering an attestor removes it from storage on chain, here we just set it to Idle status to keep history of the attestor in the db.
-    attestorEntity.status = RuntimeAttestorStatus.idle;
-
-    await Promise.all([attestorUnregistered.save(), attestorEntity.save()]);
+    // Unregistering an attestor removes it from storage on chain (Attestors::<T>::remove), so we
+    // delete the Attestors entity here to keep the GraphQL list in sync with chain state. The
+    // unregistration history is preserved separately in the AttestorUnregistered entity.
+    await Promise.all([attestorUnregistered.save(), Attestors.remove(attestorEntity.id)]);
 }
 
 export async function handleEventInvulnerableRegistered(event: SubstrateEvent): Promise<void> {
--- a/cc3-indexer/src/mappings/mappingHandlers.ts
+++ b/cc3-indexer/src/mappings/mappingHandlers.ts
@@ -44,6 +44,11 @@ const RuntimeAttestorStatus = {
     idle: 1,
     waiting: 2,
     leaving: 3,
+    // Indexer-only status: the attestor has been unregistered and purged from chain storage
+    // (Attestors::<T>::remove). We keep the Attestors entity so historical MapAttestationAttestor
+    // rows can still resolve their non-nullable `attestor` relation, but flag it as unregistered so
+    // the GraphQL `attestors` list can filter it out to stay in sync with chain state.
+    unregistered: 4,
 } as const;
 
 export async function handleEventAttestorsElected(event: SubstrateEvent): Promise<void> {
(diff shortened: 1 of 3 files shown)


cli/src/test/cc3-indexer-tests/handleEventAttestorUnregistered.test.ts
```diff
@@ -99,23 +99,23 @@ describe('handleEventAttestorUnregistered()', () => {
...
 
-        it('graphQL returns known Attestor entity', async () => {
+        it('graphQL no longer returns the unregistered Attestor entity', async () => {
+            // Unregistering removes the attestor from chain storage, so the indexer deletes the
+            // Attestors entity to keep the GraphQL list in sync with chain state.
             const response = await graphQLQuery(
                 `query { attestors(orderBy: LAST_UPDATE_BLOCK_NUMBER_ASC, last: 10) { nodes { id, attestorId, stashId, chainKey, lastUpdateBlockNumber, status, blsPublicKey } } }`,
             );
             expect(response.data.attestors.nodes).toBeTruthy();
-            expect(response.data.attestors.nodes.length).toBeGreaterThanOrEqual(1);
```
atodorov, line 107: Assert length == 0 and then the rest of this test becomes unnecessary.

atodorov (changes requested):

cc3-indexer/src/mappings/mappingHandlers.ts
```diff
@@ -285,11 +285,10 @@ export async function handleEventAttestorUnregistered(event: SubstrateEvent): Pr
...
     const attestorEntity = await checkAndGetAttestor(id, attestor.toString(), chainKeyNumber);
-    attestorEntity.lastUpdateBlockNumber = blockNumber;
-    // Unregistering an attestor removes it from storage on chain, here we just set it to Idle status to keep history of the attestor in the db.
-    attestorEntity.status = RuntimeAttestorStatus.idle;
-
-    await Promise.all([attestorUnregistered.save(), attestorEntity.save()]);
+    // Unregistering an attestor removes it from storage on chain (Attestors::<T>::remove), so we
+    // delete the Attestors entity here to keep the GraphQL list in sync with chain state. The
+    // unregistration history is preserved separately in the AttestorUnregistered entity.
+    await Promise.all([attestorUnregistered.save(), Attestors.remove(attestorEntity.id)]);
```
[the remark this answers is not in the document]
creditcoinprotoclaw: Good catch. Deleting the `Attestors` row broke the non-nullable `MapAttestationAttestor.attestor` FK for historical maps (and re-registration minted a new id, so old maps never re-linked).

Fixed in 5f1ed2c: instead of removing the entity, we now set a new indexer-only `unregistered` status (`4`) and keep the row. This preserves referential integrity for historical attestation maps while letting consumers filter out `status = 4` so the `attestors` list still matches chain state.

cli/src/test/cc3-indexer-tests/handleEventAttestorUnregistered.test.ts
creditcoinprotoclaw: Simplified in 5f1ed2c. Note the semantics changed slightly per the cursor bot feedback: we now retain the entity flagged with `status = 4` (unregistered) instead of deleting it, so the test filters for the attestor and asserts `length === 1` with `status === 4`, dropping the old loop.

cli/src/test/cc3-indexer-tests/handleEventAttestorUnregistered.test.ts
```diff
@@ -99,23 +99,20 @@ describe('handleEventAttestorUnregistered()', () => {
...
-                if (node.attestorId === attestor.address) {
-                    foundMatch = true;
-                    expect(node.status).toEqual(1); // State after unregistration is Idle
-                }
-            }
-            expect(foundMatch).toEqual(true);
+            const matches = response.data.attestors.nodes.filter((node: any) => node.attestorId === attestor.address);
+            expect(matches.length).toEqual(1);
+            expect(matches[0].status).toEqual(4); // unregistered
+            expect(BigInt(matches[0].lastUpdateBlockNumber)).toBeGreaterThan(0n);
```
atodorov: Take a note of the last block before executing the extrinsic and compare with it. See the `startingBlock` variable in some of the other tests in this directory.

cli/src/test/cc3-indexer-tests/handleEventAttestorUnregistered.test.ts
```diff
@@ -99,23 +99,20 @@ describe('handleEventAttestorUnregistered()', () => {
...
-            for (const node of response.data.attestors.nodes) {
-                expect(BigInt(node.lastUpdateBlockNumber)).toBeGreaterThan(0n);
-
-                if (node.attestorId === attestor.address) {
-                    foundMatch = true;
-                    expect(node.status).toEqual(1); // State after unregistration is Idle
-                }
-            }
-            expect(foundMatch).toEqual(true);
+            const matches = response.data.attestors.nodes.filter((node: any) => node.attestorId === attestor.address);
```
atodorov: nitpick: you can actually insert this condition in the graphQL query above, lots of examples in other tests how that is done, and only assert the response returns 1 node.

atodorov (changes requested):
Overall looks good but a few more tests improvements can be made.

cli/src/test/cc3-indexer-tests/handleEventAttestorUnregistered.test.ts
creditcoinprotoclaw: Done in e701a72. Moved the `attestorId` filter into the GraphQL query and now assert the response returns a single node.

cli/src/test/cc3-indexer-tests/handleEventAttestorUnregistered.test.ts
creditcoinprotoclaw: Done in e701a72. Capture `startingBlock` (via `getChainStatus`) right before the unregister extrinsic and assert `lastUpdateBlockNumber > startingBlock`, matching the pattern in the other tests here.

atodorov (approved):

DylanVerstraete (approved):

PR #1136, opened 2026-07-07
chore(runtime): prune executed migrations, keep pending randomness migration (testnet, spec 128)

creditcoinprotoclaw:
## What

Applies commit `c6d34d3467a31bb8a2525ca82d764871e4732a52` onto the **3.127 testnet line** and bumps `spec_version` **127 → 128**.

This prunes migrations that have already executed on-chain, keeping only the one still pending on mainnet.

## Changes

- **`SingleBlockMigrations` now contains only** `MigrateRandomnessByEpochIndexV0ToV1`.
  - On **testnet** Randomness storage is already at **version 1**, so this migration is a **no-op** here.
  - It remains registered because it is still **pending on mainnet** (storage version 0). Remove once mainnet runs a runtime that includes it.
- **Dropped attestation migrations** (`MigrateAttestationContinuityProofV0ToV1`, `MigrateAttestorsCountV1ToV2`) from the tuple: on-chain storage version is **2** on devnet/testnet/mainnet, so they executed everywhere.
- **Deleted** the `v1_init_supported_chains` and `v1_init_attestation` modules: their data-absence guards are permanently satisfied on every network.
- **Kept** `v1_init_operators` in `migrations.rs` for reference only (unregistered), with the current 4 operator accounts and try-runtime pre/post checks.

## Notes

- Files touched: `runtime/src/lib.rs`, `runtime/src/migrations.rs`, `runtime/src/version.rs` only.
- The applied end-state is **byte-identical** to the reviewed commit `c6d34d3` for `lib.rs` + `migrations.rs` (re-applied manually because the original was authored on the 3.128 base and does not cherry-pick cleanly onto 127).
- `cargo fmt` clean. Full runtime clippy/build could not complete in the constrained build env (OOM) — relying on CI (pinned rust 1.88.0, `-D warnings -A clippy::manual_inspect`) as the authoritative gate.
- **try-runtime**: no-op on testnet (Randomness already v1). The migration only *runs* on mainnet, so the try-runtime dry-run gate applies to the future mainnet PR, not this one.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ edition = "2021"
 license = "Unlicense"
 publish = false
 repository = "https://github.com/gluwa/creditcoin3-next/"
-version = "3.127.0"
+version = "3.128.0"
 
 [workspace.dependencies]
 arc-swap = "1.9.1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -944,7 +944,7 @@ dependencies = [
 
 [[package]]
 name = "archiver"
-version = "3.127.0"
+version = "3.128.0"
 dependencies = [
  "anyhow",
  "axum",
@@ -1538,7 +1538,7 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "attestation_pool"
-version = "3.127.0"
+version = "3.128.0"
 dependencies = [
  "assert_matches",
  "attestor-primitives",
@@ -1559,7 +1559,7 @@ dependencies = [
 
 [[package]]
 name = "attestor"
-version = "3.127.0"
+version = "3.128.0"
 dependencies = [
  "alloy",
  "anyhow",
@@ -1603,7 +1603,7 @@ dependencies = [
 
 [[package]]
 name = "attestor-primitives"
-version = "3.127.0"
+version = "3.128.0"
 dependencies = [
  "bls-signatures",
  "parity-scale-codec",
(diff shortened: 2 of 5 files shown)


atodorov (changes requested):
https://github.com/gluwa/creditcoin3/actions/runs/28871190163/job/85634102337?pr=1136 needs to be fixed. 

Ideally all version bump is in a separate commit of its own.

Cargo.toml
```diff
@@ -47,7 +47,7 @@ edition = "2021"
 license = "Unlicense"
 publish = false
 repository = "https://github.com/gluwa/creditcoin3-next/"
-version = "3.127.0"
+version = "3.128.0"
```
atodorov, line 50: Can you bring the version.rs update into this same commit too? It would be less succeptible to conflicts that way.

atodorov (changes requested):

Cargo.toml
beqaabu: hmm I cherry picked it from usc-dev, weird, lemme check

beqaabu: okay should be fixed now

atodorov: > hmm I cherry picked it from usc-dev, weird, lemme check

Not so weird. Originally you had the version.rs change in the 1st commit so the cherry-pick took all from usc-dev, then it saw that the chunk relevant to version.rs had already been applied + that didn't cause any conflicts so it happily applied the other 2 chunks of the original commit.

DylanVerstraete (approved):

atodorov (approved):

PR #1187, opened 2026-07-22
fix(cli): migrate test suite to ESM for execa 10 compat

creditcoinprotoclaw:
Stacked on top of #1177 (the `execa 5.1.1 → 10.0.0` bump).

## Why
execa v6+ is **pure ESM** and removed `commandSync`. The CLI test suite runs under `ts-jest` in **CommonJS** mode, so with execa 10 the `check-format / cli` job dies at `tsc`:

```
error TS2305: Module '"execa"' has no exported member 'commandSync'.
error TS2349: This expression is not callable. (import execa = require('execa'))
```

…and even after fixing the types, jest would fail at **runtime** with `Cannot use import statement outside a module` because ts-jest emits `require()` for a pure-ESM package. So this needs a full jest ESM migration, not just an API swap.

## What changed
**execa API (v5 → v10), test-only:**
- `commandSync(\`node x\`)` → `execaSync('node', parseCommandString('x'))`
- `import execa = require('execa')` → `import { execa } from 'execa'`
- raw-fd stdio `{ stdout: fs.openSync(path) }` → v10 file-redirect `{ stdout: { file: path } }`
- `killall` template → explicit argv form

**Jest ESM migration:**
- All jest configs (`jest.config.json` + 5 `*-tests.config.ts`) → `ts-jest/presets/default-esm` with `useESM` + `transformIgnorePatterns` allowing execa's ESM dep tree
- New `tsconfig.test.json` (`module: esnext`, `moduleResolution: bundler`) so tests get `import.meta` / import attributes while the CLI build stays `commonjs`
- `NODE_OPTIONS=--experimental-vm-modules` added to all `test:*` scripts

**CommonJS-only patterns that break under ESM:**
- `import x = require('*.json')` (7 test files + `src/lib/attestor/precompile.ts`) → ESM JSON imports
- `import os/path/timeDelta = require(...)` → default/named ESM imports
- `__dirname` in `blockchain-tests/helpers.ts` → `import.meta.url` (renamed to avoid `no-underscore-dangle`)
- Excluded `src/test` from the CommonJS CLI build; pointed ESLint at `tsconfig.test.json` so type-aware linting still covers tests

## Verification
- `yarn build` ✓ (CLI stays CommonJS)
- `yarn lint` ✓
- `yarn check-format` ✓ (the originally failing job)
- Empirically confirmed under the new ESM jest config: execa 10 imports cleanly and `execaSync` spawns a real subprocess (no more `Cannot use import` / `exports is not defined`). Full integration runs still need a `creditcoin3-node` binary, which isn't available in my sandbox.

**CLI runtime behavior is unchanged** — execa is only used in tests.

### DIFF
--- a/cli/src/test/blockchain-tests.config.ts
+++ b/cli/src/test/blockchain-tests.config.ts
@@ -1,9 +1,17 @@
 import type { Config } from '@jest/types';
 const config: Config.InitialOptions = {
-    preset: 'ts-jest',
+    preset: 'ts-jest/presets/default-esm',
     testEnvironment: 'node',
     testTimeout: 240000,
     globalSetup: process.env.BLOCKCHAIN_TESTS_GLOBAL_SETUP || './blockchainSetup.ts',
+    extensionsToTreatAsEsm: ['.ts'],
+    transform: {
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        '^.+\\.tsx?$': ['ts-jest', { useESM: true, tsconfig: 'tsconfig.test.json' }],
+    },
+    transformIgnorePatterns: [
+        'node_modules/(?!(execa|@sindresorhus/merge-streams|figures|get-stream|human-signals|is-plain-obj|is-stream|npm-run-path|path-key|pretty-ms|signal-exit|strip-final-newline|which-command|yoctocolors|unicorn-magic)/)',
+    ],
 };
 
 export default config;
--- a/cli/src/test/utils.ts
+++ b/cli/src/test/utils.ts
@@ -1,16 +1,8 @@
-// eslint-disable-next-line @typescript-eslint/no-require-imports
-import execa = require('execa');
+import os from 'os';
 
-// eslint-disable-next-line @typescript-eslint/no-require-imports
-import fs = require('fs');
+import path from 'path';
 
-// eslint-disable-next-line @typescript-eslint/no-require-imports
-import os = require('os');
-
-// eslint-disable-next-line @typescript-eslint/no-require-imports
-import path = require('path');
-
-import { commandSync } from 'execa';
+import { execa, execaSync, parseCommandString } from 'execa';
 
 import type { EventRecord, Balance, DispatchError } from '../lib';
 import { ApiPromise, expectNoEventError, expectNoDispatchError, newApi } from '../lib';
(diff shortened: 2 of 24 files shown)


cli/src/test/utils.ts
```diff
@@ -107,16 +99,14 @@ function runNode(name: string, extraArgs: string) {
...
-    void execa(
-        '../target/release/creditcoin3-node',
-        `--chain dev --validator --pruning archive ${extraArgs}`.split(' '),
-        {
-            detached: true,
-            stdout: fs.openSync(`${logPrefix}.stdout`, 'w'),
-            stderr: fs.openSync(`${logPrefix}.stderr`, 'w'),
-        },
-    );
+    void execa('../target/release/creditcoin3-node', nodeArgs, {
```
atodorov: @creditcoinprotoclaw can you inline the call to parseCommandString() in here ?

cli/src/test/utils.ts
```diff
@@ -140,7 +130,7 @@ export async function startAliceAndBob() {
 export function killCreditcoinNodes() {
     console.log('INFO: killing all creditcoin3-node processes');
 
-    commandSync(`killall -9 creditcoin3-node`);
+    execaSync('killall', ['-9', 'creditcoin3-node']);
```
atodorov: Why no call to parseCommand string ?

atodorov (changes requested):

cli/src/test/utils.ts
creditcoinprotoclaw: Done in fe479c4c4 — inlined `parseCommandString()` directly into the `execa()` call.

cli/src/test/utils.ts
creditcoinprotoclaw: Good catch — for consistency I now route it through `parseCommandString("killall -9 creditcoin3-node")` and spread into `execaSync(command, args)`. Fixed in fe479c4c4.

atodorov (approved):
integration-test-cli CI jobs LGTM:
https://github.com/gluwa/creditcoin3/actions/runs/30451568706?pr=1187

didac-gluwa (approved):

cli/src/test/blockchain-tests.config.ts
```diff
@@ -1,9 +1,17 @@
...
     testTimeout: 240000,
     globalSetup: process.env.BLOCKCHAIN_TESTS_GLOBAL_SETUP || './blockchainSetup.ts',
+    extensionsToTreatAsEsm: ['.ts'],
+    transform: {
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        '^.+\\.tsx?$': ['ts-jest', { useESM: true, tsconfig: 'tsconfig.test.json' }],
+    },
+    transformIgnorePatterns: [
+        'node_modules/(?!(execa|@sindresorhus/merge-streams|figures|get-stream|human-signals|is-plain-obj|is-stream|npm-run-path|path-key|pretty-ms|signal-exit|strip-final-newline|which-command|yoctocolors|unicorn-magic)/)',
+    ],
```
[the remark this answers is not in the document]
atodorov: blockchain tests seem to be able to start just fine:
https://github.com/gluwa/creditcoin3/actions/runs/30451568706/job/90579576840?pr=1187

DylanVerstraete (approved):

PR #1200, opened 2026-07-27
fix(attestor): resume startup from the finalized anchor, not an attestation lookup [ATTESTOR-V2-009]

DylanVerstraete:
Fixes [ATTESTOR-V2-009](https://github.com/chainlight-io/USC-ATTESTOR-V2-issues/issues/10) — *Checkpoint-Backed `LastDigest` Can Cause Attestor Startup Crash Loop After Revert*.

## The bug

`startup::fetch_start_point` resolved the latest finalized digest back to a row in `Attestations` and treated a miss as impossible state. That assumption breaks whenever the anchor is **checkpoint-backed**, which happens two ways:

* **after `revert_to()`** — `do_revert_to` clears every stored attestation for the chain and repoints *both* `LastCheckpoint` and `LastDigest` at the surviving checkpoint, so `LastDigest` names a digest that has no `Attestations` entry;
* **on a checkpoint-only chain** — `LastDigest` is absent and the lookup falls back to `LastCheckpoint` for the same reason.

Startup then failed **deterministically on every boot**, so a supervisor crash-looped the fleet after an emergency revert until an operator intervened or shipped a patched binary. Already-running attestors were unaffected — this is purely the startup path.

Trigger requires an authorized `revert_to()` plus a restart before the next attestation commits, so severity is availability-only (no integrity impact, no remote trigger). But it bites precisely during incident response, when restarting attestors after a revert is the natural instinct.

## The fix

Resume from the `(height, digest)` **anchor** that `fetch_last_finalized` already returns — it mirrors the runtime’s `LastDigest` → `LastCheckpoint` lookup order. This covers all four states uniformly (attestation-backed, checkpoint-backed after a revert, checkpoint-only, true genesis) and removes both the unreachable checkpoint fallback branch and the `DigestNotFound` error that encoded the bad assumption.

Net effect is a simplification: one RPC round-trip and one branch fewer.

Per the report, merely turning the panic into a recoverable error would **not** have been sufficient — startup would still reject a valid checkpoint-backed state and keep failing on every restart.

## Testing

`startup.rs` had **no tests**, because it took a concrete `Client` (a live subxt connection) and so could only be exercised against a running node. This PR introduces a narrow `StartPointQuery` trait — following the existing `CcRpcProvider` convention — implemented for `Client` as a straight delegation, which makes the resume rules unit-testable and unlocks future coverage of the module.

Four new tests:

| Test | Pins |
|---|---|
| `resumes_from_an_attestation_backed_anchor` | normal path unchanged |
| `resumes_from_a_checkpoint_backed_anchor_after_revert` | **the regression** |
| `reports_no_anchor_when_chain_is_at_genesis` | `None` anchor → genesis |
| `propagates_query_errors` | RPC failures still surface |

The fake deliberately does **not** model whether an anchor is attestation- or checkpoint-backed — that indistinguishability is the invariant, and encoding it would let the bug back in.

Also pins the runtime side of the contract: a `pallet-attestation` test asserting the post-revert state shape (anchor names the checkpoint, `Attestations` empty), cross-referenced to this startup path so a future change to either end surfaces the dependency.

## Verification

- `cargo test -p attestor` — 54 passed (50 pre-existing + 4 new)
- `cargo test -p pallet-attestation` — 262 passed
- `cargo clippy -p attestor -p pallet-attestation --all-targets` — clean
- `cargo fmt --all`, `taplo format` (Cargo.toml touched), `cargo machete` — clean

## Note for reviewers

`common/continuity/src/builder/mod.rs:605-627` uses the same `fetch_last_digest` + `get_attestation_by_digest` pattern. If it also assumes an attestation-backed anchor it would fail the same way after a revert — out of scope here, but worth a follow-up look.


didac-gluwa (dismissed):

attestor/attestor/src/startup.rs
```diff
@@ -244,3 +294,129 @@ pub async fn reconcile_metadata(cc3: &Arc<Client>) -> Result<(), Error> {
...
+        ) -> Result<Option<(Height, Digest)>, Error> {
+            if self.fail_anchor {
+                return Err(Error::MissingMaxCatchup(CHAIN));
+            }
+            Ok(self.anchor)
+        }
+    }
+
+    #[tokio::test]
+    async fn resumes_from_an_attestation_backed_anchor() {
```
BradleyOlson64, line 363: It feels like these tests don't actually check the code that would run in production.

For a tighter implementation, I think we could get rid of these and remove the `StartPointQuery` trait.

The new test in `pallets/attestation/src/tests.rs‎` covers what we actually wanted to check.

creditcoinprotoclaw:
Superseded by #1218 — reopened against a fresh checkout of usc-dev with only the single fix commit (c5d1b8b). @BradleyOlson64's open suggestion (drop the startup.rs tests + StartPointQuery trait in favour of the pallet-attestation test) is carried over and flagged on #1218 so it isn't lost.

PR #1201, opened 2026-07-27
fix(attestation): tag only active-attestor commit_attestation submissions [ATTESTOR-V2-008]

DylanVerstraete:
Fixes [ATTESTOR-V2-008](https://github.com/chainlight-io/USC-ATTESTOR-V2-issues/issues/9) — *Shared `provides` Tag Preemption Can Degrade `commit_attestation` Liveness*.

## The bug

`PrevalidateAttestationCommit` assigned the shared `(chain_key, digest)` `provides` tag to **every** `commit_attestation` call. Everything checked before it is signer-independent (attestor-list bound, continuity-proof size), and the duplicate check is `active_attestors.contains(who) && …` — so a non-active signer fell straight through to the tag.

The digest is derived purely from public chain data (header number, root, prev digest), so anyone can predict what honest attestors will produce next, and authorization is only enforced in dispatch. Any funded account could therefore:

* **reserve** the single pool slot for the next legitimate digest with a transaction guaranteed to fail (`AttestorNotActive`), or
* **evict** an already-pending legitimate submission by supplying higher priority.

Repeating that across successive digests delays attestation progress.

Severity is liveness-only: the invalid transaction never commits anything, and each attempt costs the attacker a fee. No integrity impact.

## The fix

Gate the tag on active-set membership:

```rust
let is_active = active_attestors.contains(who);
…
if !is_active {
    return Ok(ValidTransaction::default());   // admitted, but claims no slot
}
```

Non-active signers are **still admitted** — they fail in dispatch and pay the inclusion fee, which is the property that deters free prevalidation-rejected spam (and is deliberate per the extension’s own docs). They simply contribute no `provides` tag, so they cannot collide with a legitimate attestation’s slot.

Deduplication between *real* racing attestors — the reason the tag exists ([ATTESTOR-V2-999](https://github.com/chainlight-io/USC-ATTESTOR-V2-issues/issues/8)) — is unchanged. Note the fix must **not** be to drop the tag, which would regress that.

## Testing

New `prevalidate_tags_only_active_attestor_submissions`: submits the *same* attestation from an active attestor and from a non-attestor, and asserts

* the active attestor still claims a tag (dedup preserved),
* the non-attestor is still admitted but with an empty `provides`,
* the two sets of tags cannot collide.

Also asserts the empty tag on the existing `prevalidate_attestation_pays_fee_if_not_active_attestor` case.

## Verification

- `cargo test -p pallet-attestation` — 262 passed, 0 failed (4/4 prevalidation tests)
- `cargo clippy -p pallet-attestation --all-targets` — clean
- `cargo fmt --all` — clean

## Related

Companion to #1200 (ATTESTOR-V2-009 startup crash loop). I also checked the third candidate from the same review sweep — `continuity::builder::get_last_attested_block`, which uses the same `fetch_last_digest` + `get_attestation_by_digest` shape — and it is **not** affected in practice: it maps a missing attestation to `Ok(None)` rather than erroring, and has no callers in-tree. Left alone deliberately.


pallets/attestation/src/extensions.rs
```diff
@@ -134,13 +136,28 @@ where
...
+            //
+            // The tag below is derived only from public chain data, so anyone can predict the
+            // digest honest attestors will produce next. Handing a tag to an unauthorized
+            // submitter would let any funded account reserve (or, with higher priority, evict)
+            // the single pool slot for that digest with a transaction guaranteed to fail,
+            // repeatedly delaying attestation progress (ATTESTOR-V2-008). Authorization is only
+            // checked in dispatch, so it must gate the tag here.
+            if !is_active {
+                return Ok(ValidTransaction::default());
+            }
```
didac-gluwa, line 159: I would move this check before the duplication check, that way we can simplify the logic a bit:

```rust
if !active_attestors.contains(who) {
    return Ok(ValidTransaction::default());
}

if Pallet::<T>::check_duplicate(attestation) {
    return Err(TransactionValidityError::Invalid(InvalidTransaction::Stale));
}
```

didac-gluwa (dismissed):

BradleyOlson64 (dismissed):

creditcoinprotoclaw:
Superseded by #1216 — reopened against a fresh checkout of usc-dev with only the single fix commit (2673533), and addressed @didac-gluwa's review (active-set check moved before the duplicate check to simplify the flow).

PR #1205, opened 2026-07-28
fix: install Node.js from official upstream tarball instead of NodeSource

jonathanlaf:
## Summary
- `docker.yml`'s Docker build job was failing on our self-hosted Linode CI runners (`nl-ams`, ephemeral) with `npm: not found`
- Root cause: `deb.nodesource.com`'s GPG signing-key fetch intermittently returns HTTP 403 from that runner pool's IP range (confirmed: identical request succeeds cleanly from an unrelated network; same Ubuntu 26.04 image succeeded minutes earlier from the same CI infra, ruling out an OS-incompatibility). The NodeSource setup script logs the failure but exits 0, so `apt-get install nodejs` silently falls back to Ubuntu's own `universe`-component `nodejs` package, which doesn't bundle `npm` — hence the error downstream.
- Considered switching to Ubuntu's native `nodejs`/`npm` packages instead, but ruled it out: both are in `universe` (best-effort security support, not Canonical's main-archive guarantee), `npm` is stuck at 9.2.0 (~2022, a full major behind), and the version strings are noisier for Docker Scout's CVE matching.
- Fix instead installs the official Node.js release tarball directly from nodejs.org, checksum-verified against their own published `SHASUMS256.txt`, arch-aware via `TARGETARCH` (amd64/arm64).

## Verification
- Verified locally (native arm64, no emulation): `node --version`, `npm --version`, `yarn install` all clean
- Hadolint: clean
- `docker scout sbom`: cleanly detects `node@22.23.1` via binary classification, standard purl (`pkg:github/nodejs/node@22.23.1`), no distro-suffix version string
- Manually triggered the real `Docker` workflow via `workflow_dispatch` on this branch against the actual Linode runner — passed the point where the original run failed (which failed in ~1 min) and proceeded well into the Rust build (25+ min) before being cancelled once the fix was confirmed working
- Confirmed the other 4 Dockerfiles in the repo (`cc3-indexer/Dockerfile`, `cc3-indexer/postgres/Dockerfile`, `scripts/stress-test/Dockerfile`, `scripts/usc-audit-automation/Dockerfile`) don't use NodeSource — only the root `Dockerfile` was affected

## Test plan
- [ ] Full Docker CI run (triggered by this PR) completes successfully on the Linode runner

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,26 @@
 # hadolint global ignore=DL3008,DL3009,DL3013,DL3016,SC3046,DL4006,SC1091,SC2086
 FROM ubuntu:26.04@sha256:b7f48194d4d8b763a478a621cdc81c27be222ba2206ca3ca6bc42b49685f3d9e AS runtime-base
 ENV DEBIAN_FRONTEND=noninteractive
+ARG TARGETARCH
+ARG NODE_VERSION=22.23.1
+# NodeSource's apt key fetch (deb.nodesource.com) intermittently 403s from our
+# self-hosted Linode CI runners' shared IP pool. Install the official upstream
+# release directly instead, checksum-verified against nodejs.org's own SHASUMS256.txt.
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install -y --no-install-recommends ca-certificates curl && \
+    apt-get install -y --no-install-recommends ca-certificates curl xz-utils libdw1t64 libpq5 && \
     update-ca-certificates && \
-    curl -fsSL https://deb.nodesource.com/setup_20.x | bash - && \
-    apt-get install -y libdw1t64 libpq5 nodejs --no-install-recommends && \
+    case "${TARGETARCH}" in \
+      amd64) NODE_ARCH=x64;    NODE_SHA256=9749e988f437343b7fa832c69ded82a312e41a03116d766797ac14f6f9eee578 ;; \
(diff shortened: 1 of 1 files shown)


dlebee (approved):

atodorov:
@dlebee , @jonathanlaf who's going to be bringing this change downstream (and how) ? 

I've got #1210 built on top of it, but atm keeping it on hold to avoid making unnecessary changes until things settle down.

dlebee:
> @dlebee , @jonathanlaf who's going to be bringing this change downstream (and how) ? 
> 
> 
> 
> I've got #1210 built on top of it, but atm keeping it on hold to avoid making unnecessary changes until things settle down. 

I brought this into usc-dev when correcting history I believe, unless the docker file was re-overwritten either changes in dev since I'll check for you.

Okay I'll make a PR to dev today for this, I overwrote it because docker file had changes since in or just wanted to keep dev as is to avoid problems.

atodorov:
> > @dlebee , @jonathanlaf who's going to be bringing this change downstream (and how) ?
> > I've got #1210 built on top of it, but atm keeping it on hold to avoid making unnecessary changes until things settle down.
> 
> I brought this into usc-dev when correcting history I believe, unless the docker file was re-overwritten either changes in dev since I'll check for you.
> 

Not in usc-dev, see https://github.com/gluwa/creditcoin3/blob/usc-dev/Dockerfile#L8

PR #1207, opened 2026-07-28
precompiles: make metadata JSON chain output (devnet/testnet/mainnet) flag-driven instead of hand-edited

dlebee:
## Problem

`precompiles/metadata/generate-metadata-json.sh` hardcodes which `precompiles-creditcoin3-*.json` files it writes. Across branches/merges this has been hand-edited back and forth instead of being driven by anything dynamic:

- The `output_mainnet` block (3 lines: variable, write, log message) has been added and removed **at least 3 separate times** purely as a side effect of merges between branches (e.g. `94b95e98` adds it, `61d035eb` still has it, `2704d4f9`/`f063470e` remove it again).
- Right now `main`/`usc-testnet` only ever write `devnet`+`testnet`, while `usc-dev` writes all three - i.e. the *same script* behaves differently depending on which branch you're on, with no flag or parameter driving that, just drift.
- This already caused **real, silent staleness**: on `main`, `precompiles-creditcoin3-mainnet.json`'s last content update is `61d035eb`, while `testnet`'s is a later commit (`5f71fe2a`) - nothing since has kept it in sync, because the script no longer writes it and CI's "metadata is up to date" check in `sanity.yml` only diffs `devnet`/`testnet`. Meanwhile `check-solidity-source-vs-metadata.sh` still trusts `precompiles-creditcoin3-mainnet.json` as ground truth whenever a PR targets `main` (`TARGET_CHAIN=mainnet`), with no automated regeneration keeping it honest in between.

## Fix

- `generate-metadata-json.sh` now accepts `--devnet/--no-devnet`, `--testnet/--no-testnet`, `--mainnet/--no-mainnet`, and `--all`, defaulting to **devnet=on, testnet=on, mainnet=off**. No more editing the script body to change which files get written.
- `sanity.yml`'s "Check metadata JSON files are up to date" step now passes `--mainnet` (and checks the mainnet file) **only when `github.base_ref == 'main'`** - reusing the exact same signal that already drives the `Prepare ENV for Devnet/Testnet/Mainnet` steps a few lines down in that same job, instead of introducing yet another bespoke branch check.
- Updated `precompiles/metadata/README.md` to document the flags and the new CI behavior.

## Why mainnet defaults to off

Devnet/testnet metadata should stay in sync on every PR regardless of target branch (cheap, and both already share byte-identical content since the script writes the same `$json_array` to both). Mainnet metadata should only be refreshed once a change is actually being promoted towards `main` - generating/checking it on every `usc-dev`/`usc-testnet` PR would either force premature mainnet-metadata churn or (as happened here) get silently disabled instead.

## Verification

- `bash -n` syntax-checked the updated script.
- Flag-parsing logic (defaults, `--mainnet`, `--all`, `--no-devnet --mainnet`, all-off error, unknown-flag error) unit-tested in isolation via Git Bash - all behave as intended.
- Could not run the full script end-to-end in this environment (`jq`/`solc` aren't installed here), so the ABI/JSON-generation logic itself is unchanged from the current script and wasn't otherwise touched.

## Open questions for reviewer

- Is "only refresh mainnet metadata on PRs targeting `main`" actually the right trigger, or should this instead be tied to cutting an actual mainnet release/tag? Right now `github.base_ref == 'main'` matches the existing pattern used for `TARGET_CHAIN` elsewhere in this same job, but happy to change it if there's a better signal.
- Should `precompiles-creditcoin3-mainnet.json` be regenerated fresh in this PR to fix the current drift, or handled as a separate follow-up once this mechanism is agreed on?

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -118,12 +118,22 @@ jobs:
       - name: Check metadata JSON files are up to date
         working-directory: ./precompiles/metadata
         run: |
-          ./generate-metadata-json.sh
+          # Mainnet metadata is only regenerated/verified for PRs targeting main: that's the
+          # only time it's "right" to refresh it, since devnet/testnet move independently of
+          # mainnet the rest of the time.
+          generate_args=(--devnet --testnet)
+          files_to_check=(precompiles-creditcoin3-devnet.json precompiles-creditcoin3-testnet.json)
+          if [ "${{ github.base_ref }}" == "main" ]; then
+            generate_args+=(--mainnet)
+            files_to_check+=(precompiles-creditcoin3-mainnet.json)
+          fi
+
+          ./generate-metadata-json.sh "${generate_args[@]}"
           cd ../..
           # Check if metadata JSON files were modified
-          if [ -n "$(git status --short --untracked-files=no precompiles/metadata/precompiles-creditcoin3-devnet.json precompiles/metadata/precompiles-creditcoin3-testnet.json)" ]; then
-            echo "ERROR: Metadata JSON files are out of date. Please run './generate-metadata-json.sh' and commit the changes."
-            git diff precompiles/metadata/precompiles-creditcoin3-devnet.json precompiles/metadata/precompiles-creditcoin3-testnet.json
+          if [ -n "$(git status --short --untracked-files=no "${files_to_check[@]/#/precompiles/metadata/}")" ]; then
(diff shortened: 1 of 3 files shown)


DylanVerstraete:
fix already landed on `usc-dev` https://github.com/gluwa/creditcoin3/commit/94b95e98764c8cbff3f9f552afbce27a9a302441 feel free to close

PR #1208, opened 2026-07-28
cli: pin polkadot.js low-level packages to a single version to fix duplicate-@polkadot/util warning

dlebee:
## Problem

The CLI pipeline prints:

```
@polkadot/util has multiple versions, ensure that there is only one installed.
```

This isn't a lint rule in this repo - it's a runtime `console.warn` baked into `@polkadot/util` itself, which registers a version marker on `globalThis` at import time and complains if it detects a second, differently-versioned copy of itself loaded in the same process. It's accurate here: `cli/yarn.lock` really did have two physically separate installed copies of several polkadot.js low-level packages (confirmed identically present on `main`, so this isn't new/introduced by `new-usc-testnet`):

- `@polkadot/util`: `13.5.8` and `13.5.9`
- `@polkadot/networks`: `13.5.8` and `13.5.9`
- `@polkadot/x-bigint`: `13.5.8` and `13.5.9`
- `@polkadot/x-global`: `13.5.8` and `13.5.9`
- `@polkadot/x-textdecoder`: `13.5.8` and `13.5.9`
- `@polkadot/x-textencoder`: `13.5.8` and `13.5.9`

### Why this happened

Almost the entire `@polkadot/api@16.5.2` family (`api-augment`, `api-base`, `rpc-core`, `types`, `types-codec`, etc.) depends on these packages via a caret range, e.g. `"@polkadot/util": "^13.5.8"`. That's what resolved to the `13.5.8` copies.

But `@polkadot/keyring` (pulled in by `@polkadot/api`/`@polkadot/rpc-provider` via the same `^13.5.8` range) itself resolved to the newer patch `13.5.9`, and - per the standard polkadot.js monorepo convention of exact-pinning sibling packages within the same lockstep release - `keyring@13.5.9`'s own manifest pins `"@polkadot/util": "13.5.9"` (no caret, not a range). Yarn Classic doesn't retroactively collapse an exact pin into an already-resolved caret match (even though `13.5.9` satisfies `^13.5.8`), so it just creates a second, separate physical copy instead. Same story for `networks`/`x-bigint`/`x-global`/`x-textdecoder`/`x-textencoder`, all pulled in transitively by `keyring`/`util-crypto`.

## Fix

Add a `resolutions` block to `cli/package.json` pinning all six affected packages to `13.5.9` (the version already used by `keyring`/`util-crypto`, and compatible with every existing `^13.5.8` requester per semver), then regenerate `cli/yarn.lock`.

## Verification

- Ran `yarn install` in `cli/` and confirmed `cli/yarn.lock` now has exactly one lockfile block per affected package (e.g. `"@polkadot/util@13.5.9", "@polkadot/util@^13.5.8":` collapsed into a single entry, versus two separate blocks before).
- Confirmed on disk that `cli/node_modules/@polkadot/util` (and siblings) now contains a single physical copy at `13.5.9`, with no nested duplicate installed elsewhere.
- `yarn install` did fail afterwards on the `prepare` (`tsc`) lifecycle script in this environment, but that's unrelated to this change and pre-existing on `main` too: `cli/src/test/blockchain-tests/artifacts/*.json` are checked out as literal-text symlink placeholders (containing the string `../../../../../precompiles/metadata/abi/*.json`) rather than real symlinks on this Windows checkout, which `tsc` then fails to parse as JSON. Confirmed this is identical on `main` and not something this change touches or introduces.

## Open question for reviewer

Would you rather pin via `resolutions` (this PR, quick/local fix) or instead bump the whole `@polkadot/api` family to a release where all sibling `@polkadot/*` packages were republished in lockstep at a matching patch version (removes the skew at the source, but is a bigger dependency bump)? Happy to go either way.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -68,5 +68,13 @@
         "js-guid": "^1.0.0",
         "prompts": "^2.4.2",
         "time-delta": "^1.0.0"
+    },
+    "resolutions": {
+        "@polkadot/util": "13.5.9",
+        "@polkadot/networks": "13.5.9",
+        "@polkadot/x-bigint": "13.5.9",
+        "@polkadot/x-global": "13.5.9",
+        "@polkadot/x-textdecoder": "13.5.9",
+        "@polkadot/x-textencoder": "13.5.9"
     }
 }
--- a/cli/package.json
+++ b/cli/package.json
@@ -73,8 +73,10 @@
         "@polkadot/util": "13.5.9",
         "@polkadot/networks": "13.5.9",
         "@polkadot/x-bigint": "13.5.9",
+        "@polkadot/x-fetch": "13.5.9",
         "@polkadot/x-global": "13.5.9",
         "@polkadot/x-textdecoder": "13.5.9",
-        "@polkadot/x-textencoder": "13.5.9"
+        "@polkadot/x-textencoder": "13.5.9",
+        "@polkadot/x-ws": "13.5.9"
     }
 }
--- a/cli/yarn.lock
+++ b/cli/yarn.lock
@@ -1441,16 +1441,16 @@
     "@polkadot/x-global" "13.5.9"
     tslib "^2.8.0"
 
-"@polkadot/x-fetch@^13.5.8":
-  version "13.5.8"
-  resolved "https://registry.yarnpkg.com/@polkadot/x-fetch/-/x-fetch-13.5.8.tgz#aeaccbcb8b3b4dfdbfc2f78d028d71e7da136f89"
-  integrity sha512-htNuY8zFw5vzNS2mFm9P22oBJg7Az8Xbg3fMmR/A6ZDhAfxfM6IbX9pPHkNJY2Wng3tysrY5VMOUMb1IIrSf3w==
+"@polkadot/x-fetch@13.5.9", "@polkadot/x-fetch@^13.5.8":
+  version "13.5.9"
+  resolved "https://registry.yarnpkg.com/@polkadot/x-fetch/-/x-fetch-13.5.9.tgz#396a5a884d13d26a04c61b6e43240ca69fd72190"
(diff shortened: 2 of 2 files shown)


dlebee:
## CI verification

Manually dispatched `unit-test-cli.yml` (`workflow_dispatch`) on both this branch and the unfixed `new-usc-testnet` baseline to compare `yarn install` output directly, since that job's `pull_request` trigger doesn't fire for PRs targeting `new-usc-testnet`:

**Baseline (`new-usc-testnet`, unfixed)** - [run](https://github.com/gluwa/creditcoin3/actions/runs/30325789898) - `yarn install` prints:
```
warning "@polkadot/api > @polkadot/keyring@13.5.9" has incorrect peer dependency "@polkadot/util@13.5.9".
warning "@polkadot/api > @polkadot/util-crypto@13.5.9" has incorrect peer dependency "@polkadot/util@13.5.9".
```
confirming the split-version state described above.

**This branch, 1st attempt** ([run](https://github.com/gluwa/creditcoin3/actions/runs/30325784674)) - the original 6-package `resolutions` block eliminated those peer-dependency warnings, but surfaced a new one:
```
warning Resolution field "@polkadot/x-global@13.5.9" is incompatible with requested version "@polkadot/x-global@13.5.8"
```
Root cause: `@polkadot/x-fetch` and `@polkadot/x-ws` (both still resolved at `13.5.8`, untouched by the original fix) exact-pin their own `@polkadot/x-global` dependency to `13.5.8`, conflicting with the forced `13.5.9`. `13.5.9` exists on npm for both, so I added them to `resolutions` too (latest commit).

**This branch, current** ([run](https://github.com/gluwa/creditcoin3/actions/runs/30326016704)) - zero `@polkadot/*`-related warnings in the full `yarn install` output, and `cli/yarn.lock` now has exactly one lockfile block per `@polkadot/*` package (verified via a script scanning all top-level entries).

DylanVerstraete (approved):

atodorov (changes requested):
FTR CI jobs running in the context of https://github.com/gluwa/creditcoin3/pull/1213. I will update this comment once I have more info!

atodorov (approved):
CI looks good:
https://github.com/gluwa/creditcoin3/actions/runs/30363750853?pr=1213

PR #1214, opened 2026-07-28
fix(DO-2338): grant pull-requests: write so Docker Scout can post PR comments

jonathanlaf:
## Summary
- `docker/scout-action`'s `write-comment` input defaults to `true`, but was silently no-op'ing because this workflow (and its caller `docker.yml`) declared `permissions: read-all` — no `pull-requests` write scope, and GitHub Actions permission ceilings can only be narrowed at job level, never broadened
- Granted `pull-requests: write` in both `docker.yml` and `build-docker.yml`, narrowing everything else to just `contents: read` (all these jobs actually use)
- Added `keep-previous-comments: true` — without it, the action's default "one comment per job" behavior means the 3rd Scout scan (proof-gen-stress-test) silently overwrites the first two images' (main, indexer) results in the same PR comment

Part of DO-2336 (org-wide Docker Scout CVE-scan rework — surface results as PR comments instead of only in the raw log).

## Test plan
- [ ] This PR's own `docker-build` workflow run posts 3 separate Scout-result comments (main, indexer, proof-gen-stress-test), not one overwriting the others

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -17,6 +17,15 @@ name: Reusable workflow to build docker image
       GLUWA_BOT_PR:
         description: gluwa-bot PAT (pull-requests:write only) so Docker Scout can post PR comments — GITHUB_TOKEN's own permissions couldn't be elevated (see DO-2338)
         required: false
+      # gluwabot DockerHub token, read-only scope — lets Scout authenticate to the Scout
+      # backend even for a local:// image scan. Deliberately separate from
+      # DOCKER_PUSH_USERNAME/PASSWORD below: those are push-capable, and PR builds run
+      # untrusted Dockerfile/build-script content from the PR, so they must never see a
+      # credential that can push to the real image (see DO-2338).
+      DOCKERHUB_SCOUT_USERNAME:
+        required: false
+      DOCKERHUB_SCOUT_TOKEN:
+        required: false
       # Pre-existing secrets this workflow already consumed before DO-2338 — now need
       # explicit declaration since adding any `secrets:` block switches this workflow_call
       # into strict mode. Empty/unset on PR-triggered docker.yml (which never inherits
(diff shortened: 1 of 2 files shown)


.github/workflows/build-docker.yml
```diff
@@ -13,7 +13,32 @@ name: Reusable workflow to build docker image
...
+      DOCKERHUB_SCOUT_TOKEN:
+        required: false
+      # Pre-existing secrets this workflow already consumed before DO-2338 — now need
+      # explicit declaration since adding any `secrets:` block switches this workflow_call
+      # into strict mode. Empty/unset on PR-triggered docker.yml (which never inherits
+      # them, deliberately); populated on release.yml via secrets: inherit.
+      DOCKER_PUSH_USERNAME:
+        required: false
+      DOCKER_PUSH_PASSWORD:
+        required: false
```
atodorov, line 36: Is this continue to work when this reusable workflow is called with `secrets: inherit` or do we need to pass them explicitly ? 

I have no experience with how this works with the secrets section defined inside the reusable workflow and the only way I know how to test it is to build & push a DEBUG release.

atodorov (changes requested):
Asking for some clarification. Marked as "Request changes" so that we don't merge this accidentally before its clear it's going to work!

atodorov:
@jonathanlaf please feel free to request reviews from `gluwa/protocol` too. I've requested it this time to avoid unnecessary waiting.

.github/workflows/build-docker.yml
jonathanlaf: @atodorov Confirmed — no DEBUG release needed to check this.

`release.yml:200-205` already calls `build-docker.yml` with `secrets: inherit`, while `docker.yml:42-45` passes `GLUWA_BOT_PR`/`DOCKERHUB_SCOUT_USERNAME`/`DOCKERHUB_SCOUT_TOKEN` explicitly. Per GitHub's documented behavior, `secrets: inherit` forwards every secret the caller can see (org/repo/environment) into the reusable workflow, matched by name against its `workflow_call.secrets` declarations — regardless of whether they're `required: true` or `false`. So as long as those three secrets exist with those exact names at the repo/org level, `release.yml`'s inherit call will pick them up automatically, no extra wiring needed.

Also worth flagging: Bugbot's "permissions change never applied" finding above was against an earlier commit (`50c1f96`) — current HEAD (`e1547a69`) no longer tries to elevate `GITHUB_TOKEN` via `pull-requests: write` at all. Both workflows kept `permissions: read-all`, and Scout's PR comment now authenticates via a dedicated `GLUWA_BOT_PR` PAT passed as `docker/scout-action`'s `github-token`, which sidesteps the reusable-workflow permission-ceiling issue Bugbot flagged.

atodorov (approved):

BradleyOlson64:
Looks like this pr needs a rebase on top of the latest `usc-dev`

BradleyOlson64 (approved):

didac-gluwa (approved):

beqaabu (approved):

PR #1215, opened 2026-07-29
feat: writeability (rebased onto new usc-dev)

DylanVerstraete:
## What

USC **write-ability**: cross-chain messaging from Creditcoin L1 to destination chains, fees paid on the source chain. This PR carries the full feature onto `usc-dev`, aligned with the merged usc-contracts stack (`feature/SMC-1769`, PR gluwa/usc-contracts#23).

### Components

| Area | What's here |
|---|---|
| **pallet-supported-chains** | `WriteAbilityConfigs` / `OutboxFactories` / `CoreFees` storage + operator extrinsics (`set_write_ability_config`, `set_outbox_factory_addr`, `set_core_fee`); genesis seeding for chains + write-ability configs; storage migrations **V0→V1** (new maps) and **V1→V2** (core-fee `token` field removed) |
| **chain-info precompile** (`0x…0FD3`) | `get_outbox_factory_address(uint64)`, and `get_core_fee(uint32) → uint256` matching usc-contracts' `ICoreFeeProvider` **exactly** — selector `0x5b023376` is pinned by a unit test so the two repos cannot drift silently |
| **attestor** (`tasks/write_ability`) | Message attestation: resolves the Outbox on-chain (chain-info → factory `OutboxCreated` scan, finalized-head bounded), listens for `MessagePublished`, signs ECDSA message votes, gossips them on the existing p2p swarm (dedicated topics), durable scan cursor with lookback rewind, reobservation liveness, on-chain `EOAValidator` set/threshold hot-reload + set-update vote proposals |
| **common/write-ability** | Shared vote envelope, `messageHash` computation (golden-vector tested — the cross-repo signing contract with `gluwa/usc-message-relayer`), and the attestor's ABI mirror (trimmed to exactly the surface the attestor calls) |
| **cc3-indexer** | Dynamic Outbox discovery (`OutboxCreated` by topic, chain-wide, with an unauthenticated-datasource DoS cap), `OutboxMessage` entities with ack tracking, `OutboxFactoryRegistered` handler — all four handlers integration-tested |
| **write-ability-e2e** | Full-stack workflow: node + attestors + relayer + proof-gen + fee stack; publish → attest → deliver → ack round-trip incl. `claimDelivery` payout |
| **usc-messaging** | EIP-712 quoter service, e2e scripts, demo dApp. `scripts/deploy.ts` is flagged **legacy** (see Known follow-ups) |

## Compatibility contract

- **usc-contracts:** `main` at/after the #23 merge (`1f8b553`). Aligned surfaces: `get_core_fee(uint32)→uint256` (`0x5b023376`), `MessagePublished(bytes32,bytes32,bool,bytes)` with `canAck` semantics, CREATE2 factory `OutboxCreated(address,uint32,address,address,string)`.
- **relayer:** `gluwa/usc-message-relayer` ≥ v0.1.0 + #31 (independent ack + claim settlements, `messageCanAck`, `feesInNative`-aware ledger reads).
- **Runtime upgrade:** `spec_version` 131 (inherited from usc-dev's release bump); `supportedChains` storage version **2** — `MigrateV1ToV2` clears `CoreFees` (empty on every network; `token` was the struct's first field, so stale entries would mis-decode) and is version-guarded.

## Expected-red CI (policy, not defects)

- `check-version` / `Check runtime changes → "Check if version.rs was modified"` — the policy requires a spec_version bump whenever `runtime/src/` changes, and per team convention feature PRs never bump it (release commits own the version). Note it previously showed green on this branch only because a rebase artifact had accidentally downgraded `version.rs` (fixed in `51d2ff238`) — red is the honest state. The release that ships this merge bumps to spec 132.
- `integration-test-blockchain` storage-version diff — strict diff vs live devnet/testnet with no pending-migration allowlist; any PR introducing a migration fails it by construction (`supportedChains 0 → 2`).

## Review history

- Two full security-review rounds on the contracts side (usc-contracts #23: B1/B2/B3 blockers found → fixed → verified → approved).
- Alex Todorov's CHANGES_REQUESTED items: all addressed (supported-chains test refactor, indexer handler coverage, `set_core_fee`/`get_core_fee` integration tests, e2e toolchain bumps, betterleaks allowlist) — reviewer has since left; the review needs dismissing.
- Cursor Bugbot: all findings across 4 review passes fixed or triaged with rationale (latest pass: clean).

## Known follow-ups (tracked, deliberately not in this PR)

1. **`usc-messaging` fee-stack rework** — quoter must sign the v3 preimage (`acknowledgmentPrice` back, no `requiresAck` field, `payInNative`), publish scripts move to 2-arg `publishMessage`, `deploy.ts` rewrite. **Gated on a post-#23 `@gluwa/usc-contracts` npm release** (0.1.2 still ships pre-#23 contracts) — asked on the #23 thread.
2. **e2e pins** (`write-ability-e2e.yml:111/120`) stay on the pre-merge SHAs the in-repo scripts speak; bumped together with (1).
3. IaC: pin relayer chart to the tagged release; quoter chart; deploy-and-wire runbook for devnet.


.github/workflows/write-ability-e2e.yml
```diff
@@ -0,0 +1,371 @@
...
+          echo "❌ relayer did not come online"; tail -40 "$LOGS/relayer.log"; exit 1
+
+      - name: Publish a fee'd cross-chain message (requiresAck)
+        working-directory: ./usc-messaging
+        run: |
+          # Quotes from the live quoter, approves ATTEST, calls
+          # RelayerContract.publishAndCollectRelayerFee, and records the messageId to
+          # /tmp/e2e-published.json for the ack + claim assertions.
+          timeout 180 npx tsx scripts/publish-fee.mts "ci-$GITHUB_RUN_ID" >"$LOGS/publish.log" 2>&1 || true
+          tail -20 "$LOGS/publish.log" || true
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 3ae06d32. Dropped the `|| true` — the publish step now tails the log then `exit 1` on failure, so a failed publish fails fast instead of surfacing later as a confusing delivery/ack timeout. (Pre-existing write-ability code, not introduced by the rebase.)

attestor/attestor/src/tasks/write_ability/mod.rs
```diff
@@ -0,0 +1,732 @@
...
+                        Ok(Err(err)) => err.context("outbox listener died"),
+                        Err(join_err) => anyhow!("outbox listener panicked: {join_err}"),
+                    };
+                    if let Some(w) = &set_watcher {
+                        w.abort();
+                    }
+                    if let Some(p) = &set_update_proposer {
+                        p.abort();
+                    }
+                    return Err(Error::WriteAbility(err));
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 3ae06d32. Added `reobs_worker.abort()` on the listener-death error path — it was the one sibling task dropped (detached) rather than aborted there, so reobservation could keep running against a dead listener until the shared cancel token fired. Attestor crate rebuilt clean. (Pre-existing; not from the rebase.)

usc-messaging/scripts/launch-attestors.sh
```diff
@@ -0,0 +1,212 @@
...
+      echo "⚠️  Could not update the EOAValidator set (is VOTE_VALIDATOR_ADDR an EOAValidator from deploy.ts, with this key as admin?). deliverMessage will revert until the set matches. Continuing." >&2
+    fi
+  else
+    echo "⚠️  cast not found on PATH; skipping EOAValidator set sync — run updateAttestorSet(address[]) manually." >&2
+  fi
+fi
+
+cat <<EOF
+
+✅ Attestor set ready (also saved to $ATTESTOR_SET_FILE):
```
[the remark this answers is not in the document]
DylanVerstraete: Valid — the final `Attestor set ready` line prints unconditionally and the e2e gate keys on that exact string, so a failed `updateAttestorSet` (or missing `cast`) is a false-ready that only shows up later as a delivery revert.

Deliberately NOT fixing this in the re-land PR: the script continues-on-failure by design (deploy.ts also seeds the validator best-effort, so the `cast` update can be redundant and its failure benign), and making the gate fail-fast changes e2e gating behaviour — that is the write-ability authors call, not mine on a faithful re-land. Proposed minimal fix: track the sync result and only emit the gated `Attestor set ready` string when the set actually synced (or no validator is configured); on a real sync failure print a distinct line and `exit 1`. Leaving this thread open for that decision. Pre-existing code, not from the rebase.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -0,0 +1,199 @@
...
+    //
+    //    Bound the scan (and thus the cursor) at a CONFIRMED height — `tip - confirmation_depth`,
+    //    mirroring the listener, which never signs above that height. The cursor advances
+    //    permanently, so scanning past the *unconfirmed* tip would let a source reorg re-mine
+    //    OutboxCreated below the cursor and hide the Outbox for the whole process lifetime.
+    let tip = provider
+        .get_block_number()
+        .await
+        .context("read EVM tip for Outbox discovery scan")?;
+    let safe_tip = tip.saturating_sub(confirmation_depth);
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in daa9504. Discovery now bounds its scan at the GRANDPA finalized head via `get_block_by_number(Finalized)`, mirroring the listener's `FinalityPolicy::Finalized`, and only falls back to `tip - confirmation_depth` when the `finalized` tag is unavailable. So the cursor can no longer advance past a reorgable block. (Pre-existing; the listener was moved to finalized but the resolver comment/logic lagged.)

attestor/attestor/src/tasks/p2p/mod.rs
```diff
@@ -427,31 +643,56 @@ async fn handle_swarm(
...
+                    propagation_source,
+                )
+            } else if is_set_update {
+                // We subscribe to the attestor-set-update topic only to publish our proposer's own
+                // votes and keep the mesh propagating them — attestors do NOT aggregate set-update
+                // votes (the relayer does). Accept inbound frames so gossipsub keeps relaying, but
+                // take no action. Crucially, do not fall through to the block-attestation `Vote`
+                // decoder below: a SetUpdateVote SCALE payload would fail to decode and be Rejected,
+                // applying gossipsub penalties to peers relaying legitimate set-update votes (bugbot).
+                libp2p::gossipsub::MessageAcceptance::Accept
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in daa9504. Set-update frames over `MAX_SET_UPDATE_FRAME_BYTES` (derived from `MAX_ATTESTORS`) are now `Ignore`d — no propagation, no peer penalty — so oversized garbage can't be amplified, while legit-sized votes still `Accept`-relay. Kept it `Ignore` rather than `Reject` because attestors don't decode set-update votes (only the relayer aggregates them), so `Reject` would penalize honest relayers — that was the prior fix this one was in tension with.

usc-messaging/scripts/launch-attestors.sh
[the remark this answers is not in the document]
DylanVerstraete: Fixed in daa9504 (had deferred this as a deliberate continue-on-failure, but on reflection the gate should be honest). `launch-attestors.sh` now `exit 1`s on a failed `updateAttestorSet` (or missing `cast` when a validator is configured) instead of unconditionally printing `Attestor set ready` — so the e2e fails fast on a mismatched set rather than surfacing later as a confusing delivery timeout.

.github/workflows/write-ability-e2e.yml
```diff
@@ -0,0 +1,377 @@
+---
```
atodorov, line 1: Existing CI job failures:

- MegaLinter - https://github.com/gluwa/creditcoin3/actions/runs/30460615788/job/90605409149
- Attestor Compatibility - https://github.com/gluwa/creditcoin3/actions/runs/30460618526/job/90619330771 this should be on top of 3.130.0 so it's not expected to fail, please investigate
- cc3-indexer tests failing - https://github.com/gluwa/creditcoin3/actions/runs/30460619657/job/90610814567 (also were failing the same in original PR)
- integration-test-attestator-network failing - https://github.com/gluwa/creditcoin3/actions/runs/30460619657/job/90610814811

.github/workflows/write-ability-e2e.yml
```diff
@@ -0,0 +1,377 @@
...
+    uses: ./.github/workflows/remove-runner.yml
+    with:
+      RUNNER_VM_NAME: "${{ github.event.repository.name }}-${{ github.run_id }}-attempt-${{ github.run_attempt }}"
+      proxy: ${{ matrix.proxy }}
+    secrets: inherit
+
+  e2e:
+    needs:
+      - build
+    runs-on: ubuntu-24.04
```
atodorov: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612728743

.github/workflows/write-ability-e2e.yml
```diff
@@ -0,0 +1,377 @@
...
+          mkdir -p "$LOGS"
+          chmod a+x ./target/release/creditcoin3-node ./target/release/attestor* ./target/release/message-relayer ./target/release/proof-gen-api-server
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+
+      - name: Install Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: 20
```
atodorov: Must be 24!

.github/workflows/write-ability-e2e.yml
```diff
@@ -0,0 +1,377 @@
...
+      # wired: docker cp the binary out of a SHA-tagged gluwa/usc-message-relayer image.)
+      - name: Check out usc-message-relayer
+        uses: actions/checkout@v6
+        with:
+          repository: gluwa/usc-message-relayer
+          # Pin a full commit SHA (not a branch) so this creditcoin3 commit always builds against
+          # the *same* relayer code — `ref: main` made the e2e non-reproducible (the same commit
+          # could pass or fail as the relayer's default branch moved). Bump this deliberately when
+          # a relayer change needs to be exercised here.
+          ref: 9ffe2a29dfcaeccc6df4dc524d76c71955840cd4 # usc-message-relayer main @ #26 (funded-gas delivery + under-funding/timeout hardening)
```
atodorov, line 111: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612742036

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +63,206 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+// emitterAddress is a bytes32 (20-byte EVM address left-aligned in the high bytes).
+type MessagePublishedArgs = [string, string, boolean, string];
+// Outbox: MessageAcknowledged(bytes32 indexed messageId)
+type MessageAcknowledgedArgs = [string];
+
+function eventTimestamp(event: { blockTimestamp?: Date }): bigint {
+    return event.blockTimestamp ? BigInt(event.blockTimestamp.getTime()) : BigInt(Date.now());
+}
+
+export async function handleOutboxCreated(event: FrontierEvmEvent<OutboxCreatedArgs>): Promise<void> {
```
atodorov, line 93: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612927094

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +63,206 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+    // see same-block buffered writes).
+    if (!authenticated) {
+        const counter = await OutboxDatasourceCount.get(chainKey);
+        await OutboxDatasourceCount.create({ id: chainKey, count: (counter?.count ?? 0) + 1 }).save();
+    }
+
+    await createOutboxDatasource({ address });
+}
+
+export async function handleMessagePublished(event: FrontierEvmEvent<MessagePublishedArgs>): Promise<void> {
```
atodorov, line 289: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612928012

cc3-indexer/src/mappings/evmHandlers.ts
```diff
@@ -46,3 +63,206 @@ export async function handleTransactionVerified(event: FrontierEvmEvent<Transact
...
+        acknowledged: false,
+        acknowledgedAt: undefined,
+        acknowledgedTimestamp: undefined,
+        acknowledgedTxHash: undefined,
+    });
+
+    await message.save();
+}
+
+export async function handleMessageAcknowledged(event: FrontierEvmEvent<MessageAcknowledgedArgs>): Promise<void> {
```
atodorov, line 405: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612928506

cc3-indexer/src/mappings/mappingHandlers.ts
```diff
@@ -196,6 +197,42 @@ export async function handleSupportedChainRegistered(event: SubstrateEvent): Pro
     await Promise.all([chainRegistered.save(), suportedChain.save(), newChain.save()]);
 }
 
+// USC write-ability: an operator registered an OutboxFactory for a chain key
+// (supportedChains.OutboxFactoryRegistered { chain_key, outbox_factory_addr }). Recorded for
+// display / to resolve OutboxContract.factory. Discovery of Outboxes does NOT depend on this event
+// — see the chain-wide OutboxCreated watch in datasources.ts (the factory creates Outboxes before
+// it is registered here, so following this event would miss them).
+export async function handleOutboxFactoryRegistered(event: SubstrateEvent): Promise<void> {
```
atodorov, line 184: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612929363

pallets/supported-chains/src/lib.rs
```diff
@@ -288,6 +404,110 @@ pub mod pallet {
...
+        /// CTC (paid via `msg.value`); `Some(address)` in an ERC20 on Creditcoin's EVM
+        /// (attestcoin, pulled via `transferFrom`). A zero `amount` disables the fee. Only
+        /// accounts in the Operators membership (or root) can call this extrinsic.
+        ///
+        /// The value is read live by the EVM through the chain-info precompile
+        /// (`get_core_fee(uint64)`), so changes take effect on the next publish without any
+        /// contract redeploys — including a later native→attestcoin denomination switch.
+        #[pallet::call_index(4)]
+        #[pallet::weight(T::WeightInfo::set_core_fee())]
+        pub fn set_core_fee(
```
atodorov, line 481: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612940624

precompiles/chain-info/src/lib.rs
```diff
@@ -159,6 +177,65 @@ where
...
+            None => Ok(OutboxFactoryResult {
+                factory_addr: Address(H160::zero()),
+                exists: false,
+            }),
+        }
+    }
+
+    #[precompile::public("get_core_fee(uint64)")]
+    #[precompile::view]
+    fn get_core_fee(
```
atodorov: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612949050

pallets/supported-chains/src/lib.rs
```diff
@@ -133,6 +177,34 @@ pub mod pallet {
...
+                    },
+                );
+            }
+
+            for (chain_key, outbox_factory_addr) in &self.outbox_factories {
+                // Mirror the `set_outbox_factory_addr` guard: a zero factory reads as "unregistered"
+                // to the attestor/relayer, so a chainspec must not seed one.
+                assert!(
+                    !outbox_factory_addr.is_zero(),
+                    "Genesis outbox_factories has a zero address for chain_key {chain_key:?}"
```
atodorov, line 204: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612960414

pallets/supported-chains/src/lib.rs
```diff
@@ -273,6 +383,12 @@ pub mod pallet {
 
             ChainIdAndNameToUniqKey::<T>::remove(item.chain_id, item.chain_name.clone());
 
+            OutboxFactories::<T>::remove(chain_key);
+
+            WriteAbilityConfigs::<T>::remove(chain_key);
+
+            CoreFees::<T>::remove(chain_key);
```
atodorov, line 383: All of these probably need to be asserted to in

- cc3-indexer tests
- extrinsic integration test
- handled in cc3-indexer handler

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+            ),
+            Error::<Test>::ZeroWriteAbilityChainKey
+        );
+
+        assert_eq!(WriteAbilityConfigs::<Test>::get(chain_key), None);
+    });
+}
+
+#[test]
+fn remove_chain_clears_write_ability_config() {
```
atodorov: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3612982048

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+            SupportedChain::set_write_ability_config(
+                RuntimeOrigin::signed(2),
+                chain_key,
+                [0u8; 32],
+                true,
+            ),
+            BadOrigin
+        );
+
+        assert_ok!(SupportedChain::set_write_ability_config(
```
atodorov: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3613004766

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+            ),
+            Error::<Test>::ChainNotSupported
+        );
+
+        assert_eq!(WriteAbilityConfigs::<Test>::get(chain_key), None);
+    });
+}
+
+#[test]
+fn set_write_ability_config_should_error_when_chain_key_is_zero() {
```
atodorov: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3613020235

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+        let chain_key = 1;
+
+        assert_noop!(
+            SupportedChain::set_write_ability_config(
+                RuntimeOrigin::root(),
+                chain_key,
+                [0u8; 32],
+                true,
+            ),
+            Error::<Test>::ZeroWriteAbilityChainKey
```
atodorov, line 812: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3613025098

pallets/supported-chains/src/tests.rs
```diff
@@ -549,3 +554,493 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
...
+        );
+    });
+}
+
+#[test]
+fn set_core_fee_works_native() {
+    ExtBuilder.build_and_execute(|| {
+        System::set_block_number(1);
+
+        let chain_key = 1;
```
atodorov: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3613088639

precompiles/chain-info/src/tests.rs
```diff
@@ -143,6 +261,36 @@ fn get_chain_by_key_returns_default_data_with_unknown_chain_key() {
         });
 }
 
+#[test]
+fn get_outbox_factory_address_returns_default_when_not_set() {
```
atodorov: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3613114700

pallets/supported-chains/src/lib.rs
```diff
@@ -288,6 +404,110 @@ pub mod pallet {
...
+            token: Option<H160>,
+            amount: sp_core::U256,
+        ) -> DispatchResult {
+            T::OperatorsOrigin::ensure_origin(origin)?;
+
+            ensure!(
+                SupportedChains::<T>::contains_key(chain_key),
+                Error::<T>::ChainNotSupported
+            );
+            ensure!(token != Some(H160::zero()), Error::<T>::ZeroCoreFeeToken);
```
atodorov: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3613143481

precompiles/chain-info/src/lib.rs
```diff
@@ -159,6 +177,65 @@ where
...
+                // sentinel on the EVM side. The extrinsic rejects Some(H160::zero()), so a zero
+                // token here always means native, never a broken ERC20 config.
+                token: Address(fee.token.unwrap_or_default()),
+                amount: fee.amount,
+                exists: true,
+            }),
+            None => Ok(CoreFeeResult {
+                token: Address(H160::zero()),
+                amount: U256::zero(),
+                exists: false,
```
atodorov: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3613162350

cc3-indexer/abis/outbox.json
```diff
@@ -0,0 +1,21 @@
+[
```
atodorov, line 1: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3613176849

usc-messaging/contracts/src/TestDestination.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: MIT
```
atodorov, line 1: https://github.com/gluwa/creditcoin3/pull/1083#discussion_r3613186298

atodorov (changes requested):
Some of the comments around tests on the original PR were resolved, marked the ones which don't seem to have been resolved yet!

usc-messaging/scripts/deploy.ts
```diff
@@ -0,0 +1,449 @@
...
+    [chainKeyU64.toString(), payee], // (destinationChainKey, owner)
+    [`${EVM_V1_DECODER_LIB}:${evmV1Decoder}`], // link the pre-deployed EvmV1Decoder library
+  );
+  castSendSource(outboxFactory, "createOutbox(bytes32,address)", [
+    chainKeyBytes32,
+    ackValidator, // the Outbox's onlyValidator ack authority
+  ]);
+  const outbox = castCallSource(outboxFactory, "getOutbox(bytes32)(address)", [
+    chainKeyBytes32,
+  ]);
```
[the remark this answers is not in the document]
DylanVerstraete: Confirmed and addressed in `7fd288cc0` — with one constraint worth recording: the *complete* fix (rewriting `deploy.ts` for the fee-integrated stack) is gated on a post-#23 `@gluwa/usc-contracts` npm release. The latest published package (`0.1.2`) still ships the pre-#23 `SimpleOutboxFactory` and has no CREATE2 factory or FeeRegistry at all, so there is nothing attestor-discoverable to deploy from node_modules today. Until that publish lands: the script now carries a loud legacy warning, and the README stops presenting it as the canonical path and points at the fee-integrated deploy scripts that `write-ability-e2e.yml` actually exercises (which deploy from a post-#23 `usc-contracts` checkout and are attestor-compatible — the e2e proves discovery end-to-end). @kevinnguyen17 — could you publish the post-#23 package when convenient? That unblocks the `deploy.ts` rewrite.

attestor/attestor/src/tasks/write_ability/set_update.rs
```diff
@@ -0,0 +1,248 @@
...
+    // omit still-elected attestors (bugbot). Defer until every active attestor is registered — the
+    // proposer polls, and each unregistered node keeps retrying registration, so this converges.
+    if candidate.len() < active_count {
+        tracing::debug!(
+            registered = candidate.len(),
+            active = active_count,
+            "deferring set-update proposal — not all active attestors have registered an EVM address yet"
+        );
+        return Ok(None);
+    }
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in cb41a4828. Real finding — the full-registration guard (added for the earlier partial-set finding) meant one attestor that never enables write-ability stalls `EOAValidator` sync forever. The proposer now defers only while the registered subset is below committee supermajority (`> 2/3` of active, the same formula EOAValidator derives its threshold from): minority sets that would hand control to a few registrants are still refused, a never-registering straggler no longer blocks convergence, and late registrants trigger a follow-up update. Deterministic in on-chain state, so all attestors still sign identical bytes. Unit-tested (`enough_registered_is_committee_supermajority`).

didac-gluwa (dismissed):

.github/workflows/ci.yml
```diff
@@ -699,7 +699,12 @@ jobs:
...
           # designed behavior — the gate is for everything else (task failures, submission
           # errors, invalid event mappings).
-          ALLOWLIST='Failed to reconnect to CC3|Reconnected but at_latest failed|connection (error|lost)|Subscription error while fetching next block|Failed to start finalized-block subscription|ping failed'
+          # The last three entries are the write-ability resolver riding out those same deliberate
+          # outages: its periodic ERROR escalation ("still failing after prolonged retrying"), the
+          # restart-to-rebuild path after a sustained failure budget ("N consecutive times"), and
+          # alloy's close-frame notice when the ws endpoint is bounced. All three are designed
+          # loud-recovery behavior during the induced outages, not faults.
+          ALLOWLIST='Failed to reconnect to CC3|Reconnected but at_latest failed|connection (error|lost)|Subscription error while fetching next block|Failed to start finalized-block subscription|ping failed|Outbox resolution still failing after prolonged retrying|Outbox resolution failed [0-9]+ consecutive times|Received close frame'
```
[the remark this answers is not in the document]
DylanVerstraete: Correct — that entry was inert against this branch. Fixed in 86916ef6f, which adds `outbox poll stalled` and `outbox listener died` (the listener's actual sustained-failure restart path, surfaced through the write-ability supervisor). The `Outbox resolution failed N consecutive times` entry is kept deliberately: it becomes live when #1238's resolver failure budget lands, and an inert-but-forward-looking pattern costs nothing.

Verified: `integration-test-attestator-network` is now green on #1236 with this allowlist.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -0,0 +1,224 @@
...
+        let logs = provider.get_logs(&filter).await.with_context(|| {
+            format!("eth_getLogs OutboxCreated at factory {factory} [{from}..={chunk_to}] failed")
+        })?;
+        if let Some(log) = logs.last() {
+            latest = Some(
+                IOutboxFactory::OutboxCreated::decode_log(&log.inner, true)
+                    .context("decode OutboxCreated log")?
+                    .data
+                    .outbox,
+            );
```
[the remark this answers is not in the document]
DylanVerstraete: Accurate description of the mechanism, accepted as a known V1 limitation rather than fixed here.

Discovery binds the newest `OutboxCreated` for the chain key, and `OutboxFactory.deployOutbox` is permissionless by design (access control lives on `OutboxDeployer`, and the CREATE2 salt binds `msg.sender`, so a squatter can only ever deploy an Outbox the protocol never registers). Production deploys go through OutboxDeployer, which keeps the registry one-Outbox-per-chain-key, so the ambiguous case does not arise on a governed chain — the newest-wins rule exists to make it deterministic rather than emission-order dependent.

The durable fix is authorization at discovery (bind to the governance-registered factory, with a backfill so an Outbox created before its registration is not lost). That was attempted for the indexer in #1239 and closed because the fail-closed half shipped without the backfill half — our own usc-dev deploy created the Outbox 211 blocks before its factory registration, which that change would have silently dropped. Tracking both halves together as a follow-up rather than landing a partial mitigation here.

usc-messaging/scripts/deploy.ts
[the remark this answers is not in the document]
DylanVerstraete: Correct, and already the documented state: `usc-messaging/scripts/deploy.ts` is flagged as a legacy pre-#23 path in f2f20dbf8 precisely because the attestor cannot discover what it registers (2-arg `OutboxCreated`, `getOutbox`-style resolution). It is not the supported route.

The supported paths both go through the CREATE2 factory the attestor scans: `deploy-source-ethers.mts` for the e2e, and `deploy-{dest,source}-devnet.mts` for real networks. Both were exercised end-to-end on usc-dev this week — publish → attest → deliver → ack → claim, three consecutive green runs — so discovery against the registered factory is verified, not just intended. Rewriting/removing `deploy.ts` is tracked with the usc-messaging fee-stack rework (gated on the post-#23 npm publish).

usc-messaging/scripts/publish-fee.mts
```diff
@@ -0,0 +1,35 @@
...
+const provider = new ethers.JsonRpcProvider("http://127.0.0.1:9944", 42, { staticNetwork: true });
+provider.pollingInterval = 800;
+const payer = new ethers.Wallet("0x8075991ce870b93a8870eca0c0f91913d12f47948ca0fd25b49c6fa7cdbeee8b", provider);
+
+// SimpleInbox.deliverMessage abi-decodes the payload as (address destinationContract, bytes data).
+const inner = ethers.toUtf8Bytes(`e2e delivery ${process.argv[2] ?? "1"}`);
+const payload = ethers.getBytes(ethers.AbiCoder.defaultAbiCoder().encode(["address", "bytes"], [a.dest.dapp, inner]));
+const res = await fetch("http://localhost:3010/quote", {
+  method: "POST", headers: { "content-type": "application/json" },
+  body: JSON.stringify({ destinationChain: s.chainKey, targetContract: payer.address, payloadHash: ethers.keccak256(payload), gasLimit: 300000, requiresAck: true }),
```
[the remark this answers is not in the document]
DylanVerstraete: This one is a misread of the contract, no change needed.

`RelayerContract._validatePayloadQuote` requires `q.targetContract == msg.sender` — the quote's `targetContract` is the *source-chain emitter/payer*, not the destination contract. Quoting the field docs: "Source-chain dApp address (emitter). Combined with payloadHash to uniquely identify the published message." Setting it to the payer wallet is what makes the publish succeed; naming the destination dApp there would revert `QuoteTargetMismatch`.

The destination address lives inside the payload (`abi.encode(destinationContract, data)`), which is bound into the quote separately via `payloadHash` — so both are committed, just through different fields. Nothing diverges.

.github/workflows/ci.yml
```diff
@@ -699,7 +699,12 @@ jobs:
...
           # designed behavior — the gate is for everything else (task failures, submission
           # errors, invalid event mappings).
-          ALLOWLIST='Failed to reconnect to CC3|Reconnected but at_latest failed|connection (error|lost)|Subscription error while fetching next block|Failed to start finalized-block subscription|ping failed'
+          # The last three entries are the write-ability resolver riding out those same deliberate
+          # outages: its periodic ERROR escalation ("still failing after prolonged retrying"), the
+          # restart-to-rebuild path after a sustained failure budget ("N consecutive times"), and
+          # alloy's close-frame notice when the ws endpoint is bounced. All three are designed
+          # loud-recovery behavior during the induced outages, not faults.
+          ALLOWLIST='Failed to reconnect to CC3|Reconnected but at_latest failed|connection (error|lost)|Subscription error while fetching next block|Failed to start finalized-block subscription|ping failed|Outbox resolution still failing after prolonged retrying|Outbox resolution failed [0-9]+ consecutive times|outbox poll stalled|outbox listener died|Received close frame'
```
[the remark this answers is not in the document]
DylanVerstraete: Right on both halves, fixed in f256d9a5f.

The root cause was the logging, not the allowlist: `tracing::error!(%err, "⛔ task failed")` used anyhow's plain `Display`, which prints only the outermost context — so every write-ability fatal collapsed to "write-ability: outbox listener died" and the cause that distinguishes a deliberate outage stall from an ABI decode failure was thrown away. That's also exactly why the `outbox poll stalled` pattern never matched.

Now logging `{:#}` (alternate Display, the same fix the resolver already carries for the same reason), so the chain reaches the log. With the cause visible, the allowlist keeps only the specific recovery phrase `outbox poll stalled \(no progress\)` and the broad `outbox listener died` entry is gone — a decode failure or cursor IO error now fails the gate as it should.

attestor/attestor/src/tasks/write_ability/mod.rs
```diff
@@ -0,0 +1,750 @@
...
+                    // leaving it running against the now-dead listener until the shared cancel
+                    // token eventually fires.
+                    reobs_worker.abort();
+                    if let Some(w) = &set_watcher {
+                        w.abort();
+                    }
+                    if let Some(p) = &set_update_proposer {
+                        p.abort();
+                    }
+                    return Err(Error::WriteAbility(err));
```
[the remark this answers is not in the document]
DylanVerstraete: Confirmed real, and fixed — but in #1236 rather than here, since that PR already restructures this exact `select!` arm (it removes the whole-poll timeout that was the other reason the arm could be reached spuriously).

The fix there is the `shared.token.is_cancelled()` short-circuit before the harvest: on shutdown every cancel path in `watch` returns `Ok(())` and drops the sole sender, so the `rx.recv()` arm goes ready alongside the cancel branch and a non-biased `select!` picks between them at random — losing that race must not be reported as an abnormal exit. #1236's `integration-test-attestator-network` run is green with it.

Leaving this thread open deliberately until #1236 lands, so it isn't lost if that PR stalls.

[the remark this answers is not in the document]
DylanVerstraete: Fixed on the base branch now that #1236 is merged (e50f50cfb) — the `shared.token.is_cancelled()` short-circuit lands ahead of the harvest, so losing the cancel/`rx.recv()` race is no longer reported as an abnormal exit. Resolving.

attestor/attestor/src/lib.rs
```diff
@@ -504,7 +569,14 @@ impl Attestor {
...
                         } else {
-                            tracing::error!(%err, "⛔ task failed");
+                            // `{:#}` (alternate Display), not `%err`: these errors carry
+                            // `anyhow::Context` chains, whose plain Display prints ONLY the
+                            // outermost layer — so a listener stall surfaced as the bare phrase
+                            // "write-ability: outbox listener died" and threw away the cause that
+                            // names it (RPC stall vs. ABI decode failure vs. cursor IO). Printing
+                            // the chain is what lets an operator — and the CI error gate — tell a
+                            // deliberate outage recovery from a real defect.
+                            tracing::error!(err = format!("{err:#}"), "⛔ task failed");
```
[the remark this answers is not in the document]
DylanVerstraete: Correct, and precisely diagnosed — the fix above was a no-op for the reason you name. Fixed.

`err` at that log site is this crate's `Error`, not `anyhow::Error`, and its hand-written `Display` rendered the wrapped anyhow with `write!(f, "write-ability: {e}")`. An inner error formats with its own default flags regardless of the outer formatter's, so `{err:#}` never reached the `Context` chain and every write-ability fatal still read `write-ability: outbox listener died`.

The three anyhow-wrapping variants (`Init`, `P2p`, `WriteAbility`) now forward the alternate flag, so `{:#}` prints the whole chain while plain `{}` stays one-line for callers that want it. Added a unit test that asserts the specific cause the CI error gate keys on (`outbox poll stalled (no progress)`) survives into the formatted string — that's what should have caught this in the first place.

attestor/attestor/src/tasks/write_ability/mod.rs
```diff
@@ -0,0 +1,1409 @@
...
+                        // (the resolver already logged per-chunk progress for it), not a dead
+                        // endpoint — same distinction the activation loop draws below.
+                        tracing::info!(
+                            scanned_to = cursor.scanned_to(),
+                            error = %format!("{err:#}"),
+                            "🐢 Outbox rotation check advanced but did not finish this attempt; continuing"
+                        );
+                    } else {
+                        tracing::warn!(error = %format!("{err:#}"), "Outbox rotation check failed; will retry");
+                    }
```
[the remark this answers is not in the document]
didac-gluwa: Fixed in cf8ce9df7: tracked per-call scan progress directly on the cursor (`advanced_last_call`) instead of diffing `scanned_to()` before/after the call, so a rotation reset to genesis (with `resume_rotation_from_checkpoint: false`) is no longer misclassified as a stall. Also fixed the identical pre-existing pattern in the activation loop for consistency.

attestor/attestor/src/tasks/write_ability/mod.rs
```diff
@@ -0,0 +1,1402 @@
...
+                            listener_reconnect,
+                            resolved,
+                            confirmation_depth,
+                            swap_start,
+                            cursor_store,
+                            listener_tx,
+                            listener_token,
+                        )
+                        .await
+                    }));
```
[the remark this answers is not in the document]
DylanVerstraete: Confirmed, and it is not new to this PR: the attestor has been default-only since discovery moved off factory-log scanning (#1292), and `asc-message-relayer` carries the symmetric "Known gap" note in `events/factory.rs` deferring both sides to the per-Outbox multi-watch work (`activeOutboxes` + `OutboxRegistered`, drain until `pendingRemovalBlock`). No rotation has been executed on usc-devnet yet, so the gap is latent today. Tracked in #1373 together with the start-block thread below; fixing it is a feature on the trunk, not a change for this landing PR.

attestor/attestor/src/tasks/write_ability/mod.rs
```diff
@@ -0,0 +1,1402 @@
...
+                        cfg.write_ability_chain_key,
+                        resolved.address,
+                    );
+                    let listener_tx = tx.clone();
+                    // The registry doesn't report the Outbox's creation block, so the replacement
+                    // listener starts at the same boot-time floor as initial activation. Its cursor
+                    // file is keyed by address, so there is nothing persisted to resume from; a
+                    // rotation onto an Outbox that predates `scan_from` would miss its earlier
+                    // `MessagePublished` events, same as the initial-activation case.
+                    let swap_start = scan_from;
```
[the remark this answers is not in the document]
DylanVerstraete: Confirmed as the second half of the same deferral: the swap path starts the replacement listener at the boot-time `scan_from` because the registry read carries no provenance, and reading `OutboxRegistered` for the start block is part of the planned multi-watch design. Tracked in #1373 with the in-flight-drop finding; not changed in this PR.

PR #1216, opened 2026-07-29
fix(attestation): tag only active-attestor commit_attestation submissions [ATTESTOR-V2-008]

creditcoinprotoclaw:
Fixes [ATTESTOR-V2-008](https://github.com/chainlight-io/USC-ATTESTOR-V2-issues/issues/9) — *Shared `provides` Tag Preemption Can Degrade `commit_attestation` Liveness*.

Reopened against a fresh checkout of `usc-dev` with only the single fix commit (2673533), per the history realignment. Supersedes #1201.

## The bug

`PrevalidateAttestationCommit` assigned the shared `(chain_key, digest)` `provides` tag to **every** `commit_attestation` call. The digest is derived purely from public chain data (header number, root, prev digest), so anyone can predict what honest attestors will produce next, and authorization is only enforced in dispatch. Any funded account could therefore:

* **reserve** the single pool slot for the next legitimate digest with a transaction guaranteed to fail (`AttestorNotActive`), or
* **evict** an already-pending legitimate submission by supplying higher priority.

Repeating that across successive digests delays attestation progress. Liveness-only — no integrity impact, and each attempt costs the attacker a fee.

## The fix

Gate the tag on active-set membership. Non-active signers return `ValidTransaction::default()` (admitted, no `provides` tag) — they still fail in dispatch and pay the inclusion fee, so free prevalidation-rejected spam is still deterred; they simply claim no slot.

Per [Didac's review](https://github.com/gluwa/creditcoin3/pull/1201#discussion_r0), the active-set check now runs **before** the duplicate check, so `check_duplicate` only runs for active attestors (whose racing submissions are the reason it exists). Cleaner control flow, no `is_active` temporary.

Deduplication between *real* racing attestors ([ATTESTOR-V2-999](https://github.com/chainlight-io/USC-ATTESTOR-V2-issues/issues/8)) is unchanged.

## Testing

- `prevalidate_tags_only_active_attestor_submissions`: same attestation from an active attestor and a non-attestor — active claims a tag (dedup preserved), non-attestor admitted with empty `provides`, tag sets cannot collide.
- Asserts the empty tag on the existing `prevalidate_attestation_pays_fee_if_not_active_attestor` case.

## Verification

- `cargo test -p pallet-attestation` — 262 passed, 0 failed
- `cargo clippy -p pallet-attestation --all-targets` — clean
- `cargo fmt --all` — clean

### DIFF
--- a/pallets/attestation/src/extensions.rs
+++ b/pallets/attestation/src/extensions.rs
@@ -39,10 +39,12 @@ use crate::pallet::{ActiveAttestors, Call, Config, MaxCatchup, Pallet};
 /// through: they fail in dispatch (`AttestorNotActive`) and pay fees, so a deregistered or
 /// misconfigured node cannot spam free prevalidation-rejected submissions.
 ///
-/// Admitted `commit_attestation` transactions carry a `provides` tag keyed by
+/// `commit_attestation` transactions from *active attestors* carry a `provides` tag keyed by
 /// `(chain_key, digest)` so the pool holds at most one pending submission per attestation —
 /// racing duplicates from other attestors are deduplicated at admission instead of all being
-/// broadcast and racing to dispatch.
+/// broadcast and racing to dispatch. Submissions from non-active signers are admitted **without**
+/// that tag, so they cannot reserve or evict the pool slot belonging to a legitimate attestation
+/// (ATTESTOR-V2-008).
 ///
 /// Other calls are passed through untouched.
 #[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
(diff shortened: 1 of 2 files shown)


atodorov (changes requested):
@DylanVerstraete original PR still has an unresolved comment:
https://github.com/gluwa/creditcoin3/pull/1201#pullrequestreview-4790056156

@didac-gluwa please review!

BradleyOlson64 (approved):

beqaabu (approved):

didac-gluwa (approved):

PR #1218, opened 2026-07-29
fix(attestor): resume startup from the finalized anchor, not an attestation lookup [ATTESTOR-V2-009]

creditcoinprotoclaw:
Fixes [ATTESTOR-V2-009](https://github.com/chainlight-io/USC-ATTESTOR-V2-issues/issues/10) — attestor startup crash-loop after an emergency revert.

Reopened against a fresh checkout of `usc-dev` with only the single fix commit (c5d1b8b), per the history realignment. Supersedes #1200.

## The bug

`fetch_start_point` resolved the latest finalized digest back to a row in `Attestations` and treated a miss as impossible state. That breaks whenever the anchor is checkpoint-backed:

* after `revert_to()`, which clears every stored attestation for the chain and repoints both `LastCheckpoint` and `LastDigest` at the surviving checkpoint — so `LastDigest` names a digest with no `Attestations` entry; and
* on a checkpoint-only chain, where `LastDigest` is absent and the lookup falls back to `LastCheckpoint` for the same reason.

Startup then failed deterministically on every boot, so a supervisor crash-looped the whole fleet after an emergency revert until an operator intervened. Running attestors were unaffected — only the startup path.

## The fix

Resume from the `(height, digest)` anchor `fetch_last_finalized` already returns (mirroring the runtime's `LastDigest` → `LastCheckpoint` order). That covers all four states uniformly — attestation-backed, checkpoint-backed after a revert, checkpoint-only, and true genesis — and removes the unreachable checkpoint fallback branch plus the `DigestNotFound` error that encoded the bad assumption.

Merely turning the failure into a recoverable error would not have been enough: startup would still reject a valid checkpoint-backed state and keep failing on every restart.

## Testing

`startup.rs` had no tests because it took a concrete `Client` (a live subxt connection). A narrow `StartPointQuery` trait (following the existing `CcRpcProvider` convention) makes the resume rules unit-testable. Four tests including the checkpoint-backed regression; the runtime side of the contract is pinned with a `pallet-attestation` test asserting the post-revert state shape.

## Verification

- `cargo test -p attestor startup` — 4 passed, 0 failed
- `cargo test -p pallet-attestation` — passed
- `cargo clippy -p attestor -p pallet-attestation --all-targets` — clean
- `cargo fmt --all` — clean

## Open review note

@BradleyOlson64 suggested on #1200 that the `startup.rs` tests + `StartPointQuery` trait could be dropped in favour of the `pallet-attestation` test alone, since the fake doesn't model production wiring. Left as-is here (the trait keeps the resume-rule logic directly unit-testable); happy to strip it if we prefer the leaner shape — flagging so it isn't lost in the reopen.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1535,6 +1535,7 @@ name = "attestor"
 version = "3.130.0"
 dependencies = [
  "anyhow",
+ "async-trait",
  "attestor-primitives",
  "attestor_pool",
  "axum",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1535,7 +1535,6 @@ name = "attestor"
 version = "3.130.0"
 dependencies = [
  "anyhow",
- "async-trait",
  "attestor-primitives",
  "attestor_pool",
  "axum",
--- a/attestor/attestor/Cargo.toml
+++ b/attestor/attestor/Cargo.toml
@@ -16,6 +16,7 @@ name = "attestor"
 path = "./src/main.rs"
 
 [dependencies]
+async-trait = { workspace = true }
 attestor_pool = { path = "../pool" }
 common = { path = "../common" }
 metrics = { path = "../metrics" }
--- a/attestor/attestor/Cargo.toml
+++ b/attestor/attestor/Cargo.toml
@@ -16,7 +16,6 @@ name = "attestor"
 path = "./src/main.rs"
 
 [dependencies]
-async-trait = { workspace = true }
 attestor_pool = { path = "../pool" }
 common = { path = "../common" }
 metrics = { path = "../metrics" }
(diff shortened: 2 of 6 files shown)


atodorov (changes requested):
@DylanVerstraete original PR still has an unresolved comment:
https://github.com/gluwa/creditcoin3/pull/1200#pullrequestreview-4790779145

@BradleyOlson64 please review again!

didac-gluwa (dismissed):

BradleyOlson64 (dismissed):
LGTM. 

Does look like we might need another rebase though.

DylanVerstraete (approved):

BradleyOlson64 (approved):
CI failing for seemingly unrelated reasons

PR #1227, opened 2026-07-31
ci(benchmarks): stop the weights bot from re-triggering itself

DylanVerstraete:
Every PR touching **weights.rs / **benchmarking.rs accumulated pairs (or
strings) of "Auto-update pallet weights" commits. The loop:

  1. BENCHMARKS triggers on paths ["**weights.rs", "**benchmarking.rs", ...].
  2. exercise-benchmarks re-measures on a fresh Linode VM. Weight measurement
     is timing — a re-run is never byte-identical (observed deltas are sub-1%
     picosecond drift) — so `git diff` is effectively always non-empty.
  3. EndBug/add-and-commit pushes "Auto-update pallet weights".
  4. That commit touches **weights.rs, matches the workflow's own paths
     filter, and re-triggers it. Goto 2.

Each iteration costs two Linode VMs and a full --features=runtime-benchmarks
release build, and only stops via cancel-in-progress luck. writeability PR
#1215 currently shows two consecutive bot commits (2448b2328 + f6710dde2)
whose entire diff is timing noise.

Add a cheap `should-run` guard job: when the PR head commit IS the bot's own
weights commit (author gluwa-bot, subject "Auto-update pallet weights"),
skip the whole workflow — the weights were just regenerated and validated by
the run that produced them, so there is nothing left to measure. Human pushes
still get exactly one weights refresh. workflow_dispatch (no PR head) always
runs.

The two remove-runner jobs keep their always() semantics for cleanup after a
FAILED run, but are additionally gated on the guard so a skipped run does not
try to remove VMs that were never deployed.

The metadata bot (rebuild-new-metadata in ci.yml) needs no equivalent guard:
its output is deterministic, so its own commit re-triggers a run that
produces no diff and stops by itself. Only weights are timing-noisy.


.github/workflows/benchmarks.yml
```diff
@@ -49,8 +85,11 @@ jobs:
 
   rm-gh-runner-build-creditcoin-node-for-testing-benchmarks:
     needs:
+      - should-run
       - build-creditcoin-node-for-testing-benchmarks
-    if: ${{ always() }}
+    # always() so a failed build still gets its runner removed — but not when the whole
+    # workflow was skipped as the bot's own commit (no VM was ever deployed).
+    if: ${{ always() && needs.should-run.outputs.skip != 'true' }}
```
didac-gluwa, line 92: `true AND expr -> expr`

.github/workflows/benchmarks.yml
```diff
@@ -164,8 +203,10 @@ jobs:
 
   rm-gh-runner-benchmark:
     needs:
+      - should-run
       - exercise-benchmarks
-    if: ${{ always() }}
+    # See rm-gh-runner-build-...: cleanup must survive a failed run, not a skipped one.
+    if: ${{ always() && needs.should-run.outputs.skip != 'true' }}
```
didac-gluwa, line 209: `true AND expr -> expr`

.github/workflows/benchmarks.yml
DylanVerstraete: Not quite equivalent in Actions semantics: `always()` isn't a plain `true` — an `if` without a status-check function is implicitly wrapped in `success()`, so `if: needs.should-run.outputs.skip != 'true'` would only run when every needed job **succeeded**. This is the cleanup job: it must also run when the build **fails** (else the Linode VM leaks), which is exactly what `always() &&` disables. See https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions — happy to switch to `!cancelled() &&` if you prefer that spelling (same effect, minus the run-on-cancel).

.github/workflows/benchmarks.yml
DylanVerstraete: Not quite equivalent in Actions semantics: `always()` isn't a plain `true` — an `if` without a status-check function is implicitly wrapped in `success()`, so `if: needs.should-run.outputs.skip != 'true'` would only run when every needed job **succeeded**. This is the cleanup job: it must also run when the build **fails** (else the Linode VM leaks), which is exactly what `always() &&` disables. See https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions — happy to switch to `!cancelled() &&` if you prefer that spelling (same effect, minus the run-on-cancel).

.github/workflows/benchmarks.yml
DylanVerstraete: Confirmed against the docs rather than left as an assertion. GitHub's expression reference says, verbatim:

> A default status check of `success()` is applied unless you include one of these functions.

So `always() && expr` and `expr` are not the same *job* condition: the bare form is implicitly ANDed with `success()` over `needs`, the `always()` form is not. The `true AND expr` reduction holds for the boolean value, but the presence of a status-check function is also what suppresses the implicit `success()` — that's the part that gets lost.

There's direct evidence in this repo too. On `be286af5`, `build-creditcoin-node-for-testing-ci / build-creditcoin-node` **failed** while `rm-gh-runner-build-creditcoin-node-for-testing (build-creditcoin-node-for-testing-ci) / remove-runner` **succeeded** — same `always()` pattern over in `ci.yml`. Drop the `always()` there and the teardown is skipped in exactly that case, leaking the Linode VM.

Also withdrawing my own offer above to switch to `!cancelled()` — that would be a regression here, not a neutral respelling. This workflow sets `cancel-in-progress: true`, and the self-trigger loop this PR fixes currently only terminates *by* cancellation, so cancelled runs are the common path — precisely when a deployed VM is most likely to be left behind. `always()` is the correct spelling for teardown; `!cancelled()` would leak in the frequent case.

Keeping both cleanup conditions as-is. If you think the next reader will trip on the same thing, I'm happy to spell the reasoning out in a code comment at both sites.

.github/workflows/benchmarks.yml
DylanVerstraete: Same as the thread on `rm-gh-runner-build-creditcoin-node-for-testing-benchmarks` above — see https://github.com/gluwa/creditcoin3/pull/1227#discussion_r3704369274 for the docs quote ("a default status check of `success()` is applied unless you include one of these functions") and the `be286af5` case where this exact pattern kept a Linode VM from leaking after a failed build.

This is the second of the two teardown jobs, so the reasoning is identical: `always()` suppresses the implicit `success()` over `needs` so cleanup survives a failed `exercise-benchmarks`, and `skip != 'true'` excludes the one case where no VM was ever deployed. Keeping as-is.

beqaabu (approved):

didac-gluwa (approved):

PR #1239, opened 2026-08-04
feat(indexer): fail-closed Outbox discovery with quarantine + backfill

DylanVerstraete:
P1/P2-1 audit follow-up to #1215, reworked from the closed first cut: **fail-closed discovery + the backfill that makes fail-closed safe to ship.**

## The bug

`handleOutboxCreated` watched `OutboxCreated` chain-wide by topic and created an `OutboxContract` plus a **dynamic datasource** for any emitter — persistent, security-sensitive (and reorg-unsafe) indexer state any contract could mint up to a per-key cap of 32 (audit P2-1).

## Why the first cut was closed

It flipped discovery to fail-closed (emitter must equal the governance-registered factory for the chain key) — correct direction, but *fail-closed alone is fail-forever*: an `OutboxCreated` indexed before its `OutboxFactoryRegistered` was dropped permanently. That ordering race is real — on usc-dev the OutboxCreated landed ~200 blocks before the factory registration and needed a manual reindex. The first cut had to rewrite the deploy script to survive its own e2e, i.e. it encoded a deployment-ordering footgun into the indexer.

## The rework

**No more dynamic datasources at all.** `MessagePublished` / `MessageAcknowledged` move to chain-wide topic datasources, and every event is authorized per event in its handler:

- A message is indexed **iff its emitting contract has an `OutboxContract` row** — the row's existence is the authorization. The P2-1 attack surface (counterfeit persistent datasources) ceases to exist rather than being capped; a counterfeit event now costs a bounded quarantine row or a log line. Acks additionally verify the emitting contract owns the message — chain-wide watching would otherwise let any contract flip a real message to acknowledged.
- A rejected `OutboxCreated` is **quarantined** (`PendingOutbox`), gated on its chain key existing as a `SupportedChain` and capped at 8 per key. Messages observed on a pending Outbox are quarantined alongside it (capped 256/Outbox), including acks that arrive while pending.
- `handleOutboxFactoryRegistered` **promotes** whatever its registration retroactively authorizes — Outbox + full message lifecycle, original block/tx provenance preserved. `handleSupportedChainRemoved` purges the quarantine for the removed key.

Wrong deploy ordering now degrades to "indexed a few blocks late" instead of "never indexed". As a bonus, message indexing is now fully reorg-safe (dynamic-datasource metadata never was).

## Testing

- **New `handleOutboxBackfill.test.ts`** drives the wrong order end to end: announce + publish + ack *before* registration → asserts quarantine contents; register after → asserts the promoted `OutboxContract` (original provenance), the backfilled message with `acknowledged: true`, and an empty quarantine.
- `handleOutboxLifecycle.test.ts` keeps the ordered path and now asserts the impostor is quarantined, not admitted.
- `yarn build` + `yarn lint` (cc3-indexer), `yarn typecheck` + `yarn lint` (cli) all green.

## Deploy notes

- The indexer chart subPath-mounts its own `project.yaml` — the configmap must be regenerated when this ships (the datasource list changed).
- Schema adds two entities → requires a reindex.

### DIFF
--- a/.github/workflows/write-ability-e2e.yml
+++ b/.github/workflows/write-ability-e2e.yml
@@ -181,16 +181,16 @@ jobs:
           .github/wait-for-ethereum.sh 'http://127.0.0.1:8545'
           .github/wait-for-creditcoin.sh 'http://127.0.0.1:9944'
 
-      - name: Deploy fee stack (dest + source incl. AcknowledgmentValidator) + register factory
+      - name: Deploy fee stack (dest + source incl. AcknowledgmentValidator)
         working-directory: ./usc-messaging
         # deploy-source-ethers.mts deploys the fee-integrated Outbox via the CREATE2 factory plus the
         # quoter, vaults, RelayerContract and the AcknowledgmentValidator (wired as the Outbox
-        # validator); register-factory.mjs registers the factory in the CC runtime so the attestor's
-        # chain-info precompile resolves it. Addresses land in /tmp/e2e-deploy.json.
+        # validator). The source deploy registers the factory in the CC runtime BEFORE creating the
+        # Outbox, so the indexer's chain-wide discovery handler can authenticate OutboxCreated without
+        # accepting attacker-controlled emitters. Addresses land in /tmp/e2e-deploy.json.
         run: |
           npx tsx scripts/deploy-dest-ethers.mts
           npx tsx scripts/deploy-source-ethers.mts
-          node scripts/register-factory.mjs
 
       - name: Write .env from deployed addresses
         working-directory: ./usc-messaging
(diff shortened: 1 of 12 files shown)


DylanVerstraete:
Closing after review: fail-closed discovery without a backfill is a data-loss footgun that our own usc-dev deploy already demonstrates — the live Outbox was created at block 362796 and its factory registered at 363007, so under this change any fresh reindex permanently drops the production Outbox and every message on it (block-ordered processing means the registration written at 363007 can never authorize an event at 362796). The comment this PR deletes predicted exactly that.

Path to reopening: on OutboxFactoryRegistered, rescan the registered factory's prior logs and backfill missed OutboxCreated events — that makes fail-closed lossless for any deploy ordering and strictly better than the current 32-datasource cap. Until then the cap stands as the V1 mitigation.

DylanVerstraete:
Closing after review: fail-closed discovery without a backfill is a data-loss footgun that our own usc-dev deploy already demonstrates — the live Outbox was created at block 362796 and its factory registered at 363007, so under this change any fresh reindex permanently drops the production Outbox and every message on it (block-ordered processing means the registration written at 363007 can never authorize an event at 362796). The comment this PR deletes predicted exactly that.

Path to reopening: on OutboxFactoryRegistered, rescan the registered factory's prior logs and backfill missed OutboxCreated events — that makes fail-closed lossless for any deploy ordering and strictly better than the current 32-datasource cap. Until then the cap stands as the V1 mitigation.

PR #1246, opened 2026-08-05
Enable peer reputation scoring on non-attestation message paths

BradleyOlson64:


DylanVerstraete (changes requested):
The core fix is correct and well-documented — verified against the pinned gossipsub: `mark_invalid_message_delivery` → `stats_or_default_mut` returns `None` for a topic absent from `PeerScoreParams::topics`, so `Reject` on the three write-ability topics really was a silent no-op, and installing params for every subscribed topic at construction is the right shape. The credited/penalty-only split and its "positive credit is summable across topics" rationale are exactly right. One placement is wrong, though, by the PR's own criterion:

**The reobservation topic belongs in `penalty_only_topics`, not `credited_topics`.**

The doc comment says the reobservation topic "verifies against our own RPC first" — but that's not when scoring happens. In `handle_swarm` the reobservation branch returns `Accept` right after `ReobservationRequest::decode_bytes` + a chain-key equality check (`p2p/mod.rs` — `reobs_tx.try_send(request)` → `MessageAcceptance::Accept`, and the queue-full arm *also* returns `Accept`). The actual RPC verification runs later in `run_reobservation_worker`, and its verdict never feeds back into `report_message_validation_result`.

`ReobservationRequest` is four unsigned plain fields (`chain_key`, `message_id`, `tx_hash`, `block_height`) — no signature, nothing to forge. Any peer can emit well-formed requests with random message ids and farm P2 first-message-delivery credit at zero cost, up to the P2 cap, then spend that buffer absorbing P4 penalties earned by forging on the topics that do validate. That's precisely the attack this PR articulates for the set-update topic; the reobservation topic has the same accept-without-validating shape, just one hop less obvious. (The per-source rate limiter runs *before* decode and returns `Ignore`, so it bounds the farming rate but not the principle.)

Suggested change: move `reobs_topic` from `credited_topics` to `penalty_only_topics` in `p2p/mod.rs`, fix the `P2PBehavior::new` doc paragraph that claims RPC-first verification, and flip the test's expectation for the reobservation topic (its `first_message_deliveries_weight` should assert `== 0.0`, which also makes the regression guard encode the right invariant going forward).

Everything else looks good: message votes (`ingest::validate_and_count`) recover the signer and check the active set before `Accept`, and block-attestation votes go through `verify_vote` — those two genuinely earn P2. The lockstep comment tying topic construction to the `subscribe` calls plus the totality test is a solid guard against reintroducing the finding.

BradleyOlson64:
Addressed comments

BradleyOlson64:
Addressed all comments here. Folding into writeability branch so this pr doesn't need another messy rebase.

PR #1252, opened 2026-08-06
fix(attestor): damp reconnect storms with a cross-cycle flap cooldown (backport for cc3-testnet)

DylanVerstraete:
Backport of #1251 (writeability branch) to `main` — **this is the code fix for today's cc3-testnet Sepolia attestation outage** and should ride the next testnet hotfix release.

## Incident (6 Aug, ongoing at time of writing)

04:51:13 UTC: Google's Blockchain Node Engine dropped the Sepolia WS (`generic::unavailable`). All 6 Sepolia attestors (both testnet clusters) began reconnecting at ~1/s with no cross-cycle backoff — ~50 dials/sec sustained against the endpoint — and hammered themselves into a quota lockout (`generic::resource_exhausted`) they cannot escape. Sepolia attestation has been dead since, with **zero restarts and zero alerts**: the pods look healthy to k8s while emitting nothing but reconnect errors (~41k/hour each). eth-mainnet attestors on the same clusters are unaffected.

## Why the existing backoff didn't fire

`ReconnectingEthRpcProvider`'s exponential backoff is constructed fresh inside every `reconnect()` call. A rate-limited provider *accepts the handshake and then kills the first request*, so each cycle 'succeeds', the backoff resets to its 100 ms base, and the loop sustains itself.

## Fix

Cross-cycle flap damper (pure, unit-tested decision function):
- connection dying < 60 s escalates a cooldown 1 s → 5 min cap, slept BEFORE the next reconnect; surviving ≥ 60 s resets it (zero added latency on fresh incidents)
- rate-limit-shaped errors (429 / `resource_exhausted` / quota, incl. the verbatim Google close-frame from this incident) jump to a 30 s floor immediately

With this, today's storm becomes ~1 dial / 5 min / pod instead of ~11/sec — the attestors keep probing (no give-up ceiling; recovery stays automatic) without feeding the lockout.

**Note for reviewers**: the same commit lands on `writeability-off-usc-dev` via #1251; when the trunks next reconcile the merge will see both copies — identical content, so it should fold cleanly.

**This PR does not restore testnet by itself** — the quota is already burnt. Immediate ops action (separate): repoint the `cc3-test-attestor-sepolia-eth-url` secret at a fallback Sepolia provider and roll the 6 pods.

## Testing
`cargo test -p continuity --lib` (8 passed, 4 new) · clippy `-D warnings --all-features` clean

### DIFF
--- a/.github/apply-ubuntu-upgrades.sh
+++ b/.github/apply-ubuntu-upgrades.sh
@@ -5,5 +5,11 @@
 sudo systemctl stop unattended-upgrades.service
 sudo systemctl disable unattended-upgrades.service
 
+# Never upgrade the bootloader on an ephemeral CI runner: grub-pc's postinst is interactive
+# (device re-selection) and periodically breaks outright on Linode images (2026-08-06: postinst
+# exit 1 failed "Provision VM" on every PR). The VM is destroyed after one workflow — a grub
+# update can only hurt it.
+sudo apt-mark hold grub-pc grub-pc-bin grub2-common grub-common shim-signed || true
+
 sudo apt-get update
-sudo apt-get upgrade -y
+sudo DEBIAN_FRONTEND=noninteractive apt-get upgrade -y
--- a/common/continuity/src/rpc.rs
+++ b/common/continuity/src/rpc.rs
@@ -701,6 +701,21 @@ mod flap_tests {
         assert!(d >= Duration::from_millis(RATE_LIMIT_MIN_COOLDOWN_MS));
     }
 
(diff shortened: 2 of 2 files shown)


DylanVerstraete:
Parking this per Dylan: no backports to main/testnet for now — the fix lands on the usc-dev line only (#1251). Branch `backport/reconnect-flap-cooldown-main` stays; reopen when the team decides to take it to the testnet release line (it remains the code fix for the 6 Aug Sepolia attestor outage — until then testnet relies on the ops-side failover/quota fix alone).

PR #1259, opened 2026-08-11
Move attestor cc3 client retry logic to cc-client itself and use it on proof-builder

didac-gluwa:
This pull request refactors the retry logic for RPC calls by moving the `retry` module from the `attestor` crate to the shared `cc-client` crate, making it reusable across the codebase. It updates all usages to reference the new location, improves documentation and naming for clarity, and updates dependencies to ensure proper workspace usage. Additionally, it introduces a reconnecting wrapper for the Creditcoin RPC provider in the continuity builder.


BradleyOlson64 (dismissed):
LGTM

beqaabu (dismissed):

BradleyOlson64 (approved):

DylanVerstraete:
## Verdict

The refactor is sound and the motivation is real — nothing in `continuity` ever called `Client::reconnect()`, so a dropped WS was terminal until process restart. The `attestor::retry` → `cc_client::retry` move is faithful (299 lines out, the same logic in, only doc rewording and `_ =` → `() =` in the select arms), all 8 attestor call sites updated, `tokio-util` correctly hoisted to the workspace. `cargo check -p continuity -p cc-client -p attestor -p proof-gen-api-server --all-targets` is clean with zero warnings.

One change is better than the description gives it credit for: `proof-gen-api-server` previously built `Arc::new(cc3_client.clone())` per chain, and `Client::clone` gives each clone a **fresh** `reconnect_state` (documented at `common/cc-client/src/lib.rs:419`). So N chains meant N uncoordinated `BackoffState`s hammering the same endpoint. Hoisting to one shared `Arc<CcClient>` fixes that properly rather than cosmetically.

Five findings below. Findings 3–5 are cheap; 1 and 2 are the ones I'd want answered before merge, and they share a root cause — the 15s budget was chosen without reference to the bounds of the machinery it wraps.

---

### 1. The 15s call budget is smaller than the reconnect machinery it wraps

All three numbers are hardcoded:

| | value | where |
|---|---|---|
| provider call budget | **15s** | `common/continuity/src/builder/mod.rs:191`, `proof-gen-api-server/src/lib.rs:266` |
| per-dial cap inside `reconnect()` | **15s** | `common/cc-client/src/lib.rs:286` (`CONNECT_TIMEOUT`) |
| max between-attempt backoff | **30s** | `common/cc-client/src/lib.rs:297` (`BackoffState::MAX_MS`) |

`run()` wraps the *entire* `with_retries` loop, and `reconnect()` sleeps on `next_attempt_at` **before** dialing (`lib.rs:405-407`). So:

- **Unreachable host** (dial hangs, no SYN-ACK): the first `reconnect()` burns the full 15s inside `build_inner`, the outer timeout fires — exactly one attempt, no retry at all.
- **Sustained outage with grown backoff**: a single `sleep_until` can be up to 30s, twice the whole budget — zero retries, and every call still costs 15s before failing.
- Retries only really happen when dials fail *fast* (ECONNREFUSED), where you get roughly 7 attempts as backoff ramps 200ms → 12.8s.

This is the same interaction as the earlier "Call timeout blocks reconnect backoff" review comment, which is marked resolved — but the resolution turned the constant into a parameter without changing the value, so the behaviour is unchanged. Either the budget needs to exceed `CONNECT_TIMEOUT + MAX_MS`, or these callers should accept fail-fast and skip the retry wrapper; right now it pays the cost of both and gets the benefit of neither.

Worth naming the underlying tension explicitly: `retry.rs` states its policy as "unbounded retries … ride out the outage rather than crash the task", while `proof-gen-api-server` is an HTTP handler that must not hang. Those are contradictory requirements, and the current code resolves it silently in favour of the timeout.

### 2. Bulk scans now carry a cap they can never satisfy

`get_checkpoints_for_chain` (`common/cc-client/src/lib.rs:1014`) is an unbounded storage-prefix iteration — `.iter(address)` paginating `iter.next().await` over the whole `Checkpoints` prefix for the chain. `get_attestations_for_chain` is the same shape. Previously `impl CcRpcProvider for CcClient` called through with **no wall-clock cap at all**; now both are bounded at 15s.

The failure mode is worse than "fails once". When the timeout aborts mid-scan, `with_retries` restarts `f` from the beginning, so a scan whose data no longer fits in 15s can **never** succeed — every retry restarts and re-times-out. This grows with chain history, which makes it a latent time bomb rather than a present bug.

This is the "Timeout aborts healthy bulk RPCs" comment, also marked resolved by parameterization. I'd want measured p99 for those two calls against mainnet-sized state before merging.

### 3. The cancellation token is a dummy

```rust
cc_client::retry::with_retries(&self.client, &CancellationToken::new(), f)
```

A fresh token that nothing ever cancels. Every other `with_retries` call site in the repo passes a real `&shared.token`. `retry.rs`'s own module doc says "the token is also this call's only cooperative-cancellation path", and the new provider's doc says it "mirrors `ReconnectingEthRpcProvider`" — but that one explicitly detects `user::Shutdown` in the error chain and returns immediately rather than burning retry budget (`common/continuity/src/rpc.rs:77-83`).

In practice `tokio::time::timeout` drops the future, so this is not an unbounded hang. The cost is that shutdown waits up to 15s per in-flight call, and the wrapper silently opts out of a contract the module documents. `ReconnectingCcRpcProvider::new` should take a `CancellationToken` — both construction sites have a shutdown path available — or the doc should state plainly that cancellation is deliberately not wired, and why.

### 4. `CC3_CALL_TIMEOUT` does not exist

`common/continuity/src/rpc.rs:323` says "bounded by [`CC3_CALL_TIMEOUT`]", but the code uses `self.timeout` and no such constant exists anywhere in the repo:

```
$ RUSTDOCFLAGS="-D rustdoc::broken_intra_doc_links" cargo doc -p continuity --document-private-items
error: unresolved link to `CC3_CALL_TIMEOUT`
   --> common/continuity/src/rpc.rs:323:22
    |
323 |     /// bounded by [`CC3_CALL_TIMEOUT`] so a sustained outage fails fast instead of hanging.
    |                      ^^^^^^^^^^^^^^^^ no item named `CC3_CALL_TIMEOUT` in scope
```

CI will not catch this because `run` is private and rustdoc skips private items by default. It looks like a leftover from before the timeout was parameterized. Reintroducing it as a named constant would fix the dangling link *and* de-duplicate the bare `15` currently copy-pasted into two files.

### 5. No tests for the new wrapper

`common/continuity/src/rpc.rs` has no tests at all, and this PR adds timeout + retry + reconnect behaviour to it — exactly the logic worth pinning. `MockCcRpcProvider` mocks the trait, not this wrapper. A test with an always-failing closure asserting the attempt count and that the timeout fires would catch finding 1 directly.

Credit where due: `retry.rs`'s string-classification regression tests moved over intact, and those are the valuable ones — they pin every upstream `Display` string the classifier depends on, which is the fragile part of that module.

PR #1273, opened 2026-08-20
fix(cli): report spendable balance using the runtime's own rule

beqaabu:
## Problem

The stable2512 SDK bump moved pallet-staking from a legacy `balances.locks` lock to a `HoldReason::Staking` hold, so bonded funds now sit in `reserved` instead of inside `free`.

The CLI read `derive.balances.all()`'s **`availableBalance`**, which is `max(0, free - lockedBalance)` and only ever inspects locks. polkadot-js documents that field as correct only for the legacy `AccountInfo` shape and points to `transferable` instead:

> Calculated available balance. This uses the formula: `max(0, free - locked)`. This is only correct when the return type of `api.query.system.account` is `AccountInfo` [...] See `transferable` for the correct balance calculation.

Two bugs followed, both reproduced against a migrated devnet account with a 1,083,743 CTC stake and a 100 CTC attestor bond lock (`free=4004.8069 reserved=1083743.7109 frozen=100.0000`):

**1. `balance` under-reported Transferable, and it gates transactions.** `frozen` is a floor on `free + reserved`, so a reserve discounts it, but `availableBalance` subtracts the lock unconditionally:

| | before | after | truth |
|---|---|---|---|
| Transferable | 3904.8068 | **4004.8068** | 4004.8069 |

This is not cosmetic. `canPay()` gates every transaction on the same value, so `send`, `evm/fund`, `evm/withdraw`, `proxy/actions` and all the `staking` commands rejected legitimate transactions as underfunded.

**2. `attestor show-stash-balance` missed held stake entirely.** It printed `lockedBalance` directly, reporting **100 CTC** locked against **1,083,843 CTC** actually encumbered. The 100 CTC it showed was the attestor bond, reported correctly: the attestation pallet keeps that bond as a `b0ndl0ck` lock (`pallets/attestation/src/ledger.rs:76`), so it does land in `lockedBalance`. What it dropped was the separate **1,083,743 CTC** pallet-staking hold, which is ordinary staking rather than the attestor bond and is invisible to `lockedBalance`. `locked` therefore has to sum both terms; they are disjoint mechanisms, so there is no double count.

## Fix

- Use **`transferable`** (`free - max(maybeEd, frozen - reserved)`, matching pallet-balances `reducible_balance()`). `availableBalance` stays as a fallback for chains still returning the old account shape. The result is clamped at zero as belt-and-braces only: the pinned 16.5.2 derive already clamps both fields at the source, so the clamp matters only if that pin is ever moved backwards.
- Point `attestor show-stash-balance` at the shared `getBalance()` so it cannot drift from this path again. Re-deriving fields locally is how it diverged in the first place.
- `locked` keeps its existing meaning, `lockedBalance + stakingHold`.

## On `locked` (a wrong turn, corrected)

The first push derived `locked` as `total - transferable` so the rows would reconcile. That was wrong: it swept in every reserve on the account, including the proxy deposit. CI caught it on the valid-proxy integration runs, off by exactly the 1000 planck `ProxyDepositBase + ProxyDepositFactor`:

```
bond              Expected "0"                     Received "1000"
withdraw-unbonded Expected "123000000000000000000"  Received "123000000000000001000"
```

Reverted in a34e838. The trade-off is now explicit in a comment: `Transferable + Locked` does **not** sum to `Total` for an account whose reserve already covers a freeze, because the freeze is counted in `locked` while restricting nothing. That is the right call, since `locked` answers "how much stake is encumbered", not "how much can I not spend".

## Verification

Before and after, from two builds against the same devnet account (`--json`, exact plancks):

```
BEFORE   "transferable": "3904806867834855750612"
AFTER    "transferable": "4004806867834855750612"     +100 CTC
```

`transferable` checked against `reducible_balance()` for **11 devnet accounts covering every balance shape** (hold+lock, hold only, lock only, reserved-without-hold, plain): 11/11 match.

`prettier` and `eslint --max-warnings 0` clean. Unit tests 65/65, including 4 new cases in `balance.test.ts` covering field preference, the null fallback, and negative clamping.

## Reviewer notes

- **Only accounts with both `frozen > 0` and `reserved > 0` were affected.** That was rare before the lock-to-hold migration and is now the norm for anyone who both stakes and runs an attestor, because the attestation pallet still uses its own legacy `b0ndl0ck` lock (`pallets/attestation/src/ledger.rs:11`) while staking uses holds. So this is a standing bug for that class of user, not a migration-window artifact. Whether the attestation pallet should also move to holds is a separate question worth raising - and with `ExistentialDeposit = 0`, `frozen - reserved` goes negative for that user, so the `b0ndl0ck` bond restricts nothing about spendable balance once the account has staking reserves. Tracked separately.
- **Not covered:** `attestor show-stash-balance` was not exercised end to end. Its EVM to AccountId lookup via `HashedAddressMapping` did not resolve to the attestation ledger of the account I had available, so it exits with "No ledger found". The code path is now shared with `balance`, which is verified, but the command itself is untested.
- The same one-field audit (`availableBalance` vs `transferable`) is worth running over any other balance consumer, including the staking dashboard and the wallet. Inside this repo the sweep is now complete: the only remaining reader is `cli/src/test/blockchain-tests/precompiles/substrate-transfer.test.ts:44,75`, whose destination is a freshly generated account with no locks or reserves, so `availableBalance == transferable` there and it is unaffected.

### DIFF
--- a/cli/src/lib/balance/index.ts
+++ b/cli/src/lib/balance/index.ts
@@ -52,9 +52,11 @@ export async function getBalance(address: string, api: ApiPromise) {
         transferable,
         bonded: stakingInfo?.stakingLedger.active?.unwrap() || new BN(0),
         evm: new BN(0), // Get Balance does not reflect EVM balance, it must be added manually
-        // Staking-scoped encumbrance: the legacy lock plus the HoldReason::Staking hold that
-        // replaced it. Deliberately NOT `total - transferable`: that would also sweep in
-        // unrelated reserves such as the proxy deposit, which is not locked stake.
+        // Staking-scoped encumbrance: `balances.locks` (which is where the attestation
+        // pallet's own `b0ndl0ck` bond still lives) plus the HoldReason::Staking hold that
+        // pallet-staking moved to. Both terms are required - they are disjoint mechanisms, so
+        // there is no double count. Deliberately NOT `total - transferable`: that would also
+        // sweep in unrelated reserves such as the proxy deposit, which is not locked stake.
         locked: balancesAll.lockedBalance.add(stakingHold),
         total,
         unbonding: calcUnbonding(stakingInfo),
@@ -74,7 +76,13 @@ export async function getBalance(address: string, api: ApiPromise) {
  * `reducible_balance()`. polkadot-js documents `availableBalance` as legacy and points here.
  *
  * It is `null` on chains still returning the pre-FrameSystemAccountInfo shape, hence the
- * fallback, and @polkadot/api 15.x omits the outer clamp, hence the BN.max.
+ * fallback. The BN.max is belt-and-braces: the derive already clamps both fields at the source
+ * in the pinned 16.x, so it only matters if that pin is ever moved backwards.
+ *
+ * Note the hidden coupling to `ExistentialDeposit = 0` (`runtime/src/lib.rs`): that is what
+ * makes the `maybeEd` term a no-op today, so the reported figure equals `free` exactly for a
+ * stash whose freezes are covered by its reserves. If the ED ever becomes non-zero, this
+ * tightens by one ED and every `canPay()` check tightens with it.
  */
 export function getTransferable(balancesAll: DeriveBalancesAll): BN {
     const transferable = balancesAll.transferable ?? balancesAll.availableBalance;
--- a/cli/src/lib/balance/index.ts
+++ b/cli/src/lib/balance/index.ts
@@ -42,6 +42,7 @@ export interface AccountBalance {
 export async function getBalance(address: string, api: ApiPromise) {
     const balacesAll = await getBalancesAll(address, api);
     const stakingInfo = await getStakingInfo(address, api);
+    const stakingHold = await getStakingHoldBalance(address, api);
 
     const total = balacesAll.freeBalance.add(balacesAll.reservedBalance);
     const transferable = getTransferable(balacesAll);
@@ -51,12 +52,10 @@ export async function getBalance(address: string, api: ApiPromise) {
         transferable,
         bonded: stakingInfo?.stakingLedger.active?.unwrap() || new BN(0),
         evm: new BN(0), // Get Balance does not reflect EVM balance, it must be added manually
-        // Everything the account holds but cannot spend right now: reserves (where
-        // pallet-staking now parks bonded funds, as a HoldReason::Staking hold) plus whatever
-        // the freezes still withhold. Derived as `total - transferable` rather than by summing
-        // individual locks and holds, so it stays correct for any encumbrance mechanism and
-        // keeps Transferable + Locked == Total.
-        locked: total.sub(transferable),
+        // Staking-scoped encumbrance: the legacy lock plus the HoldReason::Staking hold that
+        // replaced it. Deliberately NOT `total - transferable`: that would also sweep in
+        // unrelated reserves such as the proxy deposit, which is not locked stake.
+        locked: balacesAll.lockedBalance.add(stakingHold),
         total,
         unbonding: calcUnbonding(stakingInfo),
     };
@@ -80,6 +79,16 @@ export async function getBalance(address: string, api: ApiPromise) {
 export function getTransferable(balancesAll: DeriveBalancesAll): BN {
     const transferable = balancesAll.transferable ?? balancesAll.availableBalance;
     return BN.max(new BN(0), transferable);
+}
+
+/**
+ * Bonded stake now sits in a HoldReason::Staking hold rather than a `balances.locks` lock, so
+ * `lockedBalance` alone reports zero for any stash that has been through the lock-to-hold
+ * migration.
+ */
+async function getStakingHoldBalance(address: string, api: ApiPromise): Promise<BN> {
+    const holds = await api.query.balances.holds(address);
+    return holds.filter((hold) => hold.id.isStaking).reduce((total, hold) => total.iadd(hold.amount), new BN(0));
 }
 
 export async function getBalancesAll(address: string, api: ApiPromise) {
--- a/cli/src/lib/balance/index.ts
+++ b/cli/src/lib/balance/index.ts
@@ -2,7 +2,7 @@ import { ApiPromise, parseUnits, MICROUNITS_PER_CTC } from '..';
 import { BN } from '@polkadot/util';
 import Table from 'cli-table3';
 
-import type { DeriveStakingAccount } from '@polkadot/api-derive/types';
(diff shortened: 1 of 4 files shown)


didac-gluwa (dismissed):

didac-gluwa (dismissed):

cli/src/lib/balance/index.ts
```diff
@@ -44,32 +44,58 @@ export async function getBalance(address: string, api: ApiPromise) {
     const stakingInfo = await getStakingInfo(address, api);
     const stakingHold = await getStakingHoldBalance(address, api);
 
+    const total = balacesAll.freeBalance.add(balacesAll.reservedBalance);
```
jakerumbles: Typo here `balacesAll` missing `n`. Not a big deal

jakerumbles (dismissed):
Think this all looks good. I can actually test it for real in the morning if this is still unmerged.

cli/src/lib/balance/index.ts
beqaabu: fixed on the latest commit

DylanVerstraete (dismissed):

DylanVerstraete:
**Verdict: the fix is correct.** I checked the diagnosis against the vendored `@polkadot/api` rather than taking the description's word for it, and it holds up.

What I verified:

- `cli/package.json` pins `@polkadot/api` **16.5.2**. In `@polkadot/api-derive/balances/all.js:32-34`, `transferable = bnMax(0, free - bnMax(maybeED, frozen - reserved))`, matching pallet-balances `reducible_balance()`. Line 37 computes `availableBalance = bnMax(0, free - lockedBalance)` from locks only. So the root-cause analysis is exactly right, and pointing `canPay()` at `transferable` is the correct fix.
- `RuntimeHoldReason` has exactly two variants, `Staking | Session` (`cli/src/lib/interfaces/types-lookup.ts:480-486`), so `hold.id.isStaking` correctly picks up staking and excludes the session key deposit from `locked`.
- `lockedBalance` is `bnMax(...locks)` rather than a sum (`all.js:5-21`), so `lockedBalance + stakingHold` is sound — locks and holds are disjoint mechanisms (locks constrain `free` via `frozen`, holds sit in `reserved`), so there's no double count.
- Ran it locally: `tsc --noEmit` clean, `eslint --max-warnings 0` clean on the three changed files, unit tests 65/65 as claimed.

Five comments, one of which I'd like fixed before merge.

---

### 1. The code comment in `attestor/balance.ts` is wrong, and contradicts this PR's own description

```ts
// Reuse the shared helper rather than reading derive fields directly: attestor bonds go
// through StakingInterface, so they are staking bonds and now sit in a HoldReason::Staking
// hold. Reading `lockedBalance` here reported held stake as unlocked.
```

Attestor bonds don't go through a hold. `pallets/attestation/src/ledger.rs:76` calls `T::Currency::set_lock(BOND_LOCK_ID, ...)` — they are `LockableCurrency` locks under `b0ndl0ck`, so they land in `lockedBalance`/`frozen` and never appear in `balances.holds`. The Reviewer notes section of the description says this correctly ("the attestation pallet still uses its own legacy `b0ndl0ck` lock"), so the description and the comment disagree, and the comment is the wrong one.

This isn't pedantry — the comment misidentifies *why* the old code was wrong. The 100 CTC it printed **was** the attestor bond, reported correctly. What it missed was the separate 1,083,743 CTC pallet-staking hold, which is ordinary staking, not the attestor bond. Someone trusting this comment would reasonably conclude `lockedBalance` is now redundant and drop it from the `locked` computation — which would silently stop reporting the bond for a pure attestor who has no pallet-staking stake.

The same wording appears in the description under bug #2 ("Attestor bonds go through `StakingInterface`, so they are staking bonds and are held"), worth fixing there too so the squashed commit message isn't wrong.

### 2. The clamp's stated rationale doesn't match the pinned version, and its test asserts an impossible state

The doc comment and the test both say 15.x omits the outer `max(0, ..)` that 16.x added. But the CLI pins 16.5.2, where **both** fields are already clamped at the source (`all.js:34` for `transferable`, `:37` for `availableBalance`). So `BN.max` is unreachable, and `transferable` is a `Balance` (u128) that can't hold `-50` in the first place.

Defensive code is fine — the justification just isn't true for this repo, and `clamps a negative result to zero` reads as coverage while exercising a fixture the API cannot produce. Either re-word to something like "the derive already clamps in 16.x; belt-and-braces in case of a downgrade", or drop the clamp and that test case.

### 3. Hidden coupling worth a comment: this is only exact because `ExistentialDeposit = 0`

`runtime/src/lib.rs:349` sets `ExistentialDeposit: u128 = 0`, which is why the `maybeED` term in the formula is a no-op and the reported "after" equals `free` exactly. If ED ever becomes non-zero, `transferable` silently starts subtracting one ED and `canPay()` tightens accordingly. Arguably the correct behaviour, but it'd be an unpleasant surprise to debug later — worth a line in the `getTransferable` docblock.

### 4. Minor: redundant round trips in `show-stash-balance`

`getBalance()` calls `api.derive.staking.account()` to populate `bonded`/`unbonding`, which this command never reads — it takes TotalStake/ActiveStake from the precompile ledger instead. Two extra RPC calls per invocation. Consolidating on the shared helper is still the right call; just noting the cost.

### 5. Not blocking, but this change surfaces a real design question

With ED = 0, for the migrated attestor+staker case in the description (`frozen = 100` from the bond lock, `reserved = 1,083,743` from the staking hold), `frozen - reserved` is negative, so `transferable == free`. The attestor's 100 CTC bond restricts **nothing** about spendable balance — the staking reserve already satisfies the freeze.

That is correct `reducible_balance()` semantics and the CLI is right to mirror it. But it means the `b0ndl0ck` lock provides no spending restriction at all for anyone who also stakes. That's the concrete argument for the "should the attestation pallet move to holds" question raised in the Reviewer notes: if the bond is meant to be independently encumbering, a lock can't do that job once the account has reserves.

---

Two things done well: reverting the `total - transferable` approach was the right call, and the CI catch was a good one — sweeping in the 1000-planck proxy deposit would have been a subtle wrong answer that's easy to miss by eye. The "not covered" disclosure is honest too, though the gap is narrower than stated: Transferable/Locked/Total now share the `getBalance()` path that *is* verified, so the untested surface is really just the three table rows rendering.

One completeness note on the audit flagged at the end of the description: the only remaining in-repo `availableBalance` consumer is `cli/src/test/blockchain-tests/precompiles/substrate-transfer.test.ts:44,75`. Its destination is a freshly generated mnemonic account with no locks or reserves, so `availableBalance == transferable` there and it's unaffected — but it's the same class of consumer, if you want the sweep to be complete inside this repo.

DylanVerstraete (approved):

didac-gluwa (approved):

PR #1278, opened 2026-08-23
fix(write-ability): recover a rotation that resumed past its Outbox; add factory_scan_genesis_block

didac-gluwa:
## Why

`ea1cf11da` made a factory rotation resume the new factory's `OutboxCreated` scan from the previous factory's checkpoint instead of rescanning from genesis. That is a valid floor **only when the rotated-to factory was itself deployed at rotation time** — its own doc comment names the assumption and the failure mode.

Re-point a chain key at a **pre-existing** factory whose `OutboxCreated` already sits below the checkpoint and the resumed scan runs to the confirmed tip, matches nothing, and returns `Ok(None)`. The monitor reads that as "no Outbox" and holds the `⏸️` pause.

It does not clear. The cursor is persisted at the tip *against that very factory*, so on the next boot it **matches** and is read back as a valid checkpoint — the "just restart it" reflex entrenches the failure rather than fixing it. Recovery today means deleting `/data/factory-scan-cursor-{chain_key}.json` on every attestor by hand.

This is not hypothetical: the documented T4 rotation drill re-points key 8 between two long-existing factories, and violates the assumption on **both** legs. On usc-dev today the cursor sits at ~705,530 while Outbox B was created at 608,120 and Outbox A at 362,796 — gaps of ~97k and ~343k blocks.

## What

**1. A genesis fallback.** Track `scan_floor` — the block each factory's scan started from — and persist it beside the cursor. That is the missing input: a completed scan that found nothing is only conclusive when it began at the configured floor. Otherwise rewind to the floor and scan once more before reporting no Outbox.

- Bounded by `genesis_fallback_done`, reset on every factory change: **one rewind per factory per process**, so a factory that genuinely has no `OutboxCreated` costs one extra scan and then settles back into the cheap tip-following steady state.
- The rewind is persisted **before** the cursor write, so a restart mid-recovery continues rather than reloading the poisoned checkpoint.
- A record written before `scan_floor` existed loads with the floor set to `scanned_to`, not 0 — the conservative reading, which grants it exactly one fallback. Reading it as 0 would deny the recovery to precisely the cursors most likely to need it: one written by a rotation on an older build.

**2. `factory_scan_genesis_block`.** The block every *from-scratch* discovery scan starts at, replacing the literal 0 at initial activation, at a rotation with `resume_rotation_from_checkpoint` off, at de-registration, and in the rewind. Defaults to `0`, so behaviour is unchanged unless set.

It answers one question — *where does a scan with no usable position start?* It never moves a cursor that has one: a persisted scan, or a rotation resuming from a checkpoint, keeps its position even when that sits below this value. The one deliberate exception is the fallback, whose whole job is to rewind to it.

⚠️ Set it **above** a factory's `OutboxCreated` and that Outbox becomes undiscoverable, the fallback included. Documented in all three places it is configured.

## Notes for review

- **The fallback alone does not make a rotation fast, only recoverable.** The forward leg resumes, finds nothing, rewinds, and pays a full scan. Setting `factory_scan_genesis_block` near the factory deployment height is what makes it cheap.
- Naming: `factory_scan_genesis_block` over `writability_genesis_block` — it names the scan it bounds, matches `FactoryScanCursorStore` / `factory-scan-cursor-{chain_key}.json`, and does not read as a sibling of the unrelated `start_block` already in that section.
- The relayer half is https://github.com/gluwa/usc-message-relayer/pull/43 — same mechanism, same names, same defaults. Both are needed; either alone leaves the other component stranded on a rotation.
- **Not exercised against a live rotation.** Unit tests cover the decision function, the floor asymmetry, the persistence round-trip and the legacy-record path, but the end-to-end claim is still reasoning from source plus the live cursor heights. The falsification test is one rotation: expect `🔁 … resume_from=705…`, then `🪃 no OutboxCreated found above the resumed checkpoint`, then a rescan that resolves.

## Testing

`cargo test -p attestor` 133 pass · `cargo clippy -p attestor --all-targets --all-features -- -D warnings` clean · `cargo fmt` clean.

### DIFF
--- a/attestor/attestor/src/tasks/write_ability/cursor.rs
+++ b/attestor/attestor/src/tasks/write_ability/cursor.rs
@@ -22,7 +22,9 @@
 //! at all; a factory change (governance rotation) instead resumes from the checkpoint already on
 //! disk unless `Config::resume_rotation_from_checkpoint` is turned off — the same at-least-once
 //! safety applies either way, since re-scanning already-covered blocks only wastes time, never
-//! correctness.
+//! correctness. The record also carries the *floor* each scan began at, which is what lets the
+//! resolver's genesis fallback recover a scan that resumed above the event it was looking for
+//! (see [`super::resolver`]) even across a restart.
 
 use std::path::{Path, PathBuf};
 
@@ -205,6 +207,17 @@ struct FactoryScanRecord {
     found_outbox: Option<Address>,
     #[serde(skip_serializing_if = "Option::is_none")]
     found_created_at_block: Option<u64>,
+    /// Block this factory's scan *started* from — the floor below which it has never looked. Read
+    /// back by the resolver's genesis fallback to tell "scanned everything and this factory has no
+    /// Outbox" from "resumed above the event and would never have seen it".
+    ///
+    /// `Option` purely for backward compatibility with records written before this field existed:
+    /// [`FactoryScanCursorStore::load`] maps `None` to `scanned_to`, the conservative reading
+    /// ("assume the scan began where it currently sits"), which grants such a record exactly one
+    /// genesis fallback. Reading it as 0 instead would silently deny the fallback to precisely the
+    /// cursors most likely to need it — one written by a rotation on an older build.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    scan_floor: Option<u64>,
 }
 
 /// In-memory shape of a [`FactoryScanCursorStore`] record.
@@ -213,6 +226,8 @@ pub struct PersistedFactoryScan {
     pub factory: Address,
     pub scanned_to: u64,
     pub found: Option<(Address, Option<u64>)>,
+    /// Block this factory's scan started from; see [`FactoryScanRecord::scan_floor`].
+    pub scan_floor: u64,
 }
 
 /// Persists an [`super::resolver::OutboxDiscoveryCursor`]'s `OutboxCreated` discovery progress to a
@@ -277,6 +292,9 @@ impl FactoryScanCursorStore {
             found: record
                 .found_outbox
                 .map(|outbox| (outbox, record.found_created_at_block)),
+            // Pre-`scan_floor` record: assume the scan began where it now sits, so the resolver
+            // grants it one genesis fallback rather than trusting a floor it never recorded.
+            scan_floor: record.scan_floor.unwrap_or(record.scanned_to),
         })
     }
 
@@ -295,6 +313,7 @@ impl FactoryScanCursorStore {
             chain_key: self.chain_key,
             found_outbox,
             found_created_at_block,
+            scan_floor: Some(scan.scan_floor),
         };
         let json = serde_json::to_vec_pretty(&record).context("serialize factory-scan cursor")?;
 
@@ -433,6 +452,7 @@ mod tests {
             factory: factory(),
             scanned_to: 12_345,
             found: Some((outbox(), Some(100))),
+            scan_floor: 0,
         };
         store.save(&scan).unwrap();
         assert_eq!(store.load(), Some(scan));
@@ -442,6 +462,7 @@ mod tests {
             factory: factory(),
             scanned_to: 20_000,
             found: Some((outbox(), Some(100))),
+            scan_floor: 0,
         };
         store.save(&advanced).unwrap();
         assert_eq!(store.load(), Some(advanced));
@@ -457,9 +478,46 @@ mod tests {
             factory: factory(),
             scanned_to: 2_000,
             found: None,
+            scan_floor: 0,
         };
         store.save(&scan).unwrap();
         assert_eq!(store.load(), Some(scan));
+    }
+
+    /// `scan_floor` round-trips, so the resolver can tell "scanned everything below" from
+    /// "resumed above the event" across a restart — the distinction its genesis fallback turns on.
+    #[test]
+    fn factory_scan_round_trips_the_scan_floor() {
+        let dir = TmpDir::new("factory-floor");
+        let store = FactoryScanCursorStore::new(dir.path(), 2);
+        let scan = PersistedFactoryScan {
+            factory: factory(),
+            scanned_to: 705_600,
+            found: None,
+            scan_floor: 705_530,
+        };
+        store.save(&scan).unwrap();
+        assert_eq!(store.load(), Some(scan));
+    }
+
+    /// A record written before `scan_floor` existed loads with the floor set to `scanned_to`, not
+    /// to 0. That is the conservative reading — "assume the scan began where it now sits" — and it
+    /// grants such a record exactly one genesis fallback. Defaulting to 0 would instead claim the
+    /// range below was already covered, denying the recovery to precisely the cursors most likely
+    /// to need it: one written by a rotation on a build that predates this field.
+    #[test]
+    fn factory_scan_record_without_a_floor_defaults_to_its_position() {
+        let dir = TmpDir::new("factory-legacy");
+        let store = FactoryScanCursorStore::new(dir.path(), 2);
+        let legacy = format!(
+            r#"{{"scanned_to":705600,"factory":"{}","chain_key":2}}"#,
+            factory()
+        );
+        std::fs::write(store.path(), legacy).unwrap();
+        let loaded = store.load().expect("legacy record must still load");
+        assert_eq!(loaded.scanned_to, 705_600);
+        assert_eq!(loaded.scan_floor, 705_600);
+        assert_eq!(loaded.found, None);
     }
 
     #[test]
@@ -482,6 +540,7 @@ mod tests {
             factory: old_factory,
             scanned_to: 500,
             found: None,
+            scan_floor: 0,
         };
         store.save(&scan).unwrap();
         assert_eq!(store.load(), Some(scan));
@@ -496,6 +555,7 @@ mod tests {
             factory: factory(),
             scanned_to: 42,
             found: None,
+            scan_floor: 0,
         };
         store.save(&scan).unwrap();
         assert_eq!(store.load(), Some(scan));
--- a/attestor/attestor/src/tasks/write_ability/cursor.rs
+++ b/attestor/attestor/src/tasks/write_ability/cursor.rs
@@ -271,7 +271,7 @@ impl FactoryScanCursorStore {
             Err(e) => {
                 tracing::warn!(
                     path = %self.path.display(), %e,
-                    "could not read factory-scan cursor file; starting OutboxCreated discovery from genesis"
+                    "could not read factory-scan cursor file; starting OutboxCreated discovery from the configured factory-scan genesis block"
                 );
                 return None;
             }
--- a/attestor/attestor/src/tasks/write_ability/resolver.rs
+++ b/attestor/attestor/src/tasks/write_ability/resolver.rs
@@ -32,6 +32,13 @@
 //! factory and reads back as valid on the next boot. [`genesis_fallback_target`] closes that: a
 //! scan that completes with no match, having started above [`Config::factory_scan_genesis_block`],
 //! rewinds to that floor and tries once more before the chain key is declared Outbox-less.
+//!
+//! That fallback would otherwise fire on the *common*, healthy rotation too: a factory that was
+//! itself just deployed looks identical, for one tick, to a pre-existing one — both complete a
+//! scan above the floor and find nothing, because the fresh factory's own event has not finalized
+//! yet. [`OutboxDiscoveryCursor::grace_until_tip`] holds the fallback off until the finalized head
+//! has caught up to the tip observed when the floor was (re)armed, so a same-tick emission gets a
+//! chance to finalize before its absence is trusted as evidence.
 
 use alloy::primitives::{Address, B256};
 use alloy::providers::Provider;
@@ -84,21 +91,32 @@ pub struct ResolvedOutbox {
 /// cursor file on every pod by hand, and the recovery costs one extra scan of a range that was
 /// going to be scanned anyway had the checkpoint not been trusted.
 ///
-/// Three conditions, all necessary:
+/// Two conditions, both necessary:
 /// - `found_something` is false — a match makes the scan conclusive whatever floor it began at;
 /// - `scan_floor > genesis_block` — the scan started above the floor, so there is unlooked-at range
 ///   below it. A scan that already began at the floor has covered everything and its "no Outbox"
-///   verdict is the truth;
-/// - `already_used` is false — one rewind per factory per process, so a factory that genuinely has
-///   no `OutboxCreated` settles into the cheap tip-following steady state instead of rescanning the
-///   whole chain on every retry.
+///   verdict is the truth.
+///
+/// `grace_satisfied` guards against a race, not against re-firing: a rotation onto a factory that
+/// was itself *just* deployed looks identical, for one tick, to a rotation onto a pre-existing one
+/// whose event will never be found above the checkpoint — both complete a scan above the floor and
+/// find nothing, because the fresh factory's own event has not finalized yet. Trusting that first
+/// empty result would burn the one-shot recovery on the common, healthy rotation path this fallback
+/// is not meant to touch. The caller only passes `true` once the finalized tip has caught up to the
+/// height observed when the current floor was set, so a same-tick emission has had a chance to
+/// finalize — see [`OutboxDiscoveryCursor::grace_until_tip`].
+///
+/// Nothing here bounds this to *one* rewind per factory: once it fires, the caller lowers
+/// `scan_floor` to `genesis_block`, and every later call with the same (constant, per-process)
+/// `genesis_block` finds `scan_floor > genesis_block` already false — the floor itself carries the
+/// termination invariant.
 fn genesis_fallback_target(
     found_something: bool,
     scan_floor: u64,
     genesis_block: u64,
-    already_used: bool,
+    grace_satisfied: bool,
 ) -> Option<u64> {
-    (!found_something && !already_used && scan_floor > genesis_block).then_some(genesis_block)
+    (!found_something && scan_floor > genesis_block && grace_satisfied).then_some(genesis_block)
 }
 
 /// The scan-shaping knobs [`resolve_outbox_address`] reads out of [`Config`], grouped so the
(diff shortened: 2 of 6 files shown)


attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -348,13 +437,44 @@ async fn resolve_outbox_address<P: Provider>(
     // `safe_tip < cursor.from` (tip regressed / depth grew) — never regressing the cursor.
     cursor.from = cursor.from.max(from);
 
+    // Reaching here means the chunk loop ran to completion: an attempt cut short by the caller's
+    // RPC_ATTEMPT_TIMEOUT is aborted mid-`await`, and a failing chunk returns via `?`. So a `found`
+    // of `None` at this point is a *complete* scan of [`scan_floor`, `safe_tip`] that matched
+    // nothing — which is only the same thing as "this factory has no Outbox" when the scan began at
+    // the configured floor. Otherwise rewind once and let the next retry cover the rest.
+    if let Some(rewind_to) = genesis_fallback_target(
```
DylanVerstraete, line 535: **[verified: confirmed] Genesis fallback fires spuriously on a healthy fresh-factory rotation (latest-vs-finalized race).**

The factory registration is read at `latest` (the chain-info precompile `.call()` has no block tag) while the `OutboxCreated` scan is bounded at the `finalized` head (`safe_tip`). On a normal rotation to a *freshly deployed* factory, the first post-rotation resolve sees the new factory immediately (12s tick), completes a tiny `[checkpoint..safe_tip]` scan before the new factory's `OutboxCreated` finalizes, finds nothing — and `genesis_fallback_target` fires with no grace condition (`found=None`, `scan_floor=checkpoint > genesis_block`, `done=false`). Every attestor burns its one-shot fallback and starts a full-chain chunked getLogs rescan simultaneously, while the monitor holds `Pause`: the outage stretches from ~finality-lag to the full rescan duration on the *happy path* — the exact fleet-wide cost `resume_rotation_from_checkpoint` exists to avoid. Damage is bounded (the cursor never regresses, so the event is found once finalized or en route), but it turns every routine fresh-factory rotation into the multi-minute outage this PR fights on the pre-existing-factory path.

Suggested fix: require the empty verdict to survive a grace window before rewinding — e.g. only fall back once `safe_tip` has advanced past the block height at which the rotation was first observed, or after N consecutive empty *completed* scans. That keeps the T4 recovery (still empty after the grace) while skipping the rescan when the event simply has not finalized yet.

attestor/attestor/src/tasks/write_ability/cursor.rs
```diff
@@ -277,6 +292,9 @@ impl FactoryScanCursorStore {
             found: record
                 .found_outbox
                 .map(|outbox| (outbox, record.found_created_at_block)),
+            // Pre-`scan_floor` record: assume the scan began where it now sits, so the resolver
+            // grants it one genesis fallback rather than trusting a floor it never recorded.
+            scan_floor: record.scan_floor.unwrap_or(record.scanned_to),
```
DylanVerstraete, line 297: **[verified: confirmed] Legacy-record migration triggers a fleet-synchronized genesis rescan for every healthy pre-activation chain key on upgrade.**

`scan_floor: record.scan_floor.unwrap_or(record.scanned_to)` grants one genesis fallback to *every* pre-upgrade record with `found: None` — but `{found: None, scanned_to ≈ tip}` is also the normal persisted state of a chain key whose factory is registered while its Outbox has not been created yet (the old build persisted the cursor at the tip on every completed resolve, having genuinely scanned from 0). The on-disk bytes are identical between "rotation-stranded" and "healthy waiting-for-Outbox", so on the first fleet upgrade to this build every such attestor's first completed resolve fires the fallback and starts a genesis-to-tip getLogs sweep at the same moment — an upgrade-synchronized RPC storm that will read as an incident in the logs. Bounded to once per volume (the rewound floor persists), but entirely avoidable for the benign case.

Suggested mitigation: grant the legacy fallback only when it can matter — e.g. map a missing `scan_floor` to `scanned_to` only when `record.factory` differs from the live on-chain factory (the rotation branch re-arms the fallback for free anyway), else to 0; or at minimum jitter the first post-upgrade fallback across the fleet.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -348,13 +437,44 @@ async fn resolve_outbox_address<P: Provider>(
...
+    //
+    // Deliberately *after* the rewind above, so a fallback that has been decided is durable: a
+    // restart mid-recovery resumes the rescan instead of reloading the very cursor that stranded
+    // this chain key in the first place.
     if let Some(store) = factory_scan_store {
         let to_persist = PersistedFactoryScan {
             factory,
             scanned_to: cursor.from,
             found: cursor.found,
+            scan_floor: cursor.scan_floor,
```
DylanVerstraete: **[verified: confirmed] No intermediate persistence during the fallback rescan — a restart mid-recovery discards all headway.**

The save sits after the chunk loop, and the code's own comment concedes an attempt cut short by the caller's 30s `RPC_ATTEMPT_TIMEOUT` aborts mid-`await` — so nothing hits disk unless one attempt completes the whole remaining range within 30s. The moment the fallback fires, the record is pinned at `{scanned_to: genesis, scan_floor: genesis}`, and during the multi-attempt full-chain rescan (hundreds of 2000-block chunks on a long chain) no checkpoint is ever written; a restart reloads genesis and repeats the entire sweep. Under a crashloop or rolling redeploys shorter than the rescan, the recovery never converges. The persistence granularity predates this PR, but pre-PR a long-lived deployment never had a genesis-positioned cursor — the fallback newly manufactures exactly that state on the default config.

Suggested fix: persist inside the chunk loop (per completed chunk or every N chunks) — one small local JSON write per 2000-block getLogs is negligible next to the RPC call it checkpoints.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -246,13 +331,17 @@ async fn resolve_outbox_address<P: Provider>(
...
                 resume_from,
-                resume_rotation_from_checkpoint,
+                resume_rotation_from_checkpoint = policy.resume_rotation_from_checkpoint,
+                genesis_block = policy.genesis_block,
                 "🔁 Outbox factory rotated; restarting OutboxCreated discovery"
             );
         }
         cursor.factory = Some(factory);
         cursor.from = resume_from;
+        cursor.scan_floor = resume_from;
```
DylanVerstraete: **[verified: confirmed] The load-bearing line of the T4 fix has no test driving it through the real rotation path.**

`cursor.scan_floor = resume_from` is what arms the fallback after a rotation, but no test exercises it: `resolve_outbox_address` needs a live `Provider` and has no mock, the fallback tests call `genesis_fallback_target` directly or assign cursor fields by hand, and the two pre-existing rotation tests (`cursor_carries_a_discovery_past_an_interrupted_scan`, `factory_rotation_resumes_from_checkpoint_when_enabled`) replicate the rotation reset *inline* — and their hand-copies already drifted: they set only `from`/`factory`/`found` and pad the literal with `scan_floor: 0`, without mirroring the new `scan_floor = resume_from` / `genesis_fallback_done = false` resets. A future refactor that drops or misorders this line keeps every test green while silently regressing the exact bug this PR fixes.

Suggested fix: extract the rotation reset into a method on `OutboxDiscoveryCursor` (e.g. `fn rotate_to(&mut self, factory: Address, resume_from: u64)`), call it from both production and the rotation tests, and assert `scan_floor`/`genesis_fallback_done` post-rotation.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -368,7 +488,13 @@ async fn resolve_outbox_address<P: Provider>(
         );
         return Ok(Some((outbox, created_at_block)));
     }
-    tracing::warn!(%factory, chain_key, "factory has emitted no OutboxCreated for chain_key yet");
+    tracing::warn!(
+        %factory,
+        chain_key,
+        scanned_from = cursor.scan_floor,
+        scanned_to = cursor.from,
+        "factory has emitted no OutboxCreated for chain_key yet"
```
DylanVerstraete: **[verified: plausible] A misconfigured `factory_scan_genesis_block` strands the chain key with no diagnosable signal.**

The docs warn (in three places) that a floor above the factory's `OutboxCreated` makes the Outbox permanently undiscoverable — and the fallback is disabled by the same misconfig, since `scan_floor == genesis_block` always declines. Boot-time validation is not implementable (`validate()` has no provider and every u64 is valid for some chain), but the failure is currently indistinguishable from the benign "Outbox not created yet" state: this terminal warn fires forever with no causation hint, and an operator has no reason to suspect the floor.

Suggested mitigation: when this warn fires with `policy.genesis_block > 0` (a complete from-floor scan found nothing and the fallback declined), extend the message to flag that a `factory_scan_genesis_block` set above the factory's `OutboxCreated` height would produce exactly this symptom — the one observable moment the misconfig manifests.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -67,15 +75,55 @@ pub struct ResolvedOutbox {
...
+    /// See [`Config::factory_scan_genesis_block`] — the floor every from-scratch scan starts at.
+    genesis_block: u64,
+}
+
 /// Caller-owned state for the incremental Outbox-discovery scan across activation retries. Tracks
 /// how far we have scanned and against which factory, so each retry only scans new *confirmed*
 /// blocks (no full-history log storm) while staying reorg-safe, and a factory re-registration
-/// restarts the scan from genesis.
+/// restarts the scan from the configured floor.
 #[derive(Default)]
```
DylanVerstraete: **[cleanup] `#[derive(Default)]` is now a footgun with zero call sites — remove it.**

`starting_at()`'s own doc says `default()` "starts at block 0 regardless, which is only correct when the floor is 0", and no production or test call site of `OutboxDiscoveryCursor::default()` remains in the crate. Keeping a derived constructor that is documented as usually-wrong invites the next caller to reach for the natural idiom and silently pin the floor to 0, bypassing `factory_scan_genesis_block` with no compiler or log signal. Removing the derive makes the compiler enforce the new invariant; `starting_at(0)` is a drop-in wherever the old default was wanted.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -87,9 +135,21 @@ pub struct OutboxDiscoveryCursor {
...
+    /// at for this factory, which is exactly what the genesis fallback needs to know: a completed
+    /// scan that found nothing is only conclusive when this equals the configured floor.
+    ///
+    /// Moves with `from` whenever a scan (re)starts — initial seed, rotation, de-registration,
+    /// fallback — and is persisted so the distinction survives a restart.
+    scan_floor: u64,
+    /// Whether the genesis fallback has already been spent on the current factory. Reset alongside
+    /// `from`/`found` on every factory change, so each factory gets exactly one — bounding the
+    /// recovery to a single extra full scan instead of one per retry.
+    genesis_fallback_done: bool,
```
DylanVerstraete: **[cleanup] `genesis_fallback_done` is redundant state — the persisted floor already carries the termination invariant.**

The flag is only ever set `true` in the same block that sets `scan_floor = rewind_to` (= `policy.genesis_block`), and every other transition (rotation, de-registration, both constructors) resets both together — so whenever `already_used` is true, `scan_floor > genesis_block` is already false and the flag never decides anything (config is constant per process; restarts reset it anyway). Meanwhile its doc claims it "bounds the recovery to a single extra full scan", which is actually enforced by the floor — a future change that rewinds `from` or lowers `scan_floor` independently could lean on this process-local flag for a bound it does not provide. Deleting the field and the `already_used` parameter (letting the floor invariant carry termination) removes one field from every constructor/test literal, and pairs naturally with extracting the thrice-duplicated reset triple (`from`/`scan_floor`/`found`/flag at de-registration, rotation, and the constructors) into one helper.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -232,10 +313,14 @@ async fn resolve_outbox_address<P: Provider>(
...
+        // start?", and a checkpoint is a position; raising it here would skip range the previous
+        // scan had already proven worth covering. Only the from-scratch branch takes the floor.
+        let resume_from = if policy.resume_rotation_from_checkpoint {
             cursor.from
         } else {
-            0
+            policy.genesis_block
         };
         if previous_factory.is_some() {
             // Not fired on the very first resolve (no `previous_factory` yet) — that is initial
```
DylanVerstraete, line 413: **[doc] Stale comment: `resume_from` is no longer "0 either way" for a fresh cursor.**

A fresh cursor is now `starting_at(cfg.factory_scan_genesis_block)`, so on the very first resolve `resume_from` equals the configured floor, not 0 — and a `from_persisted` cursor whose on-disk factory rotated away while the process was down also takes this branch with `resume_from` = the persisted checkpoint. Same vintage: `cursor.rs`'s unreadable-file warn still says "starting OutboxCreated discovery from genesis" where discovery now restarts from the configured floor. Both sit directly on the logic this PR changed and would misdirect someone debugging a nonzero-floor deployment.

attestor/attestor/src/tasks/write_ability/resolver.rs
didac-gluwa: Confirmed, fixed in 36bd82c48: the fallback now waits for the finalized tip to catch up to the tip observed at rotation time before firing (new `grace_until_tip` on the cursor).

attestor/attestor/src/tasks/write_ability/cursor.rs
didac-gluwa: Mitigated in 36bd82c48: a cursor reloaded from disk now defers arming that same grace window to the first live call, so it can't fire synchronously on the very next tick after a restart/upgrade.

attestor/attestor/src/tasks/write_ability/resolver.rs
didac-gluwa: Fixed in 36bd82c48: progress is now persisted after every chunk during the rescan, not only at the end of the call.

attestor/attestor/src/tasks/write_ability/resolver.rs
didac-gluwa: Fixed in 36bd82c48: extracted `OutboxDiscoveryCursor::rotate_to()` so production code and tests (incl. a new dedicated one) exercise the same rotation reset.

attestor/attestor/src/tasks/write_ability/resolver.rs
didac-gluwa: Addressed in 36bd82c48: the terminal "no Outbox" warn now hints at a misconfigured `factory_scan_genesis_block` when the scan has exhausted the floor.

attestor/attestor/src/tasks/write_ability/resolver.rs
didac-gluwa: Removed in 36bd82c48.

attestor/attestor/src/tasks/write_ability/resolver.rs
didac-gluwa: Removed in 36bd82c48 — the `scan_floor`/`genesis_block` comparison alone already carries the termination invariant.

attestor/attestor/src/tasks/write_ability/resolver.rs
didac-gluwa: Fixed in 36bd82c48.

attestor/attestor/src/tasks/write_ability/resolver.rs
DylanVerstraete: Verified in 36bd82c48: `grace_until_tip` armed from the rotation-observing tip via `rotate_to`, and the fallback declines until the finalized head passes it. Resolving.

attestor/attestor/src/tasks/write_ability/cursor.rs
DylanVerstraete: Verified: `from_persisted` defers grace-arming to the first live call, so a reloaded idle cursor cannot fire the fallback on the first post-restart tick. Note this delays (by the grace window) rather than eliminates the once-per-volume rescan for idle chain keys — bounded and acceptable, matching the relayer-side decision. Resolving.

attestor/attestor/src/tasks/write_ability/resolver.rs
DylanVerstraete: Verified in 36bd82c48: `persist_progress` closure now checkpoints per-chunk inside the scan loop (best-effort, non-fatal on write failure) — a restart mid-rescan resumes instead of repaying the sweep. Resolving.

attestor/attestor/src/tasks/write_ability/resolver.rs
DylanVerstraete: Verified in 36bd82c48: production and tests both route through `OutboxDiscoveryCursor::rotate_to`, with `the_load_bearing_rotation_reset_is_exercised_through_rotate_to` covering exactly the reset that was untested. Resolving.

attestor/attestor/src/tasks/write_ability/resolver.rs
DylanVerstraete: Verified in 36bd82c48: the terminal warn now reports `factory_scan_genesis_block` and tells on-call to check whether the floor sits above the real OutboxCreated height. Resolving.

attestor/attestor/src/tasks/write_ability/resolver.rs
DylanVerstraete: Verified: derive removed in 36bd82c48. Resolving.

DylanVerstraete: Verified: derive removed in 36bd82c48. Resolving.

attestor/attestor/src/tasks/write_ability/resolver.rs
DylanVerstraete: Verified: `genesis_fallback_done` fully removed in 36bd82c48; the floor comparison alone carries the termination invariant, with `grace_until_tip` covering the rotation race independently. Resolving.

attestor/attestor/src/tasks/write_ability/resolver.rs
DylanVerstraete: Verified: stale comment gone in 36bd82c48. Resolving.

DylanVerstraete (approved):

PR #1280, opened 2026-08-24
fix(audit): chain-info bucket-walk false negative, scan cap, and observable force_election randomness

DylanVerstraete:

Lands the two chain-info audit fixes from #1108 on current `usc-dev`, plus a review follow-up that bounds the checkpoint walks at the attestation genesis.

## Why a new branch

#1108 targets `usc-dev` but was cut from the pre-rewrite lineage, so its merge base is `77693c71` and GitHub renders it as **738 files, +155k/-37k**, `mergeable: CONFLICTING`. The real work is a handful of commits at the tip; everything else is history divergence. The substantive commits cherry-pick cleanly onto `usc-dev`; original authorship is preserved.

## Audit fixes (from #1108)

**#18 (Medium) — chain-info sparse-checkpoint false negative.** `BUCKET_SEARCH_ATTEMPTS` was hardcoded to `5`, so `find_highest_attested_before` / `find_lowest_attested_after` gave up after 5 buckets (5000 blocks) and returned a false negative when valid checkpoints existed further away. Replaced with a bound derived from the real search span, applied symmetrically to the down-walk, up-walk, and `is_height_attested`.

**#21 residual (Medium) — unbounded `Attestations::iter_prefix` scans.** Adds a defensive `MAX_ATTESTATION_SCAN_ITEMS = 100_000` cap via `.take(...)` on all 6 prefix scans, placed before `.inspect` so gas still matches items actually scanned.

## Review follow-up (8ca86c903, BradleyOlson64)

The two downward walks (`find_highest_attested_before`, `is_height_attested`) now stop at the chain's `AttestationChainGenesisBlockNumber` instead of block 0: checkpoints cannot exist below the attestation genesis, so every bucket between it and block 0 was known-empty and only charged the caller `GAS_STORAGE_LOOKUP` per bucket. One extra lookup reads the floor; the attempt bound becomes `target - genesis` and the loop breaks once the genesis bucket has been visited (inclusive: a checkpoint exactly at the genesis is still found). Tests pin the walk length by exact cost. Also fixes the inverted comment on the attestations-scan branch.

## Dropped from #1108: #26 (`force_election` randomness)

#1108 also threaded real epoch randomness into `force_election`, which previously passed a hardcoded `[0; 32]`. That code path no longer exists on `usc-dev`: #1326 made `force_election` per chain (`force_election(chain_key)`), it labels the election with `CurrentEpochIndex` and runs `elect_attestors_for_chain`, and no randomness is passed or consumed anywhere in the selection (sortition is unbuilt, RFC-0174). Re-adding a `RandomnessProvider`, an event field and a provider-trait split for a seed nothing reads would be dead weight, so those commits were dropped here. When randomness-dependent selection is built, the "make the seed observable in `ForcedElection`" idea from the original follow-up is the right shape to revisit.

The stale `gluwa-bot` metadata commits were dropped too; the branch no longer changes the runtime, so there is nothing to regenerate.

## Notes for reviewers

- **Behaviour change worth an explicit sign-off:** removing the bucket-walk ceiling means a deep walk over a sparse store now reverts on out-of-gas instead of cheaply returning "not attested". With `CHECKPOINT_BUCKET_SIZE = 1000` and `GAS_STORAGE_LOOKUP = 2600` that is ~11.5M blocks of affordable span at a 30M gas limit, and the genesis floor now removes the known-empty part of that span. The old 5-bucket ceiling was unambiguously a bug, so the direction is right — it is the failure *mode* that changed, and EVM callers need to handle a revert where they previously got `false`. atodorov's review on #1108 predates the two follow-up commits that introduced this.
- No runtime, pallet or metadata changes: the diff is confined to `precompiles/chain-info`.

## Verification

- `cargo fmt --all -- --check`: clean
- `cargo clippy -p pallet-evm-precompile-chain-info --all-targets -- -D warnings`: clean
- `cargo test -p pallet-evm-precompile-chain-info`: **46** passing (43 + 3 new), no pre-existing tests broken

Closes #1108 once merged.

---

When merging this PR:


didac-gluwa (dismissed):

precompiles/chain-info/src/lib.rs
```diff
@@ -313,6 +354,9 @@ where
 
             // If the target height is lower than the last checkpoint height, we first search through the attestations directly.
```
BradleyOlson64, line 355:
```suggestion
            // If the target height is higher than the last checkpoint height, we first search through the attestations directly.
```

precompiles/chain-info/src/lib.rs
BradleyOlson64, line 314: We can add to the filter here to ensure that the bucket pivot is above

`AttestationChainGenesisBlockNumber`

This prevents waste in checking buckets back to block 0 if AttestationChainGenesisBlockNumber is higher.

precompiles/chain-info/src/lib.rs
DylanVerstraete: Applied in 8ca86c903 — the comment now reads "higher than the last checkpoint height", which is what that branch actually handles.

precompiles/chain-info/src/lib.rs
DylanVerstraete: Done in 8ca86c903, for both downward walks (`find_highest_attested_before` and `is_height_attested`): they read `AttestationChainGenesisBlockNumber` once (one extra `GAS_STORAGE_LOOKUP`), bound the attempt count by `target - genesis` instead of `target`, and break once the genesis bucket has been visited. The floor is inclusive, so a checkpoint exactly at the genesis is still found. Tests pin the walk length by exact cost: with genesis 50_000 and target 60_000 both precompiles now charge 3 fixed lookups + 10 buckets, where they previously walked 60. Also worth knowing while you re-review: the branch was rebased onto current usc-dev and the #26 `force_election` randomness commits were dropped, because #1326 made `force_election` per chain and it no longer passes or consumes a seed; the PR description explains.

PR #1292, opened 2026-08-27
feat(write-ability): discovery-address getter on chain-info + registry-first resolution

didac-gluwa:
## Summary

This is the PR Dylan asked for in Slack (26 Aug): *"Move both resolvers off newest-log binding onto the registry. Needs a discovery-address getter on the chain-info precompile beside `get_outbox_factory_address`."* — and reinstated on asc-contracts#38 after usc-message-relayer#45 shipped an interim, config-driven workaround that deliberately avoided a runtime change.

Four pieces:

1. **`get_outbox_discovery_address(uint64)`** on the chain-info precompile (`0xFD3`), mirroring `get_outbox_factory_address` exactly: new `OutboxDiscoveries` storage map in `pallet-supported-chains`, an `OperatorsOrigin`-gated `set_outbox_discovery_addr` extrinsic, genesis seeding, provider/runtime-API method (`SupportedChainsApi` bumped to v3), Solidity interface + regenerated ABI (verified against the CI drift gate).
2. **Regenerated subxt/CLI metadata** from a node built with (1).
3. **The attestor's own `resolver.rs`** initially tried the discovery registry first with a fall-through to the factory scan — **superseded by (4) below**, which removes that fallback entirely.
4. **Registry-only rescope (8 Sep)**: per Dylan's Slack confirmation (2 Sep, CKR3180JY) — *"the OutboxDiscovery through the precompile is the expected way of resolving the outbox, all other forms are deprecated and should be removed"* — `OutboxDiscovery.defaultOutbox(chainKey)` is now the **only** resolution path. Deletes the `OutboxCreated` log-scan, `FactoryScanCursorStore`/`OutboxDiscoveryCursor`, the genesis-fallback/grace logic, and the config knobs that only existed to tune them (`resume_rotation_from_checkpoint`, `factory_scan_genesis_block`). The mirrored usc-message-relayer#50 carries the same rescope for the relayer.

## Why

`OutboxFactory.deployOutbox` is intentionally permissionless, so an attacker can deploy an Outbox for any chain key and emit an `OutboxCreated` indistinguishable from a legitimate one — a resolver trusting "newest log wins" can be pointed at a contract it doesn't control. asc-contracts#38 closes the write side (the registry is only ever written through an access-controlled deploy path). This PR closes the read side for the attestor, and (4) removes the deprecated log-scan surface entirely rather than leaving it live as an unused-but-present attack path.

## Design note

Bumping `SupportedChainsApi`'s trait-level `api_version` to 3 required dropping the old `#[api_version(2)]` tags on `write_ability_config`/`outbox_factory_address` — `sp_api` forbids a method version below the trait's own declared version, and this crate only ever implements the single current API (no versioned client-side aliasing), so it costs no behavior: a caller's `api_version() >= 2` check still passes once the runtime reports v3.

## ⚠️ Do not deploy yet

Per Dylan's review (8 Sep): resolution is now registry-only and fail-closed. An attestor built from this branch will not activate write-ability on any network until **all three** of the following exist there:

- (a) a runtime carrying this pallet storage + the `get_outbox_discovery_address` precompile getter (Dylan)
- (b) `OutboxDiscovery` deployed with the network's Outbox registered as `defaultOutbox(chainKey)` (asc-contracts side)
- (c) `set_outbox_discovery_addr(chainKey, discovery)` called by sudo/operator (Dylan)

None of the three exist on usc-devnet today. Dylan is tracking (a) and (c) on the cutover plan; do not roll an image from `writeability-off-usc-dev` past this commit to any live network before they land.

## Not yet done

- Not exercised end-to-end on any real network — see "Do not deploy yet" above.
- `e2e_anvil.rs`'s own Rust-level unit test still dodges `resolve()` with a TODO (builds `ResolvedOutbox` by hand) — bare anvil has no chain-info precompile to exercise against. Separate from the CI e2e fix below; not blocking.

## Fixed since Dylan's 8 Sep review

- ~~The write-ability e2e cannot resolve an Outbox~~: `deploy-source-ethers.mts` now also deploys `ChainRegistry` + `OutboxDiscovery` (behind its UUPS proxy), registers the Outbox as the chain key's default, and calls `set_outbox_discovery_addr` via sudo — mirroring the existing `set_outbox_factory_addr` registration. Verified locally end to end against a live `--dev` node + anvil: after this wiring, `chain-info.get_outbox_discovery_address(2)` returns the deployed registry (`exists=true`) and its `defaultOutbox(2)` returns the deployed Outbox — the exact path the attestor's resolver reads.
- `OutboxDiscoveryRegistered`/`ZeroOutboxDiscoveryAddress` enum-index nit and the "do not deploy yet" PR-body note (this section) are both addressed.

## Verified

- `cargo check`/`clippy --all-targets`/`fmt --check` clean for `attestor`, `write-ability`, `pallet-supported-chains`, `cc-client` (+ the `abi_surface` ABI-drift gate).
- 118 attestor + 17 write-ability + 52 pallet unit tests pass; `e2e_anvil.rs` compiles; CLI `tsc --noEmit` and `prettier --check` clean.
- `OutboxDiscoveryRegistered`/`ZeroOutboxDiscoveryAddress` moved to the end of their enums (Dylan + Bugbot: avoid shifting existing SCALE indices) — `subxt metadata` + the CLI's `polkadot-types-from-chain` regenerated from a freshly built node against the corrected order.
- e2e OutboxDiscovery wiring run live end-to-end (see "Fixed since Dylan's 8 Sep review" above) against a freshly built node, a real anvil destination chain, and `asc-contracts@434073943` compiled locally — not just read through.
- Rebased onto current `writeability-off-usc-dev` (picking up #1304/#1317's provider-rebuild-in-place fixes and #1316's `IOutboxDiscovery` ABI mirror) with conflicts resolved by hand; `abi-creator.sh` / `generate-metadata-json.sh` / `check-solidity-source-vs-metadata.sh` unaffected (no Solidity interface changed since the last run).


didac-gluwa:
Updated per today's Slack thread ([here](https://gluwa.slack.com/archives/CKR3180JY/p1787934834927019?thread_ts=1787668964.814519)): Kevin confirmed \`OutboxDiscovery\` is the source of truth and resolvers should call \`defaultOutbox(chainKey)\` directly, not \`outboxOf\` via \`OutboxDeployer\` — a chain key can have multiple active Outbox versions, which \`outboxOf\` doesn't support. Renamed the local \`IOutboxDeployer\` mirror to \`IOutboxDiscovery\` and switched \`resolve_outbox_from_registry\` to call \`defaultOutbox\`. No behavior regression — this code path isn't live anywhere yet.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -314,10 +314,69 @@ pub async fn resolve<P: Provider>(
...
+/// so a chain key not yet migrated onto the registry keeps resolving exactly as before.
+async fn resolve_outbox_from_registry<P: Provider>(
+    provider: &P,
+    chain_key: ChainKey,
+) -> Result<Option<Address>> {
+    let discovery = IChainInfo::new(CHAIN_INFO_PRECOMPILE, provider)
+        .get_outbox_discovery_address(chain_key)
+        .call()
+        .await
+        .context("chain-info precompile get_outbox_discovery_address() reverted")?;
```
DylanVerstraete, line 106: **Deploy-order hazard: this unconditional new-selector read turns a working resolver into a crash loop.**

`get_outbox_discovery_address(uint64)` only exists on a runtime that carries this PR. Any attestor built from this branch but pointed at a node whose runtime predates it (the normal attestor-before-runtime deploy order, and the exact skew the existing code already warns about for `get_outbox_factory_address`) gets an `unknown selector` revert here. Because the error is propagated with `?`, `resolve_outbox_address` never reaches step 1, so the *pre-existing, still-working* factory path is skipped: the activation loop counts a consecutive failure (`advanced_last_call()` is false), and after `MAX_CONSECUTIVE_RESOLVE_FAILURES` (~5 min) `run` returns `Err` and the write-ability task restarts — forever. Write-ability goes from working to permanently down on a node that has not been upgraded yet.

A revert of *this* selector is provably equivalent to "no registry can be registered" (the storage map doesn't exist on that runtime), so it should degrade to `Ok(None)` and fall through, with a warning — matching the fall-through this function's own doc comment promises. Keep hard failures for transport errors if you want, but not for the revert.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -314,10 +314,69 @@ pub async fn resolve<P: Provider>(
...
+            "chain_key {chain_key} exceeds the uint32 the Outbox registry is keyed by, so it \
+             cannot be represented on-chain"
+        )
+    })?;
+
+    let outbox = IOutboxDiscovery::new(discovery, provider)
+        .defaultOutbox(chain_key_u32)
+        .call()
+        .await
+        .with_context(|| format!("defaultOutbox reverted on registry {discovery}"))?;
```
DylanVerstraete, line 126: **`defaultOutbox` is the only interface tried, but the address stored in `OutboxDiscoveries` carries no discriminator for which registry it is — and a mismatch is fatal, not a fall-through.**

The pallet extrinsic doc, the precompile doc and the `.sol` comment added in this same PR all describe the stored address as "the `OutboxDeployer`/`OutboxDiscovery` ... whose `outboxOf`/`defaultOutbox` getter", and the relayer side (usc-message-relayer, `write-ability/src/abi.rs`) resolves via `outboxOf(uint32)` on `OutboxDeployer` with the interface chosen by config. `OutboxDeployer` has no `defaultOutbox`, so if an operator registers a deployer address (which every doc in this PR says is a valid thing to put in this map), this call reverts on every tick. A non-contract / wrong-network address is the same story: empty returndata → alloy decode error.

Both cases propagate `Err` out of `resolve_outbox_address`, so the factory path is never reached — the resolver then fails ~25 times in a row and the write-ability task returns `Err` and restarts in a loop. That directly contradicts this function's doc claim that a registry problem "falls through unchanged to the existing factory + `OutboxCreated` scan".

Suggest either (a) treating a revert/decode failure from the registry as `Ok(None)` + `warn!` and continuing to the factory scan, or (b) trying `defaultOutbox` and then `outboxOf` before giving up. Note there is also no test — unit or e2e — covering this new path, so the `defaultOutbox(uint32)` selector and the `uint32` key width are still unverified guesses.

attestor/attestor/src/tasks/write_ability/resolver.rs
```diff
@@ -335,6 +394,18 @@ async fn resolve_outbox_address<P: Provider>(
...
+    // 0. Registry read, when this chain key has one — see `resolve_outbox_from_registry`. No cursor
+    //    bookkeeping needed: unlike the log scan, a registry read is complete and authoritative on
+    //    every call, so there is nothing to resume or persist.
+    if let Some(outbox) = resolve_outbox_from_registry(provider, chain_key).await? {
+        tracing::info!(
+            %outbox,
+            chain_key,
+            "🧭 resolved Outbox on-chain (discovery registry)"
+        );
+        return Ok(Some((outbox, None)));
```
DylanVerstraete: **Registry-resolved Outboxes always report `created_at_block: None`, which silently re-opens the message-skipping hole the rotation swap was written to close.**

`mod.rs:869` does `let swap_start = resolved.created_at_block.or(scan_from);` and `scan_from` is `cfg.start_block.or(Some(head_before_resolve))` — the chain head at boot. Its comment spells out why the creation block matters: "governance can point a chain key at an Outbox created *before* this process booted — whose earlier `MessagePublished` events would then sit below `scan_from` and never be scanned."

That is precisely the case `setDefaultOutbox` exists for (the Slack quote in the doc above: "there will be multiple version[s] of outbox"). So: attestor boots at head H, later governance points `defaultOutbox(chainKey)` at Outbox B that was deployed at block H-5000 and already has messages, the monitor swaps to B, and the replacement listener starts at H — every `MessagePublished` on B below H is never observed and never voted on.

The registry path can recover a start height cheaply (e.g. filter B's own first `MessagePublished`/an `OutboxCreated` for B, or fall back to `policy.genesis_block` instead of `scan_from` when the address came from the registry) rather than returning `None`.

pallets/supported-chains/src/lib.rs
```diff
@@ -237,6 +266,14 @@ pub mod pallet {
             outbox_factory_addr: H160,
         },
 
+        /// The Outbox discovery-registry address for a supported chain has been registered.
+        /// This signals to attestors/relayers that they can resolve the Outbox for this chain
+        /// key from the registry instead of scanning the factory's `OutboxCreated` logs.
+        OutboxDiscoveryRegistered {
```
DylanVerstraete: **Inserting an event variant in the middle of the enum shifts every following event index.**

SCALE event indices come from declaration order, so this insertion moves `WriteAbilityConfigSet` 3 → 4 and `CoreFeeSet` 4 → 5. The call side was carefully kept stable (`#[pallet::call_index(5)]` below, out of order on purpose) — the event enum deserves the same treatment.

Concrete breakage: any consumer decoding `SupportedChains` events against metadata that isn't this exact build — the relayer's pinned subxt metadata, an indexer replaying history, a cached `metadata.scale` in another repo — will read a post-upgrade `WriteAbilityConfigSet` as `CoreFeeSet` (or, replaying pre-upgrade blocks with new metadata, read an old `WriteAbilityConfigSet` as `OutboxDiscoveryRegistered`). The field shapes differ (`chain_key + [u8;32] + bool` vs `chain_key + H160`), so it is either a hard decode error or silently wrong data.

Appending `OutboxDiscoveryRegistered` after `CoreFeeSet` costs nothing and keeps all existing indices.

pallets/supported-chains/src/lib.rs
```diff
@@ -271,6 +308,12 @@ pub mod pallet {
         /// rejected to fail loudly instead of silently disabling write-ability for the chain.
         ZeroOutboxFactoryAddress,
 
+        /// The Outbox discovery-registry address is the zero address. A zero registry cannot be
+        /// resolved by the attestor/relayer (it reads as "not registered"), so setting it via the
+        /// operator path is rejected to fail loudly instead of silently disabling registry-based
+        /// resolution for the chain.
+        ZeroOutboxDiscoveryAddress,
```
DylanVerstraete, line 329: **Same index-shift problem on the error enum: this insertion moves `ZeroWriteAbilityChainKey` from error index 5 to 6.**

`DispatchError::Module { index, error }` carries the pallet-error byte, so after this change a rejected `set_write_ability_config` with a zero key reports byte 6 while any client/dashboard/test holding the old metadata (or an operator reading an older error table) still maps byte 5 → `ZeroWriteAbilityChainKey` and byte 6 → out of range. Appending `ZeroOutboxDiscoveryAddress` after `ZeroWriteAbilityChainKey` avoids renumbering an existing error.

pallets/supported-chains/src/weights.rs
```diff
@@ -166,6 +166,33 @@ impl<T: frame_system::Config> crate::WeightInfo for WeightInfo<T> {
...
+	/// Storage: `System::Events` (r:1 w:1)
+	/// Proof: `System::Events` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
+	/// Storage: `SupportedChains::OutboxDiscoveries` (r:0 w:1)
+	/// Proof: `SupportedChains::OutboxDiscoveries` (`max_values`: None, `max_size`: None, mode: `Measured`)
+	//
+	// Not benchmarked (no `creditcoin3-node` binary available in this environment) — a copy of
+	// `set_outbox_factory_addr`'s weight, since the storage shape (one `SupportedChains` read plus
+	// a single map write) is identical. Re-run `benchmark pallet` for this extrinsic before
+	// shipping to a network where weight accuracy is enforced.
+	fn set_outbox_discovery_addr() -> Weight {
```
DylanVerstraete, line 186: **Hand-edited generated weights file: the new weight is a copy and the surrounding auto-generated storage docs are now attached to the wrong function.**

Two things here:

1. The block of `/// Storage: System::Number / ExecutionPhase / EventCount / Events` lines that used to precede `set_write_ability_config` now precedes this new function (and is duplicated), leaving `set_write_ability_config` documented with only `SupportedChains::WriteAbilityConfigs (r:0 w:1)`. Anyone reading this file for the proof size of `set_write_ability_config` gets the wrong picture, and the next real `benchmark pallet` run will produce a confusing diff.
2. `set_outbox_discovery_addr` ships an unmeasured weight copied from `set_outbox_factory_addr`. The storage shape does match, so the number is plausible — but it is unbenchmarked on a governance extrinsic, and the note admitting that lives only in a comment. Worth running the benchmark before this goes to a network with weight enforcement (as the comment itself asks).

DylanVerstraete (changes requested):
The rewrite is the right shape: `defaultOutbox` on the OutboxDiscovery registry, nothing else, log-scan and genesis fallback gone. Kevin's 28 Aug quote is the source of truth and the code follows it. Two of my 31 Aug threads are addressed (defaultOutbox, regenerated weights); resolving those. What still blocks merge is not the code, it is that the trunk's own tests and deploy can no longer run without a registry:

**Must fix before merge**
1. **The write-ability e2e cannot resolve an Outbox any more.** Resolution is registry-only and fail-closed, but `.github/workflows/write-ability-e2e.yml` and `usc-messaging/scripts/deploy-source-ethers.mts` are untouched: nothing deploys `OutboxDiscovery`/`OutboxDeployer`, registers the Outbox as default, or calls `set_outbox_discovery_addr(2, …)` via sudo on the dev node. The attestors will sit in the resolve-retry loop and "Assert message was delivered" times out. `e2e_anvil.rs` already TODO-skips `resolve` for the same reason. Please add the discovery deploy + default registration + sudo call to the e2e in this PR so the trunk keeps a working e2e.

**Sequencing, so nobody rolls an image from this trunk too early**
Once merged, an attestor built from writeability-off-usc-dev cannot vote on usc-devnet until three things exist there: (a) a runtime with this pallet storage + precompile getter, (b) `OutboxDiscovery` deployed with Outbox 0x95FE… registered as `defaultOutbox(8)`, (c) `set_outbox_discovery_addr(8, discovery)` by sudo. None of the three exist today. I will put those three steps on the cutover plan and own (a) and (c); please state in the PR body that images from this branch are do-not-deploy until then.

**Nits, not blocking**
- `OutboxDiscoveryRegistered` / `ZeroOutboxDiscoveryAddress` are inserted mid-enum and shift `WriteAbilityConfigSet`, `CoreFeeSet`, `ZeroWriteAbilityChainKey` indices. Appending at the end is a one-line move and avoids the whole class of stale-metadata decoding bugs.
- Rotation swap now starts at boot-time `scan_from` (the old `created_at_block.or(scan_from)`). Fine for a rotation onto an Outbox created after boot; a rotation onto a pre-existing Outbox with earlier messages would skip them. Covered by the per-active-Outbox listener work (`activeOutboxes` + `OutboxRegistered` block) that is planned anyway, so just noting it.

didac-gluwa:
@DylanVerstraete addressed all three points from your 8 Sep review:

1. **E2e blocking fix**: `deploy-source-ethers.mts` now deploys `ChainRegistry` + `OutboxDiscovery` (behind its UUPS proxy, via the same `OutboxDiscoveryProxy` ERC-1967 wrapper the contract's own Hardhat suite uses), registers the Outbox as the chain key's default, and calls `set_outbox_discovery_addr` via sudo — mirroring the existing `set_outbox_factory_addr` registration. Verified live end-to-end (rebuilt node + real anvil, `asc-contracts@434073943` compiled locally, not just read through): after this wiring, `chain-info.get_outbox_discovery_address(2)` returns the deployed registry with `exists=true`, and its `defaultOutbox(2)` returns the deployed Outbox — the exact path `resolver.rs` reads.
2. **PR body**: added the "⚠️ Do not deploy yet" section spelling out the three devnet prerequisites, per your comment.
3. **Enum-index nit**: `OutboxDiscoveryRegistered`/`ZeroOutboxDiscoveryAddress` moved to the end of their enums instead of inserted mid-list. Regenerated `metadata.scale` and the CLI's `polkadot.js` interfaces from a freshly rebuilt node against the corrected order (52 pallet tests + CLI `tsc`/`prettier` all green).

Ready for another look whenever you have time.

attestor/attestor/src/tasks/write_ability/resolver.rs
DylanVerstraete: Covered by the do-not-deploy section in the PR body and the three devnet prerequisites now on the cutover plan; resolving.

attestor/attestor/src/tasks/write_ability/resolver.rs
DylanVerstraete: Accepted as a follow-up: rotation onto a pre-existing Outbox is covered by the per-active-Outbox listener work (from-block from OutboxRegistered).

pallets/supported-chains/src/lib.rs
DylanVerstraete: Fixed in 648f45b2e, both variants appended.

pallets/supported-chains/src/lib.rs
DylanVerstraete: Fixed in 648f45b2e, both variants appended.

DylanVerstraete (approved):
Approving. The e2e now deploys ChainRegistry + OutboxDiscovery behind the proxy, registers the Outbox as default and calls `set_outbox_discovery_addr` via sudo, and run 34248823518 went green on it, so the trunk keeps a working e2e. Variants appended, metadata and CLI types regenerated, do-not-deploy note in the body. Remaining reds are branch-inherent (check-version, target-branch-check).

Merge when you like. Images from writeability-off-usc-dev are do-not-deploy on usc-devnet until the runtime upgrade, the OutboxDiscovery deployment with 0x95FE… as `defaultOutbox(8)`, and the sudo call exist; those are on the cutover plan and I will sequence them with the hash change from #1322.

PR #1293, opened 2026-08-27
brad-attest-coin-manual-test

BradleyOlson64:
1. Adds end to end manual test with scripts and readme which covers the following path
2. Modifies the precompile cli to handle 0x private key input in addition to nmemonic


BradleyOlson64:
The `High risk` flag from bugbot is actually referring to the changes in the attestcoin feature branch against usc-dev

This branch is super low impact. Only manual testing scripts + readme and a single core change: Allowing the cli to accept raw secret seeds as well as nmemonics

Merging now to get the attestcoin feature in better condition for review

PR #1313, opened 2026-09-02
feat(proof-gen): derive block_confirmation_depth from on-chain MaturityStrategy

DylanVerstraete:
## Problem

`block_confirmation_depth` in proof-gen was a hand-typed `u64` with `#[serde(default)]`, so **omitting it silently meant `0` — no reorg protection.** The attestors read the same quantity from the chain's `MaturityStrategy` in the supported-chains pallet (`EvmFinalized ⇒ 64`, `EvmSafe ⇒ 32`, `EvmLatest ⇒ 0`, `FixedDelay:n`). Two different sources for one safety parameter is how usc-dev ended up with attestors at 64 and the prover at 32 — and it could just as easily have been 0.

## Change

`block_confirmation_depth` becomes `Option<u64>` in the YAML, the CLI (`--block-confirmation-depth` / `BLOCK_CONFIRMATION_DEPTH`), and `ChainConfig`.

- **`None` (omitted) → derived at startup** from `supported_chain.maturity_strategy`, which `build_continuity_for_chain` already fetches — no additional RPC call. This is the recommended setting; the prover then agrees with the attestors by construction.
- **`Some(n)` → explicit pin.** Startup logs INFO if it matches the chain, **WARN if it disagrees**, naming both values and stating the consequence (prover confirms blocks on a different schedule from the attestors).
- **Fails startup** if the on-chain strategy is unparseable or has no block-depth equivalent. A prover guessing its reorg window is worse than one that is down. Approved as intended behaviour.

Resolved depth is logged at INFO on every boot so it is visible without reading config.

## Effect on existing deployments

Every current values file sets the field explicitly, so **no deployment changes behaviour on upgrade** — they log INFO (match) or WARN (mismatch) and keep their pinned value. To adopt derivation, delete the line. The only semantic change is for configs that *omitted* the field: they previously ran with 0 and will now run with the chain's value, which is the fix.

## Not in scope

The archiver already derives its lag from chain (`FINALIZATION_LAG` is an override; unset → on-chain lookup) — but only on this branch; no tagged release up to 3.133 has it. Until an archiver release ships it, its `finalizationLag` must stay explicit in IaC.

## Testing

`cargo check -p proof-gen-api-server --tests --bins` clean, no warnings. `cargo test -p proof-gen-api-server`: **78 passed**, including two new config tests — omitted field parses to `None` (regression guard against the old `0` default), explicit field parses to `Some(64)`. `cargo fmt` applied.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10928,6 +10928,7 @@ dependencies = [
  "serde_yaml",
  "sp-core",
  "stream",
+ "supported-chains-primitives",
  "sysinfo 0.37.2",
  "thiserror 1.0.69",
  "tokio",
--- a/proof-gen-api-server/Cargo.toml
+++ b/proof-gen-api-server/Cargo.toml
@@ -58,6 +58,7 @@ tracing = { workspace = true }
 # Creditcoin / USC dependencies
 attestor-primitives = { workspace = true }
 cc-client = { workspace = true }
+supported-chains-primitives = { workspace = true }
 continuity = { workspace = true }
 eth = { workspace = true }
 hex = { workspace = true }
--- a/proof-gen-api-server/Cargo.toml
+++ b/proof-gen-api-server/Cargo.toml
@@ -58,13 +58,13 @@ tracing = { workspace = true }
 # Creditcoin / USC dependencies
 attestor-primitives = { workspace = true }
 cc-client = { workspace = true }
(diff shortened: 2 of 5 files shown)


BradleyOlson64 (approved):

beqaabu (approved):
neat

PR #1316, opened 2026-09-03
ci(abi-gate): cover OutboxDiscovery and bump the asc-contracts pin

DylanVerstraete:
asc-contracts #38 merged this morning (`a4d05d7`) with OutboxDiscovery plus the automatic timelock. Our ABI drift gate was pinned at `e300c166`, main @ **2026-08-17** — before the contract existed. So nothing was checking the surface the registry-first resolver is about to depend on.

## What this adds

`IOutboxDiscovery` mirrored in `common/write-ability` (only the surface we read), asserted against the compiled artifact:

**Events (5)** — `OutboxRegistered`, `OutboxRemovalScheduled`, `DefaultOutboxChangeScheduled`, `PendingDefaultCancelled`, `PendingRemovalCancelled`
**Reads (6)** — `defaultOutbox(uint32)`, `activeOutboxes(uint32)`, `isActiveOutbox`, `defaultDeployer()`, `pendingDefaultOutbox`, `pendingRemovalTime`

Both halves matter for different failure modes. A moved topic0 silently blinds a subscription, and the timelock events are our only advance notice before a default changes or an Outbox is removed. A moved selector makes a read revert with empty returndata, which the resolver **cannot distinguish from "no Outbox registered for this chain key"** — it would resolve to nothing and idle, which is the quiet failure this gate exists to prevent.

## Two things captured in doc comments

- **`effectiveTime` is a unix timestamp, not a block number.** Didac asked for a block; the contract ships seconds. Our listeners are block-indexed, so there is a conversion to write. Worth being explicit before someone assumes it is a height.
- **`activeOutboxes()` filters an Outbox out the moment its scheduled removal time passes**, before `cleanup` runs. So membership disappearing is *not* the signal to stop watching one — we drain on our own in-flight state. Confirmed with Kevin on the #38 thread.

## Deliberately not asserted

`activeOutboxCount` — it existed in the revision I reviewed and was **dropped before merge**. Asserting it would have failed against a function that no longer exists.

## Verification

Ran the gate against a fresh `npx hardhat compile` of merged main. It **failed first, exactly as designed**, naming the stale pre-timelock `DefaultOutboxSet` / `OutboxRemoved` events still in my local artifacts — then passed once recompiled. That is the gate proving it can actually catch drift rather than passing vacuously.

`cargo test -p write-ability` and `-p attestor --lib` (140) pass, clippy `-D warnings` clean, fmt clean, workflow YAML parses, and the pinned SHA is verified to resolve to the #38 merge commit on `main`.

## Update 7 Sep

asc-contracts #46 moved the OutboxDiscovery timelock from unix time to block numbers. Mirror updated: `effectiveTime` → `effectiveBlock` on `OutboxRemovalScheduled` / `DefaultOutboxChangeScheduled` and in `pendingDefaultOutbox`; `pendingRemovalTime` → `pendingRemovalBlock`; `timelock()` and `MIN_TIMELOCK` (1200) are in blocks. Event topics are unchanged by a parameter rename, so only the function selector assertion moves; the listener is block-indexed already, so drain-before-drop needs no conversion. e2e checkout re-pinned to main @ `434073943`. Local: fmt, clippy `-D warnings` on `write-ability` + `attestor`, `write-ability` tests green; the e2e's ABI gate runs in CI against the new pin.

https://claude.ai/code/session_01Kv6y52MHgtgkWmnJhdXbHB

### DIFF
--- a/.github/workflows/write-ability-e2e.yml
+++ b/.github/workflows/write-ability-e2e.yml
@@ -118,7 +118,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: gluwa/asc-contracts
-          ref: e300c16676efc850a5b382033d666bc9c3a0ce23 # main @ 2026-08-17 (post-#23/#28 v3 RelayerQuote struct) — bump deliberately
+          ref: a4d05d7b6f49b01592233fb34ce7a9212c178ee9 # main @ 2026-09-03 (post-#38 OutboxDiscovery + automatic timelock) — bump deliberately
           path: asc-contracts
           # asc-contracts is private; the default GITHUB_TOKEN is scoped to creditcoin3 only and
           # gets "Repository not found". Use the org bot PAT (same one build-native uses cross-repo).
--- a/.github/workflows/write-ability-e2e.yml
+++ b/.github/workflows/write-ability-e2e.yml
@@ -118,7 +118,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: gluwa/asc-contracts
-          ref: a4d05d7b6f49b01592233fb34ce7a9212c178ee9 # main @ 2026-09-03 (post-#38 OutboxDiscovery + automatic timelock) — bump deliberately
+          ref: 434073943ec6828c3271bcf39114a04304e97d7f # main @ 2026-09-04 (#46 block-based timelock) — bump deliberately
           path: asc-contracts
           # asc-contracts is private; the default GITHUB_TOKEN is scoped to creditcoin3 only and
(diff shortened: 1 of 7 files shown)


DylanVerstraete:
e2e was red on `Deploy fee stack` after the pin bump: the scripts still opened `USCProofVerifier` / `USCRelayingQuoter` artifacts, renamed to `ASC*` in asc-contracts on 24 Aug. Fixed the artifact paths in the deploy scripts; constructors and all setters were verified unchanged against artifacts compiled at the pinned SHA. e2e re-running.

PR #1320, opened 2026-09-04
feat(attestation): derive quorum from min(active attestors, target sample size)

DylanVerstraete:

`TargetSampleSize` becomes a **cap** on the voting committee rather than the committee itself. The quorum `commit_attestation` enforces is now `2/3+1` of `min(|ActiveAttestors|, TargetSampleSize)`.

```
sample_size = activeAttestors < target_sample_size ? activeAttestors : target_sample_size
quorum      = 2/3 + 1 of sample_size
```

### Why

Deriving the threshold from the target alone made a target **above** the active attestor count unsatisfiable: every submission failed `MajorityNotReached` and attestation for the chain stopped permanently, recoverable only by governance lowering the target.

This was reachable in shipped config. The local testnet spec ships `target_sample_size: 9`, which required 7 active attestors.

### What changed

The formula lives in one place, `attestor_primitives::calculate_quorum`, called by both the runtime and the attestor so the two cannot drift.

**Runtime**
- `validate_attestation` feeds in the live active-set size it already gathers for the eligibility filter.
- New `Pallet::quorum_threshold` exposes the authoritative number.
- `calculate_threshold` now saturates its `* 2`. The setter only rejects zero and the release profile has no `overflow-checks`, so a target above `u32::MAX / 2` previously wrapped and collapsed the threshold.

**Attestor**
- New `Client::quorum` reads the active set and the target from a **single storage snapshot**, so the two inputs can never come from different blocks.
- Every site that derived a threshold from the target alone now uses it: boot, the `TargetSampleSizeChanged` handler, and both pre-submit threshold gates. Without this, raising the target above the active count would leave nodes enforcing an unreachable local quorum and silently never submitting, while the runtime happily accepted attestations nobody produced.
- `note_target_sample_size_change` → `note_quorum_change`, taking a threshold rather than a target so a caller cannot pass the wrong thing.
- Quorum is refreshed on **membership** changes too, not just target changes. Election, chill and kick previously refreshed only the BLS store and the pool allow-set. The active set moves immediately on chill/kick, not at the epoch boundary, and it is half the denominator.

### Tests

Six pallet tests asserted the old behavior. Three were asserting the bug itself (one active attestor, a high target, expect rejection). They now express an unmet quorum the correct way, by making the active set larger than the signer set, via a new `pad_active_attestors` helper.

Added:
- regression test that a target above the active count no longer blocks attestation
- primitive tests for the cap, the reachability invariant (`quorum <= active` for every target), the quorum-intersection tradeoff, and the saturation guard

| Suite | Result |
| --- | --- |
| `pallet-attestation` | 264 pass |
| `attestor` | 52 pass |
| `attestor_pool` | 16 pass |
| `attestor-primitives` | 14 pass |
| `cc-client` | 6 pass |

`cargo clippy -D warnings` and `cargo fmt` clean. No storage changes, so no migration and no `spec_version` bump.

### Operator note

Quorum intersection (two conflicting quorums cannot both exist at one height) holds **only while the cap does not bind**, because nothing selects a committee when it does. Sortition is unbuilt: see the unused `_randomness` in `do_start_election`, RFC-0174. That is the USCP2-004 finding, narrowed rather than closed.

So keep `TargetSampleSize` above the active attestor count. Two follow-ups deliberately **not** in this PR:

1. `DefaultTargetSampleSize` is still `3` (`runtime/src/lib.rs`). Chains registered without an explicit target start in the binding regime as soon as more than 3 attestors are active. Raising it is a policy call.
2. Already-registered chains keep their stored targets until `set_target_sample_size` is run per chain. It applies at the next epoch.

---

When merging this PR:


beqaabu (approved):

didac-gluwa (approved):
I still believe we should add a floor to the sample_size but the PR itself is good

PR #1329, opened 2026-09-08
ci: report peak runner resource usage in job summaries

mason-gluwa:

 Adds `.github/resource-monitor.sh` — `start` forks a 10s sampler, `report` reduces it to peak
memory / disk / load and writes a table plus an over-provisioned verdict to the job summary. Wired
into `build-native.yml` (covers every build job, hosted and self-hosted) and the three
`g6-standard-16` workflows.

Memory is `MemTotal - MemAvailable` rather than `used`, so reclaimable page cache isn't charged to
the workload.

---

When merging this PR:

---

Practical tips for PR review:

### DIFF
--- a/.github/resource-monitor.sh
+++ b/.github/resource-monitor.sh
@@ -12,46 +12,57 @@
 #
 # Never fails a job: every path exits 0.
 
-LABEL="${2:-job}"
+LABEL="${2:-${GITHUB_JOB:-job}}"
 STATE_DIR="${RESOURCE_MONITOR_DIR:-/var/tmp/resource-monitor}"
 SAMPLES="${STATE_DIR}/${LABEL}.tsv"
 PIDFILE="${STATE_DIR}/${LABEL}.pid"
+BASEFILE="${STATE_DIR}/${LABEL}.baseline"
 INTERVAL="${RESOURCE_MONITOR_INTERVAL:-10}"
-# Watch the filesystem holding the workspace; that is what fills up.
+# Defaults to the workspace, which is right for build jobs (cargo target dir).
+# Jobs that write their bulk data elsewhere MUST set RESOURCE_MONITOR_DISK_PATH
+# -- runtime-upgrade uses /mnt, the compatibility workflows /var/tmp. Watching
+# the wrong path silently reports ~0% once that data lands on its own volume.
 DISK_PATH="${RESOURCE_MONITOR_DISK_PATH:-${GITHUB_WORKSPACE:-$PWD}}"
 
-mem_total_mb() { awk '/^MemTotal:/ {print int($2/1024)}' /proc/meminfo; }
-
-sample_once() {
-    local total avail used swap_total swap_free swap_used disk load
-    total=$(awk '/^MemTotal:/ {print int($2/1024)}' /proc/meminfo)
-    avail=$(awk '/^MemAvailable:/ {print int($2/1024)}' /proc/meminfo)
-    # MemTotal-MemAvailable tracks memory the workload actually needs, unlike
-    # "used", which counts reclaimable page cache against you.
-    used=$(( total - avail ))
-    swap_total=$(awk '/^SwapTotal:/ {print int($2/1024)}' /proc/meminfo)
-    swap_free=$(awk '/^SwapFree:/ {print int($2/1024)}' /proc/meminfo)
-    swap_used=$(( swap_total - swap_free ))
-    disk=$(df -m --output=used "$DISK_PATH" 2>/dev/null | tail -1 | tr -d ' ')
-    load=$(awk '{print $1}' /proc/loadavg)
-    printf '%s\t%s\t%s\t%s\t%s\n' \
-        "$(date -u +%s)" "$used" "$swap_used" "${disk:-0}" "$load"
+# One awk pass over /proc/meminfo -> "used_mb total_mb swap_used_mb".
+# used = MemTotal-MemAvailable, which is what the workload actually needs;
+# "used" as reported by free() counts reclaimable page cache against you.
+mem_stats() {
+    awk '/^MemTotal:/ {t=$2} /^MemAvailable:/ {a=$2}
+         /^SwapTotal:/ {st=$2} /^SwapFree:/ {sf=$2}
+         END { printf "%d %d %d", int((t-a)/1024), int(t/1024), int((st-sf)/1024) }' \
+        /proc/meminfo
 }
 
-# Order matters here: these conditions are not mutually exclusive. A job can be
+# Cumulative busy/total jiffies -> "busy total". Deltas between two readings
+# give real CPU utilisation. loadavg is a 60s EMA and cannot see a burst
+# shorter than ~45s, which is the shape of most jobs here, so it is not used
+# for the verdict. iowait counts as idle on purpose: a job waiting on disk is
+# disk-bound, not CPU-bound, and should be reported that way.
+cpu_jiffies() {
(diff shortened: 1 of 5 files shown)


didac-gluwa (dismissed):

.github/workflows/build-native.yml
```diff
@@ -91,12 +91,22 @@ jobs:
           git apply /var/tmp/new-metadata.diff
           git diff --binary
 
+      - name: Start resource monitor
```
DylanVerstraete, line 94: **Blocking.** In the `use-new-metadata` path the only checkout is the PR *head branch* ("Checkout code for typedefs update"), which for every currently open PR does not contain `.github/resource-monitor.sh`, while the workflow definition comes from the merge commit and does contain this step. Once this merges, any open PR that changes runtime metadata fails `rebuild-new-metadata` with `bash: .github/resource-monitor.sh: No such file or directory` until its author rebases, and the metadata bot never lands its commit.

The Report step already has `if: always()` + `continue-on-error: true`; this step needs the same protection, e.g. `continue-on-error: true` or `if: hashFiles('.github/resource-monitor.sh') != ''`.

.github/workflows/build-native.yml
```diff
@@ -91,12 +91,22 @@ jobs:
...
+        run: |
+          .github/resource-monitor.sh start build-native
+
       - name: Build SUT
         uses: gluwa/cargo@dev
         with:
           command: build
           args: ${{ inputs.build-options }}
 
+      - name: Report resource usage
```
DylanVerstraete: On the release path (`release.yml` calls with `build-subkey`/`build-subxt`), `Install subkey` and `Install subxt` are full `cargo install --git` compiles on the `type-native` VM and run *after* this step has killed the sampler, so they never enter the peaks. A downsizing decision based on this summary would miss them. Suggest moving Report after `Upload subxt binary` (keep `if: always()`).

.github/resource-monitor.sh
```diff
@@ -0,0 +1,142 @@
...
+    else
+        echo "sized about right"
+    fi
+}
+
+case "${1:-}" in
+start)
+    mkdir -p "$STATE_DIR" || exit 0
+    printf 'epoch\tmem_used_mb\tswap_used_mb\tdisk_used_mb\tload1\n' > "$SAMPLES"
+    (
```
DylanVerstraete, line 79: The subshell (and its `sleep` child) inherits the step's stdout/stderr pipe; only `sample_once`'s stdout is redirected. actions/runner then waits its 5 s "STDIO streams did not close within 5 seconds of the exit event" grace period. Confirmed on this PR's runs: every `Start resource monitor` step took exactly 5 s. Fix:

```bash
(
    while true; do ...; done
) </dev/null >/dev/null 2>&1 &
```

.github/resource-monitor.sh
```diff
@@ -0,0 +1,142 @@
...
+#
+# Never fails a job: every path exits 0.
+
+LABEL="${2:-job}"
+STATE_DIR="${RESOURCE_MONITOR_DIR:-/var/tmp/resource-monitor}"
+SAMPLES="${STATE_DIR}/${LABEL}.tsv"
+PIDFILE="${STATE_DIR}/${LABEL}.pid"
+INTERVAL="${RESOURCE_MONITOR_INTERVAL:-10}"
+# Watch the filesystem holding the workspace; that is what fills up.
+DISK_PATH="${RESOURCE_MONITOR_DISK_PATH:-${GITHUB_WORKSPACE:-$PWD}}"
```
DylanVerstraete, line 25: None of the Start steps set `RESOURCE_MONITOR_DISK_PATH`, and the heavy jobs write outside the workspace: live-sync/fork use `/mnt` (the 114→158 GB expansion this script's own comment cites), the compat jobs `/var/tmp/*-compatibility-data`. It measures the right thing today only because the Linode VMs have a single root disk, and the summary row is labelled with the wrong path. The moment a volume is attached at `/mnt` (the natural fix for a 158 GB dataset) the Disk row reads ~0% and the verdict flips to over-provisioned for exactly the job the comment says must not be shrunk. Suggest setting `RESOURCE_MONITOR_DISK_PATH: /mnt` (resp. `/var/tmp`) in those Start steps and adjusting the comment.

DylanVerstraete (changes requested):
Sound, low-risk addition overall: `report` is guarded, the script exits 0 everywhere, and this PR's own runs show it working end to end (every wired job produced a report; a build peaks at ~5.5–6.5 GB RAM, load ≈1.4× vCPU, hosted disk 44%). One item blocks merge (inline, `build-native.yml` Start step), one is a cheap fix (5 s per Start step), two are placement/path issues. The rest are suggestions.

**Suggestions (non-blocking)**

- **Load average is a weak "CPU never saturated" signal** (`resource-monitor.sh:53`). `load1` is a 60 s EMA: a full-core burst shorter than ~45 s cannot reach 50% of vCPU, and the compat/fork jobs are exactly that bursty shape (chainspec gen, node start, npm builds, JS tests). The over-provisioned verdict is therefore biased toward shrinking. Sampling busy jiffies from `/proc/stat` per interval is direct; at minimum drop "never" and note the blind spot.
- **No disk baseline** (`resource-monitor.sh:35`). Peak disk is absolute `df used`, so the runner image and earlier jobs' leftovers are charged to the job. Record `df --output=used` at `start` and report the delta; skip the verdict when `RUNNER_ENVIRONMENT=github-hosted` since there is no plan to resize there.
- **Verdict gap at exactly 50% disk** (`resource-monitor.sh:53`): branch 2 is `-gt 50`, branch 3 is `-lt 50`, so 49 → over-provisioned, 51 → disk-bound, 50 → "sized about right". Use `-le` in one place.
- **The `[ ! -s "$SAMPLES" ]` guard cannot fire after a normal start** (`resource-monitor.sh:79`): the header is always written. A label typo between the hand-copied start/report pairs prints a misleading "no samples" warning, and a dead sampler yields a one-sample verdict. Test `-f`, default the label to `GITHUB_JOB`, warn when `SAMPLE_COUNT` is tiny.
- **Path filters**: runtime-upgrade, validator-compatibility and attestor-compatibility don't list `.github/resource-monitor.sh` in `on.pull_request.paths`, unlike `benchmarks.yml`/`promote-branch.yml` for their helper scripts.
- **Cleanup**: `mem_total_mb` duplicates the awk inside `sample_once`; five awk forks per sample could be one pass; percent is computed two ways (`pct()` and the inline awk for load).

.github/resource-monitor.sh
beqaabu: Seconding this

.github/resource-monitor.sh
```diff
@@ -0,0 +1,192 @@
...
+#   resource-monitor.sh report [label]   # stop sampling, print + summarise peaks
+#
+# Never fails a job: every path exits 0.
+
+LABEL="${2:-${GITHUB_JOB:-job}}"
+STATE_DIR="${RESOURCE_MONITOR_DIR:-/var/tmp/resource-monitor}"
+SAMPLES="${STATE_DIR}/${LABEL}.tsv"
+PIDFILE="${STATE_DIR}/${LABEL}.pid"
+BASEFILE="${STATE_DIR}/${LABEL}.baseline"
+INTERVAL="${RESOURCE_MONITOR_INTERVAL:-10}"
```
beqaabu, line 20: the 10 second sampling interval might miss the peak resource usages on jobs that have a few second spikes, might need to decrease the interval or read data from ` /sys/fs/cgroup/memory.peak`

beqaabu (changes requested):

PR #1330, opened 2026-09-09
feat(maturity): RpcSafe / RpcFinalized strategies that follow the source node's block tags

DylanVerstraete:
> **Stack** (bottom → top; each PR's diff is only its own commits, review in this order)
>
> | | PR | Phase | What | Base |
> |---|---|---|---|---|
> | **▶** | #1330 | Phase 0 | RpcSafe / RpcFinalized maturity strategies | `usc-dev` |
> | &nbsp;&nbsp; | #1362 | Phase 1a | receipts must carry the fetched block's hash | #1330 |
> | &nbsp;&nbsp; | #1365 | Phase 2b | prover confirms against the attestation cache | #1362 |
> | &nbsp;&nbsp; | #1366 | Phase 2a | archiver follows the latest attested height | #1365 |
> | &nbsp;&nbsp; | #1367 | Phase 1c | block tags resolved by cross-provider agreement | #1366 |
>
> Plan: https://claude.ai/code/artifact/400f8668-7b43-4b11-86c8-b8922d8c85ee · Live run against Sepolia `safe` via Chainstack (2026-09-15): attested heights tracked `safe` within one interval, archiver ≤ attested ≤ safe at every sample, prover 200 at attested / 422 one above; survived a 20-min RPC outage without restarts.

## Why

Every maturity strategy today is a fixed block offset. `EvmSafe` / `EvmFinalized` are the Ethereum-epoch approximations (32 / 64 blocks), which on a 2 s rollup such as Base mean 64 s / 128 s and say nothing about whether the block was posted to L1. A fixed count also cannot follow a stalled batcher, and the OP docs note `safe` can trail for hours in that case.

## What

Two new on-chain strategy strings, **`RpcSafe`** and **`RpcFinalized`**, resolved off-chain to the source node's `safe` / `finalized` block tags. Chain-agnostic on purpose: Ethereum, OP-Stack and Arbitrum nodes all serve these tags with their own correct meaning. `EvmSafe` / `EvmFinalized` / `FixedDelay` keep their exact behaviour, so nothing changes for registered chains.

- **primitives / pallet**: `MaturityStrategy::{RpcSafe, RpcFinalized}`, `RpcBlockTag`, `rpc_tag()`; `is_valid_maturity_strategy` accepts the strings. This changes the runtime wasm; no spec bump in this PR (release-owned), so `check-version` is expected red.
- **eth**: `Maturity { FixedLag(u64), Tag(BlockTag) }` with `mature_height`, and `Client::get_block_number_by_tag` using the same fallback-provider walk as block fetches.
- **streams**: `tip` and `roots` take `maturity` instead of `finalization_lag` and resolve it per head. The roots stream now fetches the whole `next_unfetched..=mature` range, so a `safe` head that jumps by a full batch is handled; a failed tag lookup is logged and retried on the next head, never guessed.
- **attestor / archiver**: resolve the on-chain strategy with `stream_eth::maturity_from_strategy`.
- **continuity / proof-gen**: `ContinuityConfig::confirmation_tag` and `EthRpcProvider::get_block_number_by_tag` make the prover confirm blocks on exactly the boundary the attestors attest on. `BlockNotOnSourceChain` now reports the effective window (`tip - confirmed`) instead of the configured depth, which is `0` under a tag policy.

## Verification

- New tests: `common/eth/tests/block_tag.rs` (tag → number, `null` → not found, clamp to observed head, fallback provider), primitives parse/accessor tests, pallet accept/reject cases, resolver test in `stream_eth`. All suites of the touched crates pass; clippy `-D warnings`, fmt, taplo and cargo-machete clean; `cargo check -p creditcoin3-runtime` passes.
- **Live on Base Sepolia (Chainstack), 3 attestors, local dev node**, chain flipped to `RpcSafe`:

  | time (UTC) | attested | node `safe` | attested − safe |
  |---|---|---|---|
  | 07:06:30 | 46584480 | 46584483 | −3 |
  | 07:06:50 | 46584620 | 46584627 | −7 |
  | 07:11:15 | 46584760 | 46584772 | −12 |
  | 07:11:20 | 46584770 | 46584772 | −2 |

  Attested heights never exceeded `safe`; they advance in bursts as batches post to L1 (~4–5 min on Base Sepolia). The proof server served a block 72 behind `safe` (61 continuity roots) and returned `BlockNotOnSourceChain` for blocks above it.

## Notes for operators

- Registering a chain with `RpcSafe` / `RpcFinalized` needs the runtime upgrade first (the pallet validator rejects unknown strings). For the live test above the storage value was written directly on a dev node.
- Tag-based maturity costs one `eth_getBlockByNumber(tag, false)` per source head per component.
- Independent of #1319 (Base / OP-Stack); the live test ran with this patch applied on top of that branch.

**Since approval:** rebased onto `usc-dev` (picked up #1320, #1326, #1335, #1343); no code changes.


common/continuity/src/builder/mod.rs
```diff
@@ -559,12 +559,21 @@ impl ContinuityBuilder {
...
         let tip = self.eth_provider.get_last_block().await?;
-        let confirmed = tip.saturating_sub(self.config.block_confirmation_depth);
+        let confirmed = match self.config.confirmation_tag {
+            Some(tag) => self
+                .eth_provider
+                .get_block_number_by_tag(tag)
+                .await?
+                .min(tip),
+            None => tip.saturating_sub(self.config.block_confirmation_depth),
+        };
```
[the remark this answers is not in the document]
DylanVerstraete: Good catch. Fixed in 2e50f485f: `ArchiverEthProvider::get_block_number_by_tag` now forwards to the live ETH fallback (the archiver only stores roots, tags are a live-node property). `MockEthRpcProvider` answers fixed `safe`/`finalized` heights so `confirmed_last_block_follows_the_block_tag` and `archiver_provider_forwards_block_tag_lookups` cover both paths.

proof-gen-api-server/src/lib.rs
```diff
@@ -143,57 +143,78 @@ impl Server {
...
                 tracing::info!(
                     chain_key,
                     block_confirmation_depth = explicit,
                     "⛓️  reorg-protection depth pinned in config; matches on-chain MaturityStrategy"
                 );
-                explicit
+                (explicit, None)
             }
-            Some(explicit) => {
+            (Some(explicit), _) => {
```
beqaabu: Explicit depth silently wins over a tag strategy - The (Some(explicit), _) arm drops on_chain_tag. Previously maturity_delay().ok_or_else(...)? ran unconditionally, so this combination failed startup. Now an operator with block_confirmation_depth: 32 pinned in YAML against an RpcSafe chain starts cleanly and confirms at tip - 32 (~64 s on Base) while attestors attest at safe (multi-minute batch cadence) — the prover serves blocks nobody attested. Only a warn! marks it.

common/continuity/src/rpc.rs
```diff
@@ -493,6 +503,16 @@ impl EthRpcProvider for ReconnectingEthRpcProvider {
         .await
     }
 
+    async fn get_block_number_by_tag(&self, tag: eth::BlockTag) -> Result<u64> {
```
beqaabu, line 518: Permanent tag failure treated as transport failure - get_block_number_by_tag goes through run(), which retries ETH_RPC_MAX_ATTEMPTS times and tears down + re-dials the shared ETH client with backoff on any Err. A node answering null for safe yields FailedToGetBlockByTag, which is deterministic and permanent, but has no carve-out (unlike user::Shutdown and inconsistent-payload). Every /continuity request then pays N reconnects, churning the client for other in-flight requests too.

proof-gen-api-server/src/lib.rs
DylanVerstraete: Agreed, that arm quietly re-enabled the exact mismatch the old `ok_or_else(...)?` refused. Fixed in 2d360ff27: `(Some(explicit), Some(tag))` is now a startup error telling the operator to remove `block_confirmation_depth`; depth-vs-depth disagreements keep the WARN. Config docs updated in both places.

common/continuity/src/rpc.rs
DylanVerstraete: Right, and it was worse than described: the closure stringified the error (`anyhow!("...{e}")`), so nothing downstream could have recognised it anyway. Fixed in 2d360ff27: the typed `FailedToGetBlockByTag` cause is preserved via `.map_err(anyhow::Error::from).with_context(..)`, and `run()` has a carve-out (`eth::anyhow_chain_is_unsupported_block_tag`, next to the shutdown and inconsistent-payload ones) that returns immediately without reconnecting. Classifier has unit tests incl. the stringified case staying non-permanent.

beqaabu (changes requested):
Resolve conflicts pls

DylanVerstraete:
Rebased onto usc-dev; see above.

didac-gluwa (approved):

beqaabu (approved):

BradleyOlson64 (approved):

proof-gen-api-server/src/lib.rs
```diff
@@ -143,57 +143,90 @@ impl Server {
...
+        });
+        let on_chain_depth: Option<u64> = strategy.maturity_delay();
+        if on_chain_tag.is_none() && on_chain_depth.is_none() {
+            return Err(anyhow!(
+                "chain_key {chain_key}: maturity strategy {strategy:?} has neither a block-depth \
+                 nor a block-tag equivalent; set block_confirmation_depth explicitly"
+            ));
+        }
+        let (block_confirmation_depth, confirmation_tag) = match (
+            chain.block_confirmation_depth,
```
BradleyOlson64, line 176: I think we can get rid of the config field chain.block_confirmation_depth

We already return early when there isn't an on-chain depth or tag.

And I don't think there's any case where we want the config depth to override a present on-chain depth or tag.

So there isn't a good use for it imo. It just gives us more room to misconfigure things so that the prover delay doesn't match the attestor delay.

proof-gen-api-server/src/config.rs
```diff
@@ -194,7 +196,8 @@ pub struct ChainConfigFile {
     pub archiver_url: Option<String>,
     /// Reorg-protection depth override. **Omit it** to derive the depth from the chain's on-chain
     /// `MaturityStrategy` (recommended -- matches the attestors by construction). Set it only to
-    /// deliberately pin a value; startup warns if it disagrees with the chain.
+    /// deliberately pin a value; startup warns if it disagrees with the chain and fails if the
+    /// chain follows a block tag (`RpcSafe` / `RpcFinalized`).
     #[serde(default)]
     pub block_confirmation_depth: Option<u64>,
```
BradleyOlson64, line 202: As said elsewhere, I think we should remove this. Don't see a case where we want the prover's depth to disagree with the attestors.

proof-gen-api-server/src/lib.rs
DylanVerstraete: Agreed, and the stack above this PR goes further than removing the override: #1365 removes the prover's depth entirely. It no longer has a reorg window of its own to configure or reconcile. `validate_blocks` gates on the attestation cache (a proof needs an attestation at or above the height anyway), `ContinuityConfig::{block_confirmation_depth, confirmation_tag}` and `get_confirmed_last_block` are deleted, and this whole reconciliation block in `lib.rs` goes with them, along with the prover's dependency on `supported-chains-primitives`. So the disagreement you're describing becomes impossible by construction rather than by config discipline. I left this PR's version as-is so #1365's diff shows the removal cleanly; happy to fold it down here instead if you'd rather not see the intermediate state.

proof-gen-api-server/src/config.rs
DylanVerstraete: Same answer as above: #1365 deletes the effective field. The one thing it keeps is the YAML key (and the CLI flag) as a deprecated no-op that logs a warning, because the `creditcoin-proof-gen-api` chart in cc-networks-iac still renders `block_confirmation_depth` and a hard removal would fail every pod on the image bump before the chart is updated. If you'd prefer it gone outright and to sequence the chart change first, say so and I'll drop the key in #1365.

primitives/supported-chains/src/lib.rs
```diff
@@ -24,20 +24,55 @@ pub struct SupportedChain {
...
+///   [`EvmLatest`](Self::EvmLatest), [`FixedDelay`](Self::FixedDelay)): the mature height is
+///   `tip - delay`. `EvmSafe` / `EvmFinalized` are the Ethereum-epoch approximations (32 / 64
+///   blocks); they are *not* RPC tags and mean very different things on a 2 s rollup.
+/// - **RPC tags** ([`RpcSafe`](Self::RpcSafe), [`RpcFinalized`](Self::RpcFinalized)): the mature
+///   height is whatever the source node reports for `eth_getBlockByNumber("safe" | "finalized")`.
+///   On Ethereum that tracks justification / finality; on OP-Stack and Arbitrum rollups it tracks
+///   L1 batch inclusion / L1 finality, which a fixed block count cannot follow.
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum MaturityStrategy {
     EvmFinalized,
```
BradleyOlson64, line 42: Seems like we're handling this case through `RpcFinalized`

Can we remove the cases `EvmSafe` and `EvmFinalized`?

Would need to make sure we change any exisiting maturity strategies to the new variant before removing.

So removing these would have to be a follow up pr. Still helpful imo, since having multiple `safe` and `finalized` options is an invitation to use the wrong one.

DylanVerstraete: Agreed, and it is planned as Phase 3 of the same effort, in this order: (1) add `set_maturity_strategy(chain_key, strategy)`, since today the pallet only has `register_chain` / `remove_chain` and re-registering takes a fresh key and orphans the chain's attestation history; (2) change `DefaultMaturityStrategy` away from `EvmSafe` and reject the three `Evm*` names on new registrations while still parsing them for the chains already holding them; (3) move the live chains one at a time (the registry currently has 9 `EvmSafe`, 2 `EvmFinalized`, 5 `FixedDelay`); (4) delete the variants. Steps 1 and 2 are a pallet change, so they ride a runtime upgrade and an attestor roll. Kept out of this stack on purpose so the maturity work can land without a runtime change.

PR #1350, opened 2026-09-11
fix(proof-gen): share one Arc<CcClient> and supervise the cc3 event task

DylanVerstraete:
First PR of the prover liveness audit (2026-09-11), findings **7** and **1**.

## Problems

- **Finding 7.** `CcClient::clone` creates an independent connection slot (a fresh `ArcSwap`). `Server` value-cloned the client into each `ContinuityBuilder` and again into the event task. After a socket close the event task reconnected *its* copy; the builders stayed on the dead socket permanently (reproduced with a loopback WS node: the reconnected client succeeds, the builder-side client keeps failing).
- **Finding 1.** `StreamCC3` ends itself only when history is permanently gone (pruned state, or a reconnect gap past `MAX_BACKFILL_BLOCKS`) and its contract is that the consumer restarts. `Server::run` spawned the event task and dropped the handle; the HTTP server kept serving from caches that no longer tracked the chain, so newer proofs failed with `BlockNotReady` and missed checkpoint/reversion events were never repaired.

## Changes

- One `Arc<CcClient>` is created in `Server::new` and handed to every builder and to the event task. `start_cc3_event_subscription` takes `Arc<CcClient>`.
- `Server::run` supervises the event task through a new `supervise()` helper. When the task ends (error, clean return, or panic) the server marks the replica dead (`ContinuityService::mark_event_stream_dead`), asks axum to drain, waits up to 5 s for in-flight requests, and returns an error so the process **exits nonzero** and the orchestrator restarts it from a fresh finalized head. This is the same policy the attestor's production task applies to the same stream.
- `/api/v1/health` reports `degraded` with `cc3_event_stream_dead: <reason>` during the drain window. Readiness/503 semantics come in a later PR of this series (finding 5).

## Tests

- 4 unit tests for `supervise`: event-task error, event-task panic, shutdown signal, HTTP server exit.
- `tests/liveness_event_task.rs` with a disposable in-process WebSocket "Creditcoin node" (`tests/cc3_ws_fixture`): reconnect through the shared `Arc` repairs every holder with exactly one redial; pruned history makes the event task return `End of unbounded event stream`.
- `cargo test -p proof-gen-api-server` (81 unit, 27 route, 2 liveness), clippy `-D warnings`, fmt, taplo, machete.

## Series

A (this) → B finding 2 (stream progress watchdog) → C findings 3+4 (ETH repair lock, tip failover) → D findings 5+8 (readiness, consistent startup) → E finding 6 (admission budget, single-flight fills). Each stacks on the previous branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10932,6 +10932,7 @@ dependencies = [
  "sysinfo 0.37.2",
  "thiserror 1.0.69",
  "tokio",
+ "tokio-tungstenite 0.26.2",
  "tower 0.5.3",
  "tower-http 0.6.8",
  "tracing",
--- a/proof-gen-api-server/Cargo.toml
+++ b/proof-gen-api-server/Cargo.toml
@@ -73,8 +73,9 @@ default = []
 integration-tests = []
 
 [dev-dependencies]
-alloy-node-bindings = "0.11"       # kept at 0.11 due to compatibility with alloy
+alloy-node-bindings = "0.11"             # kept at 0.11 due to compatibility with alloy
 async-trait = { workspace = true }
 parameterized = "2.1.0"
 reqwest = { workspace = true }
+tokio-tungstenite = { workspace = true }
 wiremock = "0.6"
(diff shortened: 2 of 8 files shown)


DylanVerstraete:
Stack rebased onto usc-dev 57dc1a0af (after the maturity rework #1330/#1362/#1365/#1366/#1367 landed). Conflicts were confined to `common/eth/src/lib.rs` (kept both the rework's tag-lookup `call_timeout` and this stack's `fallback_urls`/HTTP deadlines) and `common/continuity/src/rpc.rs`. fmt, clippy -D warnings and the cc-client/continuity/stream_cc3/proof-gen/eth test suites pass on the stack top. Old heads are backed up locally as `backup/fix-prover-liveness-{a..e}-20260921`.

didac-gluwa (approved):

PR #1351, opened 2026-09-11
fix(stream_cc3): progress watchdog on the finalized subscription; ws ping/pong on the cc3 client

DylanVerstraete:
Second PR of the prover liveness audit, finding **2**. Stacked on #1350; retarget to `usc-dev` once that merges.

## Problem

A finalized-heads subscription can stay open and silent (dropped server-side, a proxy that stopped forwarding, a half-open peer). Both the seed and the steady-state loop in `StreamCC3` awaited `try_next()` with no deadline, so the stream and every consumer (prover caches, attestor production) hung forever with no error to react to. `cc_client::Client` used `RpcClient::from_insecure_url`, which leaves jsonrpsee WebSocket pings off, so a half-open socket never surfaced a transport error either. The workspace `reconnecting-rpc-client` feature was inert: it is a different client type that was never instantiated.

## Changes

- **Progress watchdog** in `StreamCC3` (`with_progress_timeout`, default 90 s). On expiry the stream reads the node's finalized head point-to-point:
  - node ahead of us → the subscription is dead: reconnect, re-subscribe, the existing parent walk fills the gap
  - node not ahead → finality is stalled upstream: keep waiting, log at warn
  - probe fails → reconnect
  The seed applies the same deadline so construction cannot hang on a subscription that never delivers.
- **`stream::cc3::Progress`** (height, last-advance time, silent-recovery count) shared through the config. The prover publishes it on `/health` as `cc3_finalized_height`, `cc3_finalized_age_seconds`, `cc3_silent_recoveries`. This is the per-block progress signal the flattened event stream had lost.
- **WebSocket ping/pong** on the cc3 client (30 s interval, 45 s inactive limit, 1 failure). Mirrors what subxt builds (same per-subscription buffer) plus pings. `Client::finalized_head_number()` added for the probe.
- **`cc_client::ws_fixture`** (cargo feature `ws-fixture`, tests only): an in-process fake Creditcoin node with a controllable finalized head, per-connection subscription silencing, socket closing and pruned-state errors. The proof-gen tests from #1350 now use it instead of a local copy.

The attestor consumes the same stream and inherits the watchdog with the default timeout; no attestor code changes.

## Tests

`common/streams/cc3/tests/watchdog.rs`:
- silent subscription while the node keeps finalizing → replaced with exactly one redial, heights 2 and 3 arrive gap-free, `silent_recoveries == 1`
- stalled finality → no redial across several deadlines, resumes on the same socket
- a subscription that never delivers its first block → seed redials instead of hanging

`cargo test -p stream_cc3 -p cc-client -p proof-gen-api-server` green; clippy `-D warnings` on cc-client (both feature sets), stream_cc3, proof-gen-api-server, attestor; fmt, taplo, machete.


DylanVerstraete:
CI note: `integration-test-attestator-network` failed twice here for two different reasons. Attempt 1: the attestors never produced the genesis attestation (the recurring bootstrapping stall seen on several PRs today; no code path here is involved). Attempt 2: the attestors bootstrapped and every checkpoint compare passed, but the archiver log gate flagged one line: the archiver's attested-height poller (usc-dev, #1366) re-dialled Creditcoin while the job's deliberate CC3 restart had the node down and logged `WARN Creditcoin reconnect failed ... Connection refused (os error 111)`. The gate matches `error:` case-insensitively anywhere in a line and the subxt error text contains it. This stack's cc3 client detects the dead socket faster (ws ping/pong), which is why it surfaced here first. Added 9d8f7664d: that one message from that one source is allowlisted in the gate step, scoped like the existing eth-restart noise. The three failing `integration-test-cli` matrix jobs are the `full validator cycle` setup-hook timeout (20 s to connect to the node) that also hit #1345 today; not related to this PR.

didac-gluwa:
Small note on the progress record — not a blocker, reading the stack end to end.

`p.note(n)` in `common/streams/cc3/src/lib.rs` runs before the `backfill.drain(..).rev()` yield loop, so `cc3_finalized_height` reports the head of a batch before any block in that batch has reached the consumer. The comment above it makes the tradeoff explicit (leading by one batch rather than lagging by one block), and staleness detection is unaffected either way, since a wedged consumer stops polling the generator altogether.

Just worth keeping in mind that the health value is an *emit* height, not an *applied* height — #1353 keys readiness off the same record.

Also checked the ws client: `build_rpc_client` does match what subxt builds (`subxt-rpcs-0.44.3/src/client/jsonrpsee_impl.rs:107`, same `max_buffer_capacity_per_subscription(4096)` over `WsTransportClientBuilder::default()`), and the watchdog's `finalized_head_number()` probe is bounded by jsonrpsee's default 60s `request_timeout` even though it has no explicit deadline of its own.

didac-gluwa (dismissed):

DylanVerstraete:
Agreed on both counts, thanks for reading it end to end. The progress record is an *emit* height by design: it is what tells a watchdog the generator is alive and moving, and a wedged consumer stops polling the generator so staleness still trips. I have made the same distinction explicit in the #1353 description so nobody reads `cc3_finalized_height` as "applied". Good to have the ws client builder and the 60 s jsonrpsee request timeout on the probe confirmed independently.

didac-gluwa (approved):

didac-gluwa (approved):

Issue #1355, opened 2026-09-11
Attestor halts permanently on any chain without a receipts trie, and reports it as a reorg (reproducible on exSat, chain id 7200)


zhenek73:
## Summary

An attestor cannot follow any EVM chain that does not maintain a receipts trie. It walks
through empty blocks fine, then stops **permanently** at the first block containing a
transaction, and reports the failure as a reorg that never happened.

This is not hypothetical — it is reproducible today against exSat's EVM layer (chain id
7200), and the misleading error cost us a full day of debugging in the wrong direction.

## Reproduction

Point an attestor at `wss://evm.exsat.network/` and let it run:

```
ERROR stream_eth::roots: Eth connection error
      err=Computed transactions/receipts roots do not match block header for block 59226042
      (possible reorg between RPC calls)
WARN  stream_eth::roots: Reconnecting to Eth...
```

It repeats on the identical block number every ~2 minutes, indefinitely, across restarts.
`last_attested_block` never moves again.

## Root cause

exSat does not maintain a receipts trie or a state trie. Every header it serves reports
`receiptsRoot` and `stateRoot` as 32 zero bytes — on every block, with or without
transactions:

```console
$ curl -s -X POST https://evm.exsat.network -H 'Content-Type: application/json' \
    -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x38EF2C8",false]}'
# block 59801416 — 1 transaction
#   transactionsRoot 0x4f8a21e5839d3c1e9249ce90a972119380cce4a477ad23027058bfb2d38ae7bf
#   receiptsRoot     0x0000000000000000000000000000000000000000000000000000000000000000
#   stateRoot        0x0000000000000000000000000000000000000000000000000000000000000000
```

Checked across many consecutive blocks — structural, not one bad RPC replica.

In `EthBlock::try_create` (`common/eth/src/lib.rs`) the recomputed receipt root is compared
to the header. Empty blocks are unaffected because of the early return above it. Any block
with at least one transaction produces a real non-zero computed root, which can never equal
a permanent zero, so it is rejected as `BlockHeaderRootsMismatch` — surfaced to the operator
as a suspected reorg.

## Why the diagnosis is expensive

The error names the wrong cause. A "possible reorg between RPC calls" sends an operator to
look at RPC stability, load balancers and chain health, none of which are the problem. It
also looks transient, so the natural reaction is to wait or restart — and both appear to
[...]

PR #1357, opened 2026-09-13
fix(eth): skip the receipt-root check for exSat, which never populates it

zhenek73:
Reported in #1355.

## What happens today

An attestor pointed at exSat's EVM layer (chain id 7200) advances through empty blocks
and then stops permanently at the **first block containing a transaction**, repeating
this forever:

```
ERROR stream_eth::roots: Eth connection error
      err=Computed transactions/receipts roots do not match block header for block 59226042
      (possible reorg between RPC calls)
WARN  stream_eth::roots: Reconnecting to Eth...
```

It is not a reorg, and it is not transient. It repeats on the identical block number
indefinitely, across restarts, and would recur on every subsequent transaction-bearing
block. `last_attested_block` never moves again.

## Why

exSat's EVM layer does not maintain a receipts trie or a state trie. Every block header
it serves reports `receiptsRoot` and `stateRoot` as 32 zero bytes — on every block,
whether or not it has transactions:

```console
$ curl -s -X POST https://evm.exsat.network -H 'Content-Type: application/json' \
    -d '{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["0x38DB18A",true]}'
# block 59801418 — txCount 1
#   transactionsRoot 0x4f8a21e5839d3c1e9249ce90a972119380cce4a477ad23027058bfb2d38ae7bf
#   receiptsRoot     0x0000000000000000000000000000000000000000000000000000000000000000
#   stateRoot        0x0000000000000000000000000000000000000000000000000000000000000000
```

Verified across many consecutive blocks, with and without transactions — this is
structural, not one bad RPC replica.

`EthBlock::try_create` (`common/eth/src/lib.rs`) recomputes the receipt root and compares
it to the header. For an **empty** block the comparison is skipped by the existing
early return, so those pass. For a block with at least one transaction the recomputed
root is a real non-zero hash which can never equal a permanent zero, so every such block
is rejected as `BlockHeaderRootsMismatch` — surfaced to the operator as a reorg, which
sends debugging in exactly the wrong direction.

## Why skipping the check is safe here

Three things were verified independently against the live chain before proposing this:

1. **`transactionsRoot` is genuine and canonical on exSat.** Recomputing the
   Merkle-Patricia root from each block's own transactions matches the header byte for
   byte (checked on both the block that stalls the attestor, 59226042, and a later one,
   59801416). The transaction-root check above therefore still does its job, including
   guarding against reorg-induced cross-fetch mismatches — the same reasoning the
   existing pre-Byzantium exemption already relies on.

2. **exSat's headers are internally self-consistent.** The reported block hash equals
   `keccak(rlp(header))` computed *including* the zero roots, so the header hash chain is
   sound and nothing about it is ambiguous.

3. **The receipts root is not part of anything attested or proven.** In this repository
   `receipts_root` appears in exactly one place — the comparison this PR touches. The
   attested digest is `Block::hash_payload(block_number, root, prev_digest)`, where `root`
   is the merkle root the inclusion proofs are built on. Skipping the comparison changes
   no attestation, no digest and no proof — only whether the block is accepted for
   processing at all.

## The change

Extends the existing `skip_receipt_root` exemption — already present for pre-Byzantium
Ethereum mainnet, for an unrelated but structurally identical reason — to cover exSat, and
expands the comment to explain both cases. No behaviour changes for any other chain.

```rust
const EXSAT_CHAIN_ID: u64 = 7200;

let skip_receipt_root = (chain_id == ETHEREUM_MAINNET_CHAIN_ID
    && expected_number < ETHEREUM_BYZANTIUM_BLOCK)
    || chain_id == EXSAT_CHAIN_ID;
```

## Testing

`cargo check -p attestor` passes with this change applied (clean checkout of `usc-dev`,
patch on top). Note that `cargo check -p eth` on its own fails both with and without this
patch — the crate does not build in isolation because `tokio`'s `signal` feature is enabled
by a sibling crate — so the meaningful check is the one through `attestor`.

## Alternative, if you prefer it

Hard-coding a second chain id follows the existing style, but a per-chain
`skip_receipt_root_check` flag in the source-chain config would generalise better — exSat
is unlikely to be the last L2 that skips the receipts trie. Happy to rework it that way;
this version was kept minimal to match what is already there.

## Context

Found while building [Bitcoin Witness](https://github.com/zhenek73/bitcoin-witness), which
attests exSat EVM blocks in order to prove Bitcoin UTXO facts on Creditcoin. Until this is
fixed upstream, running an attestor against exSat requires a proxy that fills the zeroed
field in from the block's own receipts, which is a workaround nobody should need.


DylanVerstraete:
We won't support exSat

PR #1359, opened 2026-09-14
ci: seed configuration-specific Rust caches from usc-dev

DylanVerstraete:

---

When merging this PR:

---

Practical tips for PR review:


DylanVerstraete:
Closing in favour of #1338, which reaches the same diagnosis independently and got there first.

#1338 already carries most of what was here: a `push` trigger on usc-dev for `rust.yml`, per-job cache keys for check, clippy and test, and a default-branch producer for the native cache. Its treatment of the three cheap jobs is better than this branch's, too: it leaves `cargo-fmt`, `cargo-audit` and `cargo-machete` on the shared key with `save-if: false`, so they restore but never squat it, where this PR spent two extra cache entries to reach the same result. The two branches also conflict in `rust.yml` and `build-native.yml`, and each added a rival warming workflow.

The two pieces of this PR that #1338 did not have are now on that branch as 58410b30d:

- the self-hosted (Linode) warmer, which is #1338's own documented "known gap" and becomes the only warmer that matters once #1360 removes the compatibility builds;
- folding the runner family and `$HOME` into the native shared key, without which the hosted and Linode warmers compute the same immutable key and one of them can never save.

Everything else here is superseded. No objection to reopening if #1338 stalls.

PR #1367, opened 2026-09-15
feat(eth): resolve block tags by cross-provider agreement, keyed on the block hash

DylanVerstraete:
> **Stack** (bottom → top; each PR's diff is only its own commits, review in this order)
>
> | | PR | Phase | What | Base |
> |---|---|---|---|---|
> | &nbsp;&nbsp; | #1330 | Phase 0 | RpcSafe / RpcFinalized maturity strategies | `usc-dev` |
> | &nbsp;&nbsp; | #1362 | Phase 1a | receipts must carry the fetched block's hash | #1330 |
> | &nbsp;&nbsp; | #1365 | Phase 2b | prover confirms against the attestation cache | #1362 |
> | &nbsp;&nbsp; | #1366 | Phase 2a | archiver follows the latest attested height | #1365 |
> | **▶** | #1367 | Phase 1c | block tags resolved by cross-provider agreement | #1366 |
>
> Plan: https://claude.ai/code/artifact/400f8668-7b43-4b11-86c8-b8922d8c85ee · Live run against Sepolia `safe` via Chainstack (2026-09-15): attested heights tracked `safe` within one interval, archiver ≤ attested ≤ safe at every sample, prover 200 at attested / 422 one above; survived a 20-min RPC outage without restarts.

Phase 1c, plus the second half of 1a, of the plan to make the attested set the single maturity boundary: https://claude.ai/code/artifact/400f8668-7b43-4b11-86c8-b8922d8c85ee

## What was there

`get_block_number_by_tag` walked the provider list and returned the first block any provider served. It also received the full block from the RPC and returned only its number. For the one call that decides the attestor's maturity boundary that is the wrong shape twice: a single load-balanced peer on a stale fork could set the boundary for the whole attestor, and the one datum that pins block identity was discarded at the one place it was free.

## What is there now

`get_block_by_tag` returns `TaggedBlock { number, hash }` and asks **every** configured provider at once.

1. The lowest reported height is the candidate, because a provider further along has by definition already passed it.
2. Every other responder must have the candidate block under the same hash. Same-height responders already said so; further-along responders are asked for the block at that height.
3. A different hash is `Error::BlockTagDisagreement` and the lookup refuses to pick a side. The tip and roots streams already log and retry on the next head, so a transient fork view costs one head, not an attestation.

Two edge cases are decided explicitly. A provider that reports the tag past a height it cannot serve is broken rather than forked, and fails the lookup. A provider that errors on the confirmation read has already voted the candidate mature by reporting a higher tag, so it is warned about and skipped, the same treatment a tag-call error got before.

Providers that error or answer `null` still do not take part. **With one provider this degenerates to that provider's answer**, so this PR is the code half of a two-part change: per-replica provider diversity in the AttestorSet (1b, cc-networks-iac) is what gives it teeth. Today every replica in a set reads the boundary from one `ethUrlSecretRef`.

`get_block_number_by_tag` stays as a thin wrapper, so `Maturity` and the streams are untouched.

## Why the hash stops here

The plan said "carry the hash through `Maturity` and fetch by `BlockId::Hash`". I stopped short deliberately, as in #1362: fetching by hash serialises `get_block` and `get_block_receipts` and roughly halves archiver throughput, and #1362's receipts-to-block hash check already pins each fetched block's integrity for free. The hash is used where it decides something, the boundary identity across providers, and not threaded through code that has no use for it.

## Also removed

`EthRpcProvider::get_block_number_by_tag` in the continuity crate, its three impls and its test. Dead since #1365 took the prover's maturity opinion away.

## Tests

`common/eth/tests/block_tag.rs` mock now salts block hashes, so two mocks can be on one chain or on different forks. New: agreement takes the lowest height and confirms it against the further provider, and both are asked every time; a forked provider is a disagreement at different heights and at the same height; the disagreement surfaces through `Maturity` as a per-head error. Existing fall-through and clamping tests unchanged. Clippy `-D warnings` across eth, continuity, proof-gen, archiver, attestor, stream_eth; fmt; 189 tests green across those crates.


DylanVerstraete:
pre-commit.ci run

common/eth/src/lib.rs
```diff
@@ -968,34 +988,145 @@ impl Client {
...
-        for (label, provider) in providers {
-            match provider
-                .get_block(BlockId::Number(tag.into()), false.into())
-                .await
-            {
-                Ok(Some(block)) => {
-                    for (err_label, err) in errors.drain(..) {
+        if let Some((candidate_label, candidate)) = reported
+            .iter()
+            .min_by_key(|(_, block)| block.number)
```
beqaabu: A stale fallback provider permanently freezes maturity for the whole pipeline. `common/eth/src/lib.rs:1044` picks the candidate via `min_by_key(block.number)` across all responders. A backup RPC that stops syncing but keeps serving answers safe = N forever; its block at N hashes identically to canonical, so the agreement check passes and returns `Ok(N)`. `common/streams/eth/src/tip.rs:81` only advances on `tip_new > tip`, so the tip pins at N indefinitely even though the primary is healthy. Verified both ends. This inverts what a fallback means — before this PR a fallback was only consulted when the primary errored, so a stale backup was harmless; now it's a hard brake. There's no staleness bound, no warning (only debug!), and no metric.

common/eth/src/lib.rs
beqaabu, line 1: No timeout anywhere — one black-holed provider hangs the tip stream forever. get_block_by_tag does join_all over all providers, and grep -E 'timeout|Timeout' common/eth/src/lib.rs returns nothing; init_rpc builds providers with a default reqwest client. A fallback that accepts TCP but never responds makes the lookup never return, and tip.rs awaits it inline in the stream loop — no reconnect, no error, no log. A per-provider tokio::time::timeout treated like the tolerated Err branch at lib.rs:1106 fixes it

common/eth/src/lib.rs
DylanVerstraete: You're right, and it's the more serious of the two: lowest-height turned every fallback into a brake. Fixed in e87dee4e3. The primary's answer is the candidate (the freshest fallback's when the primary has none), and the others confirm its identity: same height, compare hashes; further along, fetch the block at the candidate height and compare. A provider that hasn't reached the candidate cannot confirm and cannot veto; it is logged at warn with how far behind it is, so a persistently stale fallback is visible. Hash mismatch at the candidate height stays a hard `BlockTagDisagreement`. Resolution without full agreement now logs at warn, not debug. Tests: fallback ahead confirms, fallback behind steps aside, stale fallback cannot pin with and without a primary answer.

common/eth/src/lib.rs
DylanVerstraete: Fixed in e87dee4e3: every provider call in the lookup goes through a `timed` wrapper bounded by `Client::call_timeout` (15 s default, one Ethereum slot; `with_call_timeout` to override), and a timeout is handled exactly like the tolerated Err branch. Test uses a mock that answers `eth_chainId` and black-holes everything else; the lookup returns the primary's answer in under the timeout. Worth saying: `init_rpc` still has no timeout, so a black-holed URL hangs client *construction*; that is pre-existing and not on the tag path, I'd take it as a separate fix if you want it.

beqaabu (dismissed):

beqaabu (dismissed):

common/streams/eth/src/roots.rs
```diff
@@ -326,8 +335,11 @@ async fn stream_rpc(
     // head, gaps backfilled; with a block tag or an attested bound the bound moves in jumps and
     // a whole range is released at once. A bound that cannot be resolved for a head is skipped:
     // the next head retries, and the walk guarantees no block is skipped or fetched twice.
-    let heads = futures::stream::once(futures::future::ready(next))
-        .chain(stream_headers.map(|header| header.number));
+    let heads = futures::stream::once(futures::future::ready(next)).chain(end_on_silence(
+        stream_headers.map(|header| header.number).boxed(),
+        config.client.clone(),
+        config.head_silence_timeout,
+    ));
```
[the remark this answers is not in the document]
DylanVerstraete: Same finding as on #1366, fixed there in 8930f0471 (this branch is rebased on it): the first head seeds the watchdog's baseline, and a silent-from-start stream adopts its first probe as the baseline.

archiver/src/main.rs
```diff
@@ -376,12 +376,20 @@ async fn main() -> Result<()> {
...
                     let next_wanted = last_height.map(|h| h + 1).unwrap_or(start_height);
+                    // The stream fetches up to the published bound clamped to the source head
+                    // it has observed; judge the silence against the same number, or a node
+                    // lagging the attestors would look like a stall and be torn down every
+                    // timeout while it is healthy and simply behind.
                     let published = *rx.borrow();
-                    if published.is_none_or(|bound| bound < next_wanted) {
+                    let source_head = chain_head.load(Ordering::Acquire);
+                    let fetchable = published.map(|bound| bound.min(source_head));
+                    if fetchable.is_none_or(|bound| bound < next_wanted) {
```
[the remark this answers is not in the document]
DylanVerstraete: Same finding as on #1366, fixed there in a8327a813 (this branch is rebased on it): a zero or stale HTTP head is treated as unknown and never clamps, so it cannot hide a hung fetch.

beqaabu (dismissed):

BradleyOlson64 (dismissed):

beqaabu (dismissed):

beqaabu (dismissed):

beqaabu (dismissed):

didac-gluwa (dismissed):

common/eth/src/lib.rs
```diff
@@ -968,34 +1035,184 @@ impl Client {
...
+                .filter(|(label, _)| *label != candidate_label)
+                .map(|(label, block)| {
+                    let provider = providers
+                        .iter()
+                        .find(|(l, _)| l == label)
+                        .map(|(_, p)| *p)
+                        .expect("label came from this provider list");
+                    async move {
+                        if block.number < candidate.number {
+                            return (label.clone(), Confirmation::Behind(block.number));
```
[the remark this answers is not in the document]
DylanVerstraete: Fixed in 03e28ea50. Every non-candidate responder is now asked for the block at the candidate height and compared by hash, whatever its own tag says. `Behind` is reserved for a lagging provider whose `eth_getBlockByNumber` at that height returns null, so it can still abstain but never skips the check when it has the block. New test `a_forked_primary_ahead_of_honest_fallbacks_is_still_challenged` pins the exact scenario (forked primary at 90, honest fallback tag at 88 → `BlockTagDisagreement` at 90); the stale-fallback test now uses a mock capped at head 45 so it exercises the genuine "no block there" path.

didac-gluwa (dismissed):

BradleyOlson64 (dismissed):

DylanVerstraete:
Rebased onto usc-dev after #1330 and #1362 landed as squash commits (old head backed up locally). Only the already-merged commits were dropped; the code is byte-identical to the previously approved head apart from `.github/workflows/ci.yml`, which now sits on top of #1338/#1379's CI changes with #1366's archiver wait steps intact. fmt, clippy (-D warnings) and the eth/archiver/stream_eth/continuity/proof-gen test suites pass locally on the rebased stack top. Per Dylan, the PRs will be merged without re-collecting approvals since the content was approved before the rebase.

PR #1377, opened 2026-09-18
ci(timing): report jest suite and hook timings, not just job duration

beqaabu:
## Why

`ci-timing.py` stops at the job. That is enough to notice that `integration-test-cli` takes
63 minutes, and not enough to do anything about it. A job number cannot say whether a change
worked or whether the run simply landed on a fast runner, and it cannot say which of the
sixteen suites the time is in.

That matters right now because #1376 makes three specific, falsifiable claims about where
that job's time goes, and this is what checks them.

## What it adds

`suites <run-id>` pulls the jest `--verbose` output out of each matching job's log and
reports, per suite: the total, the time spent inside tests, and the difference — the time in
`beforeAll` / `beforeEach` / `afterAll`. That last column is the one worth having. Against
the current `usc-dev` baseline it prints, without anyone having to go looking:

```
integration-test-cli / proxy=yes / no-funds / type=
  job 44.2m, jest 2521s over 11 suites
  suite                       total    tests    hooks  hooks%  ok skip
  withdraw-unbonded         1161.9s     1.4s  1160.5s    100%   1    4
  distribute-rewards         327.9s    89.1s   238.8s     73%   1    4
  ...
```

Nineteen minutes of setup for a 1.4 second assertion, invisible in the 44m job duration
sitting next to it.

`compare-suites before.json after.json` diffs two of these. A suite that disappears is
reported as `SKIPPED` rather than as missing data: when every test in a file is skipped on a
leg, jest runs none of its hooks either, so the suite vanishing *is* the saving.

```
.github/ci-timing.py suites <baseline-run-id> --out before.json
.github/ci-timing.py suites <pr-run-id>       --out after.json
.github/ci-timing.py compare-suites before.json after.json
```

## Validation

A measurement tool that is wrong is worse than none, so:

- **It reproduces hand-derived numbers exactly.** Against run `35284073101`, the parser
  returns withdraw-unbonded `1206.7 / 53.0 / 1153.7`, proxy `465.0 / 451.2 / 13.8`,
  distribute-rewards `408.2 / 169.0 / 239.2` — the same figures I previously read out of
  those logs by hand — and its per-job totals agree with the `Time:` line jest prints
  (3809s, 2521s).
- **Null comparison is null.** Diffing two unrelated PRs' runs of the same suites gives every
  delta within **6s**, on suites ranging from 16s to 1207s. The instrument's noise floor is
  well under 1% at suite level, against the hundreds of seconds a real change moves. That is
  what makes a single run's result trustworthy.

## Details worth knowing

- Both durations in jest's output are optional — it omits the suite time for fast suites and
  the per-test time for sub-millisecond tests — so both are optional in the regexes rather
  than assumed present.
- Unreadable logs warn and are skipped rather than aborting the run: logs expire, and a run
  with one unreadable job is still worth reporting on. A run with no jest output at all exits
  with a clear message.
- `--job` defaults to `integration-test-cli`, since that is the job this exists for, and takes
  any substring. `--job ''` parses every jest job in the run, which also covers
  `attestor-cli-testing`, `cc3-indexer-testing`, `integration-test-blockchain` and the
  archiver suites.

No workflow changes: this is a local analysis script, like the rest of `ci-timing.py`.


didac-gluwa (dismissed):

DylanVerstraete (dismissed):

mason-gluwa (dismissed):

beqaabu:
Good catch, this was real and I've fixed it in `2d7eed065`.

Reproduced it first: a two-test suite with no reported duration gave `hooks_s = -0.8`. The regexes already treated both durations as optional, which was right — the arithmetic downstream just didn't honour it.

An unreported duration is now `None` rather than `0.0`, `hooks_s` is `None` when the total is unknown, and both tables print `-` instead of a fabricated number. `compare-suites` shows `?` for a delta it can't compute rather than inventing a baseline.

Worth noting it was reachable in practice, not just theoretically: `--job ''` is documented as parsing every jest job in a run, and the unit-test suites finish in milliseconds so they report no suite time.

Verified the fix doesn't move anything real — re-running against run 35284073101 gives byte-identical totals, tests and hooks across all 6 jobs and every suite.

PR #1388, opened 2026-09-19
Allow maturity strategies to be updated at runtime

BradleyOlson64:
Changes:
1. Add `setMaturityStrategy` in pallet supported chains
2. Add listen for maturity strategy changes in the attestor and archiver. Reset in-flight attesting on a maturity change.
3. Indexing for events


attestor/attestor/src/tasks/validation.rs
```diff
@@ -173,7 +179,7 @@ async fn handle_quorum(
     if in_flight.is_none() {
         pool_rx.mark_valid(permit);
         *in_flight_height = Some(agg.height);
-        *in_flight = Some(spawn_submission(shared.clone(), agg));
+        *in_flight = Some(spawn_submission(shared.clone(), agg, maturity_gen));
     } else {
         tracing::info!(?digest, height, "🗃️ stash for later");
         pool_rx.mark_for_later(
```
DylanVerstraete, line 179: **Blocker (Bugbot's resolved high-severity finding is still live).** This permit was taken before `aggregate_and_validate` awaited RPC. If `MaturityStrategySet` lands in that window while another submission is in flight, we arrive here and `mark_for_later` pushes `signed` into the pool that was *just reset* — `Pool::mark_for_later` does `pool.valid.push(signed); pool.mark_valid(permit)` with no `validation_generation` check (unlike `defer`). When it is drained below, `stash_maturity_gen` is snapshotted *fresh* (post-change), so `superseded()` is false and the old-strategy aggregate submits.

Two ways out: carry the dequeue-time `maturity_gen` with the stash (e.g. wrap `SignedQuorum` or add a field) and use *that* at drain time instead of re-snapshotting; or make `mark_for_later`/`mark_valid` reject a permit whose `validation_generation` is stale, the way `defer` already does. The first keeps the pool oblivious to maturity, which matches the rest of the design.

attestor/attestor/src/tasks/validation.rs
```diff
@@ -334,14 +368,18 @@ async fn handle_submission_result(
     while let Some(stashed) = pool_rx.take_next_validated() {
         let stash_height = stashed.height;
         let stash_digest = stashed.digest;
+        // Same reasoning as the dequeue site: `revalidate_stashed` awaits RPC, so the snapshot
+        // has to precede it. Taken per stash rather than once for the whole drain so a stash
+        // that was itself queued after the change is not judged against a pre-change value.
+        let stash_maturity_gen = current_maturity_gen(shared);
```
DylanVerstraete: Follow-on to the comment at the `mark_for_later` site: re-snapshotting here is what lets a stash queued *before* the change look current. If the snapshot travels with the stash this becomes `stashed.maturity_gen` and the race closes.

attestor/attestor/src/tasks/production.rs
```diff
@@ -505,6 +529,87 @@ async fn handle_one(
...
+            // Publish before rebuilding: `build_stream_attestation` reads it.
+            *shared.maturity.write() = maturity;
+
+            // Drop everything derived from the old maturity. This mirrors the reversion handler
+            // below — same invalidation shape, different trigger — except that the attestation
+            // chain itself has not moved, so `latest_cc3` stays as it is.
+            shared.proof_cache.clear();
+            shared
+                .pool_send
+                .note_attestation_chain_reversion(latest_cc3.height, latest_cc3.digest);
```
DylanVerstraete: **Blocker: a strategy change while we are still waiting for the genesis `BlockAttested` stalls the bootstrap.** This arm also runs from `wait_for_block_attested` (via `handle_one`). There the pool reset drops *our own* genesis vote (inserted by `emit_local` → `pool_send.send`), `proof_cache.clear()` drops the proof, and nothing re-emits it: the outbox re-gossips it to *peers* every 30 s, but gossipsub never loops a node's own message back, so every node ends up one vote short. With quorum equal to the set size (what CI runs, and small production sets) no node ever reaches quorum — exactly the `genesis BlockAttested still missing` stall. This is a realistic sequence: register chain → start attestors → operator sets `RpcSafe`.

Suggest: while bootstrapping (`latest_cc3` is the `{genesis, ZERO}` anchor), only rebuild the streams and skip the pool/cache reset — nothing maturity-derived exists yet — or re-run the genesis emission right after the reset.

attestor/attestor/src/tasks/production.rs
```diff
@@ -646,11 +751,17 @@ async fn wait_for_block_attested(
...
-    // authoritative `latest_cc3` from this function's return value.
-    let mut scratch = AttestationInfo::default();
+    // Scratch finalized cursor for `handle_one`. We never route `BlockAttested` through it, and
+    // the caller keeps the authoritative `latest_cc3` from this function's return value. It is
+    // seeded at the genesis height rather than zero because two arms *read* it: the reversion
+    // arm (which also writes it) and the maturity-strategy arm, which rebuilds the eth pipeline
+    // anchored on it. `{genesis, ZERO}` is the same anchor `run` builds the pipeline from before
+    // the first attestation finalizes, so a strategy change during the bootstrap wait rebuilds
+    // to exactly where we already were instead of rewinding the streams to block 0.
+    let mut scratch = AttestationInfo {
```
DylanVerstraete: Seeding the scratch at genesis fixes where the rebuild anchors, but see the comment on the pool reset above: during this wait the reset itself is the problem, not the anchor.

attestor/attestor/src/tasks/production.rs
```diff
@@ -505,6 +529,87 @@ async fn handle_one(
...
+        // it, every attestation still in production was built against it, and every vote we hold
+        // (ours and our peers') is for a height the network may no longer agree is mature.
+        // Rebuild the eth pipeline from the latest on-chain attestation and drop the rest. What
+        // is already committed on chain is untouched — `latest_cc3` is exactly that boundary, so
+        // no attested height is ever revisited.
+        CcEvent::MaturityStrategySet(_, strategy_string) => {
+            let strategy: supported_chains_primitives::MaturityStrategy = strategy_string
+                .as_str()
+                .try_into()
+                .map_err(|e| Error::InvalidMaturityStrategy(shared.chain_key, e))?;
```
DylanVerstraete: Design note, not a blocker: an unparseable/unresolvable strategy makes `handle_one` return `Err`, which ends the production task, and startup fails the same way (`lib.rs:198-200`), so the attestor is down until it is upgraded. Fail-closed is defensible for an attestor and the archiver deliberately chose the opposite, but the asymmetry should be spelled out in the operator runbook ("upgrade attestors before introducing a new strategy kind") and the error message should say that is what happened.

archiver/src/main.rs
```diff
@@ -371,9 +380,160 @@ async fn main() -> Result<()> {
...
+                        // stale tail would wedge the stream rather than heal.
+                        //
+                        // The cut is floored at the latest attested height. Attestations already
+                        // committed on Creditcoin stay valid under the new strategy, so the
+                        // roots backing them have to survive even where the new strategy would
+                        // now call their blocks premature — proof-gen serves exactly that range.
+                        // A floor we cannot read is a floor we cannot respect, so a failed read
+                        // leaves the tail alone rather than risking an attested root.
+                        let head = known_head(&chain_head, &head_seen_at)
+                            .unwrap_or_else(|| chain_head.load(Ordering::Acquire));
```
DylanVerstraete: **Blocker: this can wipe the archive.** `known_head` returns `None` when the head was never read *or* is stale, and the fallback is the raw counter, which is `0` in the never-read case. Then `FixedLag(n).mature_height(_, 0)` is `None` and `Tag(..).mature_height(_, 0)` clamps to `Some(0)`; either way `keep_to` below collapses to the attested floor — or to `0` when the chain has no attestation yet (an `END_HEIGHT` sweep with `CHAIN_KEY` on a chain whose attestors have not started) — and `remove_above` deletes everything above it. The comment on `mature` ("`None` only when nothing is mature at all") does not hold for an unknown head.

Treat an unknown head like the two `Err` branches: rebuild the stream, skip the trim, log why.

archiver/src/main.rs
```diff
@@ -371,9 +380,160 @@ async fn main() -> Result<()> {
...
+                        let new_boundary = stream_eth::roots::Boundary::Source(new_maturity);
+
+                        // Same bounded-backoff dial as the stream-reconnect path below: a WS
+                        // blip while rebuilding must not kill an archiver that is otherwise
+                        // healthy. `boundary` is only moved once the new stream exists, so a
+                        // failure here cannot leave the published boundary describing a stream
+                        // that was never built.
+                        let mut delay = RECONNECT_BASE_DELAY;
+                        let new_ws = loop {
+                            match eth::Client::new(cfg.rpc_ws.as_str(), None).await {
```
DylanVerstraete: Coordination note: the archiver liveness stack (#1348) adds `--rpc-fallback-urls` and builds the stream client with them; this rebuild path uses a bare `eth::Client::new(cfg.rpc_ws, None)`. Whoever lands second should route the rebuild through the same constructor as startup, or a strategy change silently drops the fallbacks.

archiver/src/store.rs
```diff
@@ -218,6 +218,59 @@ impl RootStore {
...
+    /// the archiver had already computed can then sit above the new mature height, where the
+    /// source block is still re-organisable. Leaving them would be worse than not having them —
+    /// `put_roots` treats a later canonical replacement at an occupied height as a hard
+    /// [`StoreError::ReorgDetected`], and until that fires the API would serve a root for a block
+    /// that no longer exists. Deleting them lets the stream re-derive the range once the new
+    /// strategy considers it mature.
+    ///
+    /// Heights at or below `height` are untouched: attestations already committed on Creditcoin
+    /// stay valid under the new strategy, so the roots backing them must survive.
+    pub fn remove_above(&self, height: u64) -> Result<usize> {
```
DylanVerstraete: Overlaps with `RootStore::truncate_above` that #1347 (canonical-anchor check, bounded self-heal) adds for the reorg case. Same operation, two names; please merge them into one when the stacks meet — the persisted-count handling here is the more careful of the two and should win.

cc3-indexer/src/mappings/mappingHandlers.ts
```diff
@@ -196,6 +197,52 @@ export async function handleSupportedChainRegistered(event: SubstrateEvent): Pro
...
 
+export async function handleMaturityStrategySet(event: SubstrateEvent): Promise<void> {
+    const {
+        event: {
+            data: [chainKey, chainId, maturityStrategy],
+        },
+    } = event;
+
+    const from = event.extrinsic?.extrinsic.signer;
+    assert(from, 'Signer is missing');
```
DylanVerstraete, line 208: Consistent with the sibling handlers, so not a blocker, but note this asserts a signed extrinsic: if `set_maturity_strategy` ever arrives through a non-signed origin (scheduler, governance) the indexer crashes on the assert instead of recording the row with an unknown `who`.

pallets/supported-chains/src/tests.rs
```diff
@@ -554,3 +554,208 @@ fn register_chain_rejects_invalid_maturity_strategy(#[case] strategy: String) {
         assert!(chain_key.is_none());
     });
 }
+
+// ------------------------------- [ set_maturity_strategy ] ---------------------------------- //
+//
+// The genesis chain from `ExtBuilder` is chain key 1 (chain id 200, "Ethereum") registered with
+// `MATURITY_FIXED_DELAY_10`.
+
+const TEST_CHAIN_KEY: u64 = 1;
```
DylanVerstraete, line 563: Pallet coverage is thorough (invalid strings, unprivileged origins, unchanged rejection, cross-chain isolation, respelling). The attestor side has only the `should_unlock_if_unfinalized` predicate test; at minimum the "restated strategy resolves the same → no rebuild" branch and the bootstrap case flagged in `production.rs` deserve tests.

DylanVerstraete (commented):
Read the whole diff plus the pool, the archiver head tracking and the maturity resolver it leans on. The design is right and well argued: operator-gated extrinsic, event only on a real change, attestors drop everything derived from the old maturity and rebuild from the latest attestation, archiver reacts only when it resolves maturity itself. The generation-snapshot placement in the validation task is correct, the `Superseded` semantics are well reasoned, and the archiver decision table with its tests is solid. Pallet tests are thorough.

Three correctness issues need fixing before this merges (inline):

1. **A pre-reset quorum can still be submitted under the old strategy** — Bugbot's high-severity thread was resolved without a code change; `mark_for_later` has no generation check, so a stash queued across the reset is drained with a fresh snapshot.
2. **A strategy change during bootstrap stalls genesis** — the pool reset drops our own genesis vote and nothing re-emits it; gossipsub never loops our own message back, so with quorum = set size nobody ever reaches quorum.
3. **The archiver trim can wipe the archive when the head is unknown** — `known_head` → `None` falls back to `0`, which collapses `keep_to` to the attested floor (or 0).

Smaller points inline as well (fail-closed attestor on an unknown strategy, overlap with the archiver liveness stack's `truncate_above` and fallback URLs in #1347/#1348, no test on the production arm, indexer signer assert). Leaving the verdict to Dylan; from my side the three numbered items are blockers.

cc3-indexer/src/mappings/mappingHandlers.ts
BradleyOlson64: Not a problem. Should always be signed.

didac-gluwa (dismissed):

attestor/attestor/src/tasks/validation.rs
BradleyOlson64: Not a concern with our new targeted approach of restarting attestors after maturity update.

BradleyOlson64:
All bot comments are now outdated or addressed. Rebased and ready for another review

DylanVerstraete (dismissed):

didac-gluwa (approved):

pallets/supported-chains/src/lib.rs
```diff
@@ -149,6 +150,16 @@ pub mod pallet {
...
 
+        /// The maturity strategy of a registered chain has been changed. Off-chain consumers
+        /// (attestors, archivers) read the strategy once at startup and do not react to this
+        /// event, so it is a record of the change for indexers and operators rather than a
+        /// signal any node acts on. See `set_maturity_strategy` for what operators must do.
+        MaturityStrategySet {
+            chain_key: ChainKey,
+            chain_id: ChainId,
+            maturity_strategy: String,
+        },
```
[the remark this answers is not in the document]
BradleyOlson64: Fixed

cc3-indexer/schema.graphql
```diff
@@ -570,6 +570,20 @@ type ChainRemoved @entity {
...
+type MaturityStrategySet @entity {
+    id: ID!
+    # Chain key
+    chainKey: BigInt! @index
+    # Chain id
+    chainId: BigInt! @index
+    # The maturity strategy the chain now uses
+    maturityStrategy: String!
+    # At Block
+    at: BigInt!
```
jakerumbles: Is this leftover from the initial design of having the attesters update themselves where the new maturity strategy takes effect at a specific block in the future?

PR #54, opened 2023-11-08
CSUB-860: Automatically upgrade WASM runtime on Devnet

atodorov:
- Builds on top of #52. 
- Needs a functioning CI in order to execute and see that it works

PR #1412, opened 2026-09-22
feat(write-ability): sign messageId directly, mirror asc-contracts #54

didac-gluwa:
## Summary

Tracks [gluwa/asc-contracts#54](https://github.com/gluwa/asc-contracts/pull/54) ("Allow Inbox deliverMessage retry after destination failure"), currently **open, unmerged**. That PR:

- Replaces `Inbox`'s six-field `messageHash` vote preimage with a direct check that `messageId == OutboxTypes.computeMessageId(outbox, emitter, sequence, keccak256(payload), sourceChainId)`, and `validateVotes` now takes `messageId` itself as the signed digest.
- Adds a non-indexed `uint64 sequence` field to `Outbox.MessagePublished` (ahead of `canAck`) — this moves the event's `topic0`.
- Splits `MessageDelivered`/`MessageExecutionFailed` into `MessageReceived`/`MessageExecuted`/`DestinationFailed`, and makes destination failures retryable instead of terminal — but the attestor never touches `IInbox` directly (by design), so none of that affects this repo. That side lives entirely in the companion `asc-message-relayer` PR.

This PR mirrors the two changes that *do* touch the attestor:

- `common/write-ability/src/abi.rs` — add `sequence` to the `IOutbox::MessagePublished` sol! mirror, so decoding keeps matching the deployed Outbox. The existing `abi_surface` drift test (`common/write-ability/tests/abi_surface.rs`) will catch a mismatch here automatically once the `write-ability-e2e` pin moves past #54.
- `common/write-ability/src/hash.rs` — remove the now-obsolete six-field `message_hash()` builder and its golden vectors. There's nothing left to derive: `messageId` is already the value attestors sign.
- `attestor`'s `listener.rs` / `reobservation.rs` — stop computing a digest and sign the finalized (or independently re-verified) `messageId` directly. `IndexedMessage` drops the now-dead `emitter`/`payload`/`message_hash` fields.
- `MessageVote` drops the redundant `message_hash` field (always equal to `message_id` under the new scheme) — halves the gossiped vote size.
- Recompiled the `e2e_anvil` fixture Outbox (`fixtures/test-outbox.sol`) with the new event/function shape and reran the real listener → sign → ingest → aggregator path against a live Anvil node.

## Not included (deliberately)

`write-ability-e2e.yml`'s pins for `asc-contracts` and `asc-message-relayer` are **not** bumped here — asc-contracts#54 hasn't merged yet, and bumping now would break CI. Bump both pins together, in a follow-up, once:
1. asc-contracts#54 merges, and
2. the companion `asc-message-relayer` PR lands (it owns `deliverMessage`'s new `sequence` arg, the `MessageDelivered`/`MessageExecutionFailed` → `MessageReceived`/`MessageExecuted`/`DestinationFailed` split, and the relayer's own vote-hash mirror).

## Test plan

- [x] `cargo test -p write-ability -p attestor --all-features` — 151 passed (0 failed)
- [x] `cargo test -p attestor --test e2e_anvil -- --ignored` — real Anvil-backed run, passes against the recompiled fixture
- [x] `cargo clippy -p write-ability -p attestor --all-targets --all-features -- -D warnings` (matching CI's `rust.yml` flags) — clean
- [x] `cargo fmt -p write-ability -p attestor -- --check` — clean
- [ ] `write-ability-e2e.yml` (self-hosted, cross-repo) — not run here; gated on the pin bump follow-up above


BradleyOlson64 (approved):
LGTM

